PR #6, opened 2021-10-08
Add Algorithms prototype

timvermeulen:
Very bare-bones, currently undocumented and untested.

Includes:
* The `CollectionConsumer` family of protocols
* The `CollectionSearcher` family of protocols
* A handful of basic algorithms: `contains(_:)`, `firstRange(of:)`, `lastRange(of:)`, `ranges(of:)`, `starts(with:)`, `ends(with:)`, `trim(while:)`
* Implementations of the Z algorithm and the two-way search algorithm as `CollectionSearcher`s for finding a subsequence


Sources/Algorithms/Searchers/Searcher.swift
```diff
@@ -0,0 +1,77 @@
+public protocol CollectionSearcher {
+  associatedtype Searched: Collection
+  associatedtype State = ()
+  func initialState(_ searched: Searched) -> State
+  func search(
+    _ searched: Searched,
+    subrange: Range<Searched.Index>,
+    _ state: inout State
+  ) -> Range<Searched.Index>?
```
milseman, line 9: An interesting challenge will be whether or how we guarantee that this will be invoked. E.g. State might be specific to the last index returned, so we may want to ensure we're invoking it monotonically.

It's unfortunate that access control with protocols is so coarse-grained. Ideally, one could make a public conformer without making these low-level interfaces public. Essentially, all that is really required at an ABI level is `@_usableFromInline`.

Elsewhere we have underscored the name for some of these things, but hopefully we can avoid forcing that pattern.

Sources/Algorithms/Consumers/Consumer.swift
```diff
@@ -0,0 +1,24 @@
+public protocol CollectionConsumer {
+  associatedtype Consumed: Collection
+  func consume(_ consumed: Consumed, subrange: Range<Consumed.Index>) -> Consumed.Index?
```
rxwei, line 3: The imperative verb `consume` gives a mutating vibe to it according to the API guidelines.  Are there other naming options that could better align with the semantics of the method?

PR #8, opened 2021-10-09
Add in grapheme break data testing infrastructure

milseman:


milseman:
@swift-ci please test

milseman:
@rxwei I wasn't able to figure out how to explicitly type the captures there, but it's close. I know you're working in more type plumbing, so this might be a good (real) example for that.

milseman:
@swift-ci please test

PR #12, opened 2021-10-12
Regex literal pitch

hamishknight:


Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+There is also a conflict for infix operators containing `/` in cases where whitespace isn't used, for example:
+
+```swift
+x+/y/+z
+```
+
+Should the operators be parsed as `+/` and `/+` respectively, or should this be parsed as `x + /y/ + z`?
+
+In this case, things can be more readily disambiguated by the user inserting additional whitespace. We therefore could potentially continue to parse `x+/y/+z` as a binary operator chain, and require additional whitespace to interpret `/y/` as a regex literal. However, depending on the source compatibility impact, for consistency with the above proposed deprecation of prefix and postfix operators containing `/`, it may be prudent to also deprecate infix operators containing `/`. This would let us have a consistent rule, and would simplify the parsing logic required. Note though this deprecation would not affect the existing division operator `/`.
```
hamishknight, line 127: Ah whoops it's totally not possible to deprecate infix operators containing `/` as we have the `/=` operator...

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+public protocol RegexLiteralProtocol {
+  init()
+  
+  // Informal builder requirements for building a regex literal
+  // will be specified here.
+}
+```
+
+Types conforming to `ExpressibleByRegexLiteral` will be able to provide a custom type that conforms to `RegexLiteralProtocol`, which will be used to build the resulting regex value. A default conforming type `Regex` will be provided by the standard library, the details of which are out of scope for this pitch.
```
milseman, line 86:
```suggestion
Types conforming to `ExpressibleByRegexLiteral` will be able to provide a custom type that conforms to `RegexLiteralProtocol`, which will be used to build the resulting regex value. A default conforming type will be provided by the standard library (`Regex` here), but further details of that type are out of scope for this pitch.
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+Giving conforming types the flexibility to provide their own regex handling logic allows for e.g additional linguistic-level processing to be built on top of regular expressions (e.g. using Unicode word-break iterators, collation and/or locale support, fuzzier matching, etc). It could also be used to provide regex types that call directly into `NSRegularExpression`, or other existing regex libraries.
```
milseman, line 88:
```suggestion
Libraries can extend regex handling logic for their domains. For example, a higher-level library could provide linguistically richer regular expressions by incorporate locale, collation, language dictionaries, and fuzzier matching. Similarly, a libraries wrapping different regex engines can support custom regex literals.
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Opting into certain regex features
+
+We intend for the compiler parsing logic to completely parse [the PCRE syntax][PCRE], however types conforming to `RegexLiteralProtocol` may not be able to handle the full regex grammar. To signal the features that they support, conforming types will be able to annotate builder functions with availability; which allows them to communicate the features they don't intend on supporting, in addition to the features they may support in the future.
```
milseman, line 92:
```suggestion
We intend for the compiler to completely parse [the PCRE syntax][PCRE]. However, types conforming to `RegexLiteralProtocol` might not be able to handle the full feature set. The compiler looks for corresponding function declarations inside `RegexLiteralProtocol` and will emit a compilation error if missing. Conforming types can use `@availability` on these function declarations to communicate versioning and add more support in the future.
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+This also would give the standard library the flexibility to roll out new regex features over time.
```
milseman, line 94:
```suggestion
The feature set initially supported by the standard library is outside the scope of this pitch. But, the approach of lookup combined with availability allows us to extend support over time.
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Impact of using `/` as the delimiter
+
```
milseman, line 97: I think it makes sense to start with the easy/benign and work forwards from there. So probably start with comment syntax, etc., which will also help the reader develop their syntactic reasoning.

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+### Further extension to Swift language constructs (better name?)
+
+TODO: I'm referring to things like a Regex-backed `enum` for declaring tokens.
+
+TODO: Mention restructuring pattern-matching enhancements
+
+
+### Semantic details
+
+Further details about the semantics of regex literals, such as what definition we give to character classes, the initial supported feature set, and how to switch between grapheme-semantic and scalar-semantic usage, is outside the scope of this pitch and worthy of their own discussions.
```
milseman, line 204:
```suggestion
Further details about the semantics of regex literals, such as what definition we give to character classes, the initial supported feature set, and how to switch between grapheme-semantic and scalar-semantic usage, is still under investigation and outside the scope of this pitch.
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+However one of the main goals of this pitch is to introduce a familiar syntax for regular expression literals, which has been the motivation behind e.g using the PCRE regex syntax. Given the fact that `/` is an existing term of art for regular expressions, we feel that if the aforementioned parsing issues can be solved in a satisfactory manner, we should prefer it as the delimiter.
+
+### Raku-style syntax
+
+TODO: rejected as we want familiarity. If we're going to do something custom or different, then we'd like to design something beyond regex. At that point, we're talking about something pretty different (and certainly not tied to strings, e.g. data processing literals), and this word does not preclude improvements there.
+
+### Fully custom literals
+
+It may be worth exploring the ability for users to define custom literals, where they have full generality and no default type. This would require fleshing out the library-driven compilation story (bootstrapping as well as compiler and type system API). However that it is mostly orthogonal to this effort.
```
milseman, line 224:
```suggestion
It may be worth exploring the ability for users to define custom literals, where they have full generality and no default type. This would require fleshing out the library-driven compilation story (bootstrapping as well as compiler and type system API). However that it is mostly orthogonal to this effort.
```

And I think I want to address partially-custom delimiters in some fashion as well.

kylemacomber (commented):
I think we're saying "out of scope" too much in the current draft. I'd prefer we not say it at all, since the whole point of the pitch it to encourage feedback.

Instead, maybe we should preface the forum post saying this pitch is a discussion starter and we don't think it's complete enough to graduate to a full review as is. And then maybe we emphasize what we'd like specific feedback on (instead of emphasizing what we don't want feedback on).

Wdyt?

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Introduction
+
+We propose to introduce a first-class regular expression literal into the language that can take advantage of library support to offer extensible, powerful, and familiar textual pattern matching.
+
+This is a component of a larger string processing picture. We would like to start a focused discussion surrounding our approach to the literal itself, while acknowledging that evaluating the utility of the literal will ultimately depend on the whole picture (e.g. supporting API). To aid this focused discussion, details such as the representation of captures in the type system, semantic details, extensions to lexing/parsing, additional API, etc., are out of scope of this pitch and thread. Feel free to continue discussion of anything related in the [overview thread][overview].
+
+## Motivation
+
+Regular expressions are a ubiquitous, familiar, and concise syntax for matching text that satisfies a particular format, and extracting subranges from the match that contain interesting information. Syntactically, a regex literal in Swift should:
```
kylemacomber, line 13:
```suggestion
Regular expressions are a ubiquitous, familiar, and concise syntax for matching and extracting text that satisfies a particular pattern. Syntactically, a regex literal in Swift should:
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+- Leverage broader familiarity
```
kylemacomber, line 15:
```suggestion
- Be familiar and facilitate knowledge reuse
```

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+- Fit in with Swift's library extensibility story for literals (e.g. [`ExpressibleByStringInterpolation`][stringinterpolation])
+- Address source-destabilizing changes such as the parsing of delimiters
```
kylemacomber, line 17: I was confused by this bullet until I read on further, I could tell if it was talking about how Swift regular expression should support the parsing of delimiters (like matching parenthesis or quotation marks), but this is about the source-breaking implications of using `/` as the delimiter.

I think we can just delete this bullet since using the `/` isn't a motivation on its own—choosing `/` follows from the the underling motivation for Swift regex literals to be familiar.

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
+Further motivation, examples, and discussion can be found in the [overview thread][overview].
+
+## Proposed Solution
+
+We propose the introduction of a regular expression literal that supports [the PCRE syntax][PCRE], in addition to new standard library protocols `ExpressibleByRegexLiteral` and `RegexLiteralProtocol` that allow for the customization of how the regex literal is interpreted, similar to the existing `ExpressibleByStringInterpolation` protocol for string literals with interpolations. The compiler will parse the PCRE syntax within a regex literal, synthesizing calls to corresponding builder methods. Types conforming to `ExpressibleByRegexLiteral` will be able to provide a builder type that opts into supporting various regex constructs through the use of normal function declarations and `@available`.
```
kylemacomber, line 23: Do we want to saying anything (not necessarily in this paragraph) about why we chose the PCRE syntax compared to any other flavor? (I have so little experience with the different flavors of Regex, I don't know if PCRE is self-evidently the right one for those in the know.)

hamishknight: Makes sense to me, what do you think @milseman? As I understand it, our main motivation for choosing PCRE is the familiarity, is that correct?

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+
```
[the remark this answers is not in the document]
hamishknight: Added a line for this, though not sure whether it fits with Kyle's feedback, what do you think?

Documentation/Evolution/RegexLiteralPitch.md
hamishknight: Just noticed this says `@availability` in code voice, should that be `@available`?

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,230 @@
...
+## Detailed Design
+
+A regular expression literal will introduced using `/` delimiters, within which the compiler will parse [PCRE regex syntax][PCRE]:
+
+```swift
+let regex = /([0-9A-F]+\d)(?:\.\.([0-9A-F]+))?/
```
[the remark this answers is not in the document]
milseman: I updated with a new example, can you take a look?

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,259 @@
...
+    return .comment(text)
+  case let (l, u) <- /([0-9A-F]+)(?:\.\.([0-9A-F]+))?/:
+    return .scalars(Unicode.Scalar(hex: l) ... Unicode.Scalar(hex: u ?? l))
+  case let prop <- GraphemeBreakProperty.init:
+    return .property(prop)
+  }
+}
+```
+
+We'd like more community feedback on these future directions.
```
kylemacomber, line 231: Do we actually want community feedback on these directions in this thread? Seem like that'd be better left for the overview thread.

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,259 @@
...
+We propose to introduce a first-class regular expression literal into the language that can take advantage of library support to offer extensible, powerful, and familiar textual pattern matching.
+
+This is a component of a larger string processing picture. We would like to start a focused discussion surrounding our approach to the literal itself, while acknowledging that evaluating the utility of the literal will ultimately depend on the whole picture (e.g. supporting API). To aid this focused discussion, details such as the representation of captures in the type system, semantic details, extensions to lexing/parsing, additional API, etc., are out of scope of this pitch and thread. Feel free to continue discussion of anything related in the [overview thread][overview].
+
+## Motivation
+
+Regular expressions are a ubiquitous, familiar, and concise syntax for matching and extracting text that satisfies a particular pattern. Syntactically, a regex literal in Swift should:
+
+- Be familiar and facilitate knowledge reuse
+- Fit in with Swift's library extensibility story for literals
```
kylemacomber, line 16:
```suggestion
- Be familiar and facilitate knowledge reuse
- Facilitate some reuse of existing regexs (e.g. found on StackOverflow)
- Fit in with Swift's library extensibility story for literals
```

I think we could add one more bullet here about facilitating re-use of actual regexs and not just knowledge... also 3 is the magic number for number of bullets.

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,257 @@
...
+
+  // __C3 = /([0-9A-F]+)/
+  let __C1 = builder.customCharacterClass(["0"..."9", "A"..."F"])
+  let __C2 = builder.oneOrMore(__C1)
+  let __C3 = builder.captureGroup(__C2)
+
+  let __D1 = builder.concatenate(__A4, __B1, __C3)
+  return T(regexLiteral: builder.finalize(__D1))
+}()
+```
```
beccadax, line 65: This might be premature, but I have Thoughts.

1. We are both calling methods on a mutable builder instance, *and* returning values that are passed to later methods. So is the model here based on the methods mutating the builder to accumulate match operations into it, or is it based on the methods constructing an immutable value-type AST from the bottom up, or is it somehow both? (Maybe `builder` is a context that the AST nodes can be made in, but the relationship between those nodes is expressed by the return values? In that case, the initializer should probably be given some estimate information, much like `StringInterpolationProtocol`.)

2. If these methods are making AST nodes, should the method names indicate this? Perhaps by following the general `makeFoo` convention, or the result builder `buildFoo` convention? On the other hand, if they are having side effects on the builder, should they have verb-y names like `appendFoo` as we see in `StringInterpolationProtocol`?

3. Do we want to name metacharacters' methods based on specific semantics they're expected to follow? It might be better to have more mechanical translation rules, like:

> `\<identifier character>` turns into `makeEscaped_<identifier character>()` (except for characters that Swift string literals already assign a meaning to).

So for instance, the `__A2` line would be `let __A2 = builder.makeEscaped_w()`. That would free us from actually baking in a semantic for each character at this level, and would give us a consistent answer for characters that are used for different things in different dialects or aren't used in any of the dialects we know of. I also suspect it would be easier to remember than a lot of alternatives.

4. In `StringInterpolationProtocol`, I found that I didn't need a `finalize` because `init(stringInterpolation:)` could express any finalization logic I might need. The same might be true here.

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,257 @@
...
+This approach of lookup combined with availability allows the stdlib to support more features over time.
+
+### Impact of using `/` as the delimiter
+
+#### On comment syntax
+
+Single line comments use the syntax `//`, which would conflict with the spelling for an empty regex literal. As such, an empty regex literal would be forbidden.
+
+Multi-line comments use the `/*` delimiter. As such, a regex literal starting with `*` wouldn't be parsed. This however isn't a major issue as an unqualified `*` is already invalid regex syntax. An escaped `/\*/` regex literal wouldn't be impacted.
+
```
beccadax, line 102: A third comment issue that might be worth discussing: If we wanted to support multi-line regex literals, the most natural way to do that given current Swift syntax would be:

```
let regex = ///
            pattern
            ///
```

However, `///` is a comment syntax, and specifically it's the documentation comment syntax. Do we want to rule out this feature, or do we think there's a heuristic we can use to rescue it?

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,257 @@
...
+```swift
+SomeBuilder {
+  SomeType()
+  { print("hello") }
+  AnotherType()
+}
+```
+
+`{ print("hello") }` will be parsed as a trailing closure to `SomeType()` rather than as a separate element to the result builder.
```
beccadax, line 175: There's an even better analogy here to leading-dot syntax, which therefore isn't supported at the top level of a statement in a result builder.

Documentation/Evolution/RegexLiteralPitch.md
beccadax: (Someone on the overall pitch thread pointed this out—can't remember who, but you may want to give credit.)

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -0,0 +1,257 @@
...
+Regular expressions are a ubiquitous, familiar, and concise syntax for matching and extracting text that satisfies a particular pattern. Syntactically, a regex literal in Swift should:
+
+- Be familiar and facilitate knowledge reuse
+- Fit in with Swift's library extensibility story for literals
+
+Further motivation, examples, and discussion can be found in the [overview thread][overview].
+
+## Proposed Solution
+
+We propose the introduction of a regular expression literal that supports [the PCRE syntax][PCRE], in addition to new standard library protocols `ExpressibleByRegexLiteral` and `RegexLiteralProtocol` that allow for the customization of how the regex literal is interpreted (similar to [string interpolation][stringinterpolation]). The compiler will parse the PCRE syntax within a regex literal, and synthesize calls to corresponding builder methods. Types conforming to `ExpressibleByRegexLiteral` will be able to provide a builder type that opts into supporting various regex constructs through the use of normal function declarations and `@available`.
```
al45tair, line 22: Could we also support UTS#18? http://www.unicode.org/reports/tr18/

AFAIK PCRE doesn't support all of that, but it'd be good if we could at least parse it.

Documentation/Evolution/RegexLiteralPitch.md
kylemacomber: Agreed, these goals from @beccadax are great!

Documentation/Evolution/RegexLiteralPitch.md
kylemacomber: I do think these details @beccadax calls out are worth focusing on in this pitch. IIUC they're the protocol informal requirements and as such are more part of the literal than the concrete conforming `Regex` type we'll eventually provide.

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -12,8 +12,10 @@ This is a component of a larger string processing picture. We would like to star
 
 Regular expressions are a ubiquitous, familiar, and concise syntax for matching and extracting text that satisfies a particular pattern. Syntactically, a regex literal in Swift should:
 
-- Be familiar and facilitate knowledge reuse
-- Fit in with Swift's library extensibility story for literals
+- Support a syntax familiar to developers who have learned to use regular expressions in other tools and languages
+- Allow reuse of many regular expressions not specifically designed for Swift (e.g. from Stack Overflow or popular programming books)
+- Allow libraries to define custom types that can be constructed with regex literals, much like string literals
+- Diagnose at compile time if a regex literal uses capabilities that aren't allowed by the type's regex dialect
```
milseman, line 18: Nice! Thanks @beccadax

Documentation/Evolution/RegexLiteralPitch.md
```diff
@@ -98,6 +100,8 @@ This approach of lookup combined with availability allows the stdlib to support
 
 Single line comments use the syntax `//`, which would conflict with the spelling for an empty regex literal. As such, an empty regex literal would be forbidden.
 
+While not conflicting with the syntax proposed in this pitch, it's also worth noting that the `//` comment syntax (in particular documentation comments that use `///`) would likely preclude the ability to use `///` as a delimiter if we ever wanted to support multi-line regex literals. It's possible though that future multi-line support could be inferred from the regex options provided. For example, a regex that uses the multi-line option `/(?m)/` could be allowed to span multiple lines.
```
milseman, line 103: Does that mean that the parser would understand `(?m)` and switch behavior or heuristics?

Also, raw regex literals give us a future route here too.

Documentation/Evolution/RegexLiteralPitch.md
milseman: Let's continue talk about TR18 syntax on the thread

PR #14, opened 2021-10-12
Regex scalar classes

natecook1000:
This also includes support for literal metacharacters like `\d` and `\w`, and a `\uhhhh` metacharacter sequence for specifying a literal Unicode scalar that doesn't combine when matching.

### DIFF
--- a/Sources/Regex/CharacterClass.swift
+++ b/Sources/Regex/CharacterClass.swift
@@ -1,27 +1,80 @@
-public enum CharacterClass {
-  // Any character
-  case any
-  // Character.isDigit
-  case digit
-  // Character.isHexDigit
-  case hexDigit
-  // Character.isWhitespace
-  case whitespace
-  // Character.isLetter or Character.isDigit or Character == "_"
-  case word
+public struct CharacterClass: Hashable {
+  /// The actual character class to match.
+  var cc: Representation
+  
+  /// The level (character or Unicode scalar) at which to match.
+  var matchLevel: MatchLevel
+  
+  public enum Representation {
+    // Any character
+    case any
+    // Character.isDigit
+    case digit
+    // Character.isHexDigit
+    case hexDigit
+    // Character.isWhitespace
+    case whitespace
+    // Character.isLetter or Character.isDigit or Character == "_"
+    case word
+  }
+  
+  public enum MatchLevel {
+    /// Match at the extended grapheme cluster level.
+    case character
+    /// Match at the Unicode scalar level.
+    case unicodeScalar
+  }
 
-  public func matches(_ c: Character) -> Bool {
-    switch self {
-    case .any: return true
-    case .digit: return c.isNumber
-    case .hexDigit: return c.isHexDigit
-    case .whitespace: return c.isWhitespace
-    case .word: return c.isLetter || c.isNumber || c == "_"
+  /// Returns the end of the match of this character class in `str`, if
+  /// it matches.
+  public func matches(in str: String, at i: String.Index) -> String.Index? {
+    switch matchLevel {
+    case .character:
+      let c = str[i]
+      let next = str.index(after: i)
+      switch cc {
+      case .any: return next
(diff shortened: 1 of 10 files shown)


Sources/Regex/CharacterClass.swift
```diff
@@ -1,29 +1,94 @@
...
 
-  public func matches(_ c: Character) -> Bool {
-    switch self {
-    case .any: return true
-    case .digit: return c.isNumber
-    case .hexDigit: return c.isHexDigit
-    case .whitespace: return c.isWhitespace
-    case .word: return c.isLetter || c.isNumber || c == "_"
+extension CharacterClass {
+  public static var any: CharacterClass {
```
rxwei, line 58: How will the user create a character class with a different match level?

Sources/Regex/CharacterClass.swift
```diff
@@ -1,29 +1,94 @@
...
+  }
+  
+  public static var hexDigit: CharacterClass {
+    .init(cc: .hexDigit, matchLevel: .character)
+  }
+  
+  public static var word: CharacterClass {
+    .init(cc: .word, matchLevel: .character)
+  }
+
```
milseman, line 77: For now, how about a `var scalars: Self` and `var graphemeClusters: Self` or something more verbose like `scalarSemantic`?

Sources/Regex/RegexParse.swift
```diff
@@ -120,12 +122,46 @@ extension Parser {
...
-    case .character(let c, _)?:
+
+    case .character(let c, isEscaped: false):
       lexer.eat()
       partialResult = .character(c)
+
+    case .character("u", isEscaped: true):
+      lexer.eat()
+      var digits = ""
+      for _ in 0..<4 {
```
milseman, line 133: Any reason not to have the lexer tokenize this for us?

Sources/Regex/RegexParse.swift
```diff
@@ -120,12 +122,46 @@ extension Parser {
...
+      guard let value = UInt32(digits, radix: 16),
+            let scalar = UnicodeScalar(value)
+      else {
+        try report("invalid hex value after \\u: \(digits)")
+      }
+      partialResult = .unicodeScalar(scalar)
+      
+    case .character(let c, isEscaped: true):
+      lexer.eat()
+      if Token.MetaCharacter(rawValue: c) != nil {
```
milseman, line 149: Again, can the lexer construct the tokens?

Sources/Regex/VirtualMachine.swift
```diff
@@ -132,6 +145,22 @@ extension RECode: RandomAccessCollection {
   }
 }
 
+extension RECode {
+  public func withMatchLevel(_ level: CharacterClass.MatchLevel) -> RECode {
```
milseman, line 149: Should code instead have a level associated with it? Would it make sense to have one override the other or both?

Sources/Regex/RegexParse.swift
natecook1000: The lexer should definitely do this - I'll move it up and add a couple other formats.

Sources/Regex/CharacterClass.swift
natecook1000: That's a great question — tbh I'm not sure about the utility of distinguishing between capturing substrings, characters, and unicode scalars. If every capture is always a substring, we benefit in terms of simplicity (for things like this) and also providing ranges, since substrings double as a range pointing into the source input.

Sources/Regex/VirtualMachine.swift
natecook1000: I think that makes sense; in that case we'd need a `default` level, or to make it optional, so that we could distinguish between things like these — the second one should still match at the grapheme cluster level:

```swift
Regex {
    "a"
    .any
    "b"
}.scalarSemantic

Regex {
    "a"
    .any.graphemeClusterSemantic
    "b"
}.scalarSemantic
```

PR #15, opened 2021-10-15
Update RegexLiteralPitch.md with thread feedback

milseman:


milseman:
@hamishknight I think it makes sense to work on a v2 pitch that:

1. Discusses alternate delimiters more prominently
2. Clarifies exactly what the interior syntax is (e.g. superset of PCRE / UTS\#18 / ICU?)
3. Reworks the parsing ambiguity section a little bit so it's clear at a glance what we're actually proposing on changing and where the questions are

milseman:
@hamishknight I'm closing the PR, but feel free to yank prose from it if helpful.

PR #16, opened 2021-10-15
Add initial draft of character classes proposal

natecook1000:


Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Digits: `\d`,`\D`
+
+`\d` matches a "digit", which is any `Character` or Unicode scalar with the Unicode general category `Decimal_Number`.
+
+| Matching Level | Note |
+|-------------------------------|---------|
+| Character | any character where the general category of the first unicode scalar is `Decimal_Number` |
+| Unicode Scalar | any scalar where the general category is `Decimal_Number`, not including any combining marks |
+| POSIX | `[0-9]` |
```
kylemacomber, line 167: Either here or above, it might be nice to have one code example that uses POSIX semantics.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- `\Uhhhhhhhh`: Must use 8 hexidecimal digits.
+
+**DSL:** This functionality is supported through the DSL's use of string literals.
+
+```swift
+Pattern {
+    "abc"
+    "\u{032}"
+    "def"
+}
```
kylemacomber, line 289: Is this example any different than:

```swift
Pattern {
    "abc\u{032}def"
}
```

If not maybe we we should pick something more interesting or just use `"Cafe\u{301}"` again?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+| `[:upper:]`  | `[A-Z]`           |
+| `[:word:]`   | `[A-Za-z0-9_]`    |
+| `[:xdigit:]` | `[0-9A-Fa-f]`     |
+
+### Custom Classes: `[...]`
+
+Users can create custom character classes by using literal characters and most predefined character classes within square brackets.
+
+- Individual characters and character classes described above can be used as is.
+- Ranges of Unicode scalar values can be specified by separating the start and end of the range (inclusive) with a hyphen. Note that because `Character` isn't a `Comparable` type, ranges can only match Unicode scalar values. Crossing the UTF-16 surrogate pair range (U+D800 to U+DFFF) is permitted, but values in that range do not match anything, regardless of the string's encoding.
```
kylemacomber, line 338: I'm pretty sure `Character` is `Comparable`:

```swift
1> Character("A") < Character("B")
$R0: Bool = true
```

Also I think in the Regex Literal pitch we showed at least pseudo code that was proposing representing custom classes as a closed range of characters?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+While most Unicode-defined properties can only match at the Unicode scalar level, some are defined to match an extended grapheme cluster. For example, `/\p{RGI_Emoji_Flag_Sequence}/` will match any flag emoji character, which are composed of two Unicode scalar values.
+
+`\P{...}` matches the inverse of the characters matched by the corresponding `\p{...}` character class.
+
+**DSL:** This functionality is supported through APIs that take a predicate as a parameter, focused either on a `Character` or `UnicodeScalar` instance.
+
+```swift
+Pattern {
+    "abc"
+    OneOrMore(.unicodeProperty(where: \.isAlphabetic))
```
kylemacomber, line 307: I wonder if we'll want a spelling of this that's more amenable to hard-coding in the innards of the matching engine?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
+# Regular Expression Literals: Character Classes
```
kylemacomber, line 1: I wonder if the title should be less regex specific, e.g. "Declarative String Processing Character Classes"

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+
+- **Author:** [Nate Cook](https://github.com/natecook1000)
+- **Status:** Draft pitch
+
+## Introduction
+
+A regular expression literal comprises a combination of character classes, quantifiers (e.g. `?` and `+`), assertions, and a few other metacharacters. Character classes include metacharacters like `\d` to match a digit, `\s` to match whitespace, and `.` to match any character. Individual literal characters can also be thought of as character classes, as they at least match themselves, and, in case-insensitive matching, their case-toggled counterpart. 
+
+For the purpose of this work, then, we consider a *character class* to be any part of a regular expression literal that can match an actual component of a string. This proposal details these character classes, their denoting sequences in Swift regular expression literals, and describes their matching semantics.
+
```
milseman, line 11: We need a framing paragraph or two, which I can help author.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+Pattern exactPatt = Pattern.compile("é");
+exactPatt.matcher(cafe).find()      // false
+
+Pattern equivPatt = Pattern.compile("é", Pattern.CANON_EQ);
+equivPatt.matcher(cafe).find()      // true
+```
+
+Finally, there is wide variance of the actual definitions of character classes between different languages. As one example, should the `\d` character class, which matches digits, include the digits from the _Halfwidth and Fullwidth Forms_ Unicode block? ICU-based regular expression engines, like those used in Java and Foundation's `NSRegularExpression`, include these extra digits, while other languages do not. Determining exactly what characters are matched by each character class of a particular language or library can be difficult, making it harder to fully understand what kinds of strings a pattern will or will not match.
+
```
milseman, line 69: Lots of good prose in this section, but I do feel that it could be streamlined a bit. We're mostly saying:

- `Character` is extended grapheme cluster (link to Unicode), which means `"e\u{301}"` (provide name) is a single character. `dropFirst()` would remove it as a whole from the string.
- `Character`s compare under Unicode canonical equivalence (link to Unicode), thus `"e\u{301}"` is equal to (precomposed)
- We want regex behave the same as operating over `Character`, a mode where it behaves as operating over scalars, and various ways of selecting or defining certain sets (e.g. POSIX) etc.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+## Proposed Solution
+
+We propose a treatment of character classes in Swift regular expression that is consistent with the Swift string model, configurable at the point of declaration, and comprehensible for users getting to know Swift regular expressions.
+
```
milseman, line 73: What about configurable at point of use? Or configurable by API? I don't think we can make this decision here, as we don't have any existing pitched API to reference, nor are we pitching any such API here.

This also buries the lead and the point of this pitch. Support pattern matching over strings, including regex, requires defining character classes. We have some prior art in Character properties and Unicode Scalar properties, but those are incomplete for matching. Let's beef them up, clarifying what they correspond to in regex literals (and using regex as a forcing function for this much needed semantic work).

There's so much semantic work that we can focus on that instead of where mode is configured, while acknowledging that configuring the mode is a big unknown and pending the next step and investigation. Either way, we want to flesh out this semantic space so we can make concrete progress here.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+Just as the default view of a string's contents are as a collection of grapheme-cluster characters, with comparisons using canonical equivalence, regular expression matching by default will do the same. This assures that similar code produces predictable results as we add APIs that accept regular expressions and other patterns:
+
+```swift
+let str = "Cafe\u{301}"
+str.contains("é")       // true
+str.contains(/é/)       // true
+```
+
+Moreover, in accordance with the [Unicode guidelines for regular expressions](https://www.unicode.org/reports/tr18), we use Unicode property data rather than specific lists of characters whenever possible.
```
milseman, line 84: And what about grapheme clusters?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+
+For compatibility with other languages, and in some cases to allow users to opt into higher performance, you can configure the character class matching level of regular expression at the declaration site. For example, you can opt into the Unicode scalar-based matching of other languages by using the `unicodeScalarSemantics` property on a regular expression instance.
+
+This example declares `pattern` with Unicode scalar semantics, instead of the default character semantics. `pattern` doesn't match the string above because `"e"` and `"\u{301}"` are treated as separate values.
+
+```swift
+let pattern = /^Caf.$/.unicodeScalarSemantics
+str.contains(pattern)   // false
+```
+
```
milseman, line 97: This doesn't look like a good idea if `pattern` is of type Regex. The API doesn't know, and this use might be far from the declaration of the regex. Producing non-aligned indices from API should be explicit at the use site.

We can say that figuring out how the different modes play together and switch is out of scope, let's do it explicitly and consistently if so. We don't even have a pitched `contains` here or elsewhere, so it's a little odd to be making this decision here.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+Initially, we plan to allow configuration between character, Unicode scalar, and POSIX-compatible character class matching.
+
```
milseman, line 99: Is POSIX-or-not the only character class switching option supported by those other engines?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- the matched characters can be validated by accessing the properties directly.
+
+For example, the `\w` character class, for "word" characters, will have a matching `matchesAsWordCharacter` property.
+
+```swift
+"C".matchesAsWordCharacter  // true
+"é".matchesAsWordCharacter  // true
+"‡".matchesAsWordCharacter  // false
+```
+
```
milseman, line 116: Why is this better than isWordCharacter or isWord?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+We could further extend the API to allow the redefinition of individual character classes, such as making `\d` match only the characters in the range `0` through `9`.
+
+```swift
+let fiveDigits = /\d/.customizingClass(\.matchesAsDigit, toMatch: /[0-9]/)
+```
```
milseman, line 121: Are we talking about this? I thought we said just POSIX, grapheme, or scalar modes? Are we pitching these API?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+`\X` matches any grapheme cluster (`Character`), even when the regular expression is otherwise matching at the Unicode scalar or POSIX level.
+
+**DSL:** The "match anything" pattern is spelled `.anyCharacter`. `.anyCharacter` will match at the level of the overall pattern unless given a more specific level, so `.anyCharacter.graphemeScalarSemantics` is equivalent to the `\X` metacharacter.
+
+```swift
+Pattern {
+    OneOrMore(.anyCharacter)
+}
+```
+
```
milseman, line 158: Does `anyCharacter` have an option for excluding newlines? Is that a parameter or what?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+### Digits: `\d`,`\D`
+
+`\d` matches a "digit", which is any `Character` or Unicode scalar with the Unicode general category `Decimal_Number`.
```
milseman, line 161: What's the difference between Character.isNumber and a digit? 

I think all of these would make sense to propose the Character and Unicode.Scalar API first, then show how it corresponds to a built-in character classes in regex and an example of what it might look like in `Pattern`, noting that `Pattern` is still under development.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- `Titlecase_Letter`
+- `Other_Letter`
+- `Decimal_Number`
+
+| Matching Level | Note |
+|-------------------------------|---------|
+| Character | any character where the general category of the first unicode scalar is in the above list |
+| Unicode Scalar | any scalar where the general category is in the above list |
+| POSIX | `[A-Za-z0-9_]` |
+
```
milseman, line 195: How does this relate to `Character.isLetter`?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- `NEWLINE (NEL)` (U+0085)*
+- any character in the Unicode general category `Z`/`Separator`
+- the CR/LF sequence
+
+`\h` matches the tab character (U+0009) or any character in the Unicode general category `Space_Separator`.
+
+`\v` matches the five line separator characters above, marked with *, as well as (U+2028) and (U+2029), not including the CR/LF sequence.
+
+`\R` matches the same characters as `\v`, and in addition matches the CR/LF sequence.
+
```
milseman, line 225: What about `Character.isWhitespace`? What about `Unicode.Scalar.Properties.isWhitespace`? How does this fit together?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- `\t`: `.tabCharacter` or `"\u{09}"`
+- `\r`: `.carriageReturn` or `"\u{0d}"`
+- `\n`: `.lineFeed` or `"\u{0a}"`
+- `\f`: `.formFeed` or `"\u{0c}"`
+- `\0`: `.nul` or `"\u{0}"`
+- `\e`: `.escape` or `"\u{1b}"`
+- `\a`: `.bell` or `"\u{07}"`
+- `\b`: `.backspace` or `"\u{08}"`
+- `\cX`: `.control(_: UnicodeScalar)`
+
```
milseman, line 270: Is `\n` just LF, or is this configurable? How does this relate to `Character.isNewline`?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- `\t`: `CHARACTER TABULATION` (U+0009)
+- `\r`: `CARRIAGE RETURN (CR)` (U+000D)
+- `\n`: `LINE FEED (LF)` (U+000A)
+- `\f`: `FORM FEED (FF)` (U+000C)
+- `\0`: `NUL` (U+0000)
+- `\e`: `ESCAPE` (U+001B)
+- `\a`: `BELL` (U+0007)
+- `\b`: `BACKSPACE` (U+0008) (within a character set only)
+- `\cX`: The control character indicated by Control-`X`; in the code point range `0..<32`.
+
```
milseman, line 258: Did you check or reference a more modern set of control characters? Does the grapheme break `Control` property align with this?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- `\0`: `.nul` or `"\u{0}"`
+- `\e`: `.escape` or `"\u{1b}"`
+- `\a`: `.bell` or `"\u{07}"`
+- `\b`: `.backspace` or `"\u{08}"`
+- `\cX`: `.control(_: UnicodeScalar)`
+
+
+### Unicode values: `\u`, `\U`, `\x`
+
+Metacharacters that begin with `\u`, `\U`, or `\x` match a character with the specified Unicode scalar values. The format used by Swift string literals (`\u{hhh...}`) and several other legacy regular expression formats are all supported.
```
milseman, line 274: How does this interact with grapheme boundaries?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+}
+```
+
+### Unicode named values and properties: `\N`, `\p`, `\P`
+
+`\N{NAME}` matches a Unicode scalar value with the specified name. For example, `/\p{name=HANGUL SYLLABLE GAG}/` matches the character `"각"`.
+
+`\p{PROPERTY}` and `\p{PROPERTY=VALUE}` match a Unicode scalar value with the given Unicode property (and value, if given). Binary properties (such as `White_Space`) and either full or abbreviated general categories (e.g. `Uppercase_Letter` or `Lu`), can be used without a value. For other properties, provide the property and value to match. For example, `/\p{Numeric_Type=Decimal}+/` matches the full range of the string `"0① ₂³"`.
+
+While most Unicode-defined properties can only match at the Unicode scalar level, some are defined to match an extended grapheme cluster. For example, `/\p{RGI_Emoji_Flag_Sequence}/` will match any flag emoji character, which are composed of two Unicode scalar values.
```
milseman, line 298: Does that consume both scalars in other regex engines, or just one of the flag characters?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+**DSL:** This functionality is supported through APIs that take a predicate as a parameter, focused either on a `Character` or `UnicodeScalar` instance.
+
+```swift
+Pattern {
+    "abc"
+    OneOrMore(.unicodeProperty(where: \.isAlphabetic))
+    "def"
+}
+```
+
```
milseman, line 311: So... how does this generalize to grapheme semantic mode and/or is that outside scope?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+| `[:digit:]`  | `[0-9]`           |
+| `[:graph:]`  | `[\x21-\x7E]`     |
+| `[:lower:]`  | `[a-z]`           |
+| `[:print:]`  | `[[:graph:] ]`    |
+| `[:punct:]`  | `[-!"#$%&'()*+,./:;<=>?@[\\\]^_{|}~]` |
+| `[:space:]`  | `[ \t\n\r\f\v]`   |
+| `[:upper:]`  | `[A-Z]`           |
+| `[:word:]`   | `[A-Za-z0-9_]`    |
+| `[:xdigit:]` | `[0-9A-Fa-f]`     |
+
```
milseman, line 332: Impact on grapheme-semantic mode? How does `[:space"]` match `\r\n`?

Documentation/Evolution/CharacterClasses.md
milseman: Clarifying enumeration of grapheme clusters should be a main aim of this discussion.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+### Custom Classes: `[...]`
+
+Users can create custom character classes by using literal characters and most predefined character classes within square brackets.
+
+- Individual characters and character classes described above can be used as is.
+- Ranges of Unicode scalar values can be specified by separating the start and end of the range (inclusive) with a hyphen. Note that because `Character` isn't a `Comparable` type, ranges can only match Unicode scalar values. Crossing the UTF-16 surrogate pair range (U+D800 to U+DFFF) is permitted, but values in that range do not match anything, regardless of the string's encoding.
+- Include a literal hyphen by escaping, or by placing it first or last in the custom class.
+- Negate a custom class by placing a caret immediately after the opening bracket (e.g. `[^a-zA-Z]`).
```
milseman, line 340: Why are we talking about syntax? IMO this pitch's value-add is to clarify semantics given any syntactic direction (e.g. even if literals never ship, this is still valuable and worthwhile work). Pure syntax is covered in the regex literal pitch and this pitch is starting to tackle a carefully-scoped subset of the semantic work.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+
+
+## Alternatives Considered
+
+### Applying Regular Expressions to String Views
+
+A prior design allowed a user to choose their desired matching level and semantics by applying a regular expression to a string's `UnicodeScalarView`, `UTF16View`, or `UTF8View`. While initially appealing, this approach posed multiple problems:
+
+- Since we want regular expressions created with a literal to be usable all the same places as those created using the DSL, anything predicate-based needs to have the input type specified at the time of creation. It wouldn't make sense to add a `(Character) -> Bool` predicate to a regular expression, and then apply it to the string's `UTF8View`.
```
milseman, line 349: How would you pass an arbitrary `(Character) -> Bool` for use within a regex literal anyways?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- The matching semantics of a regular expression are closely tied to the way it's composed, so it would likely be unpredictable or even nonsensical to apply a regular expression written for one view of a string to be applied to another view.
```
milseman, line 350: Any more or less nonsensical than to pass a `Regex` that you don't control the declaration of to such an API?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+- It's unclear whether the two UTF-encoded views would treat their elements as individual `UInt16` or `UInt8` values, or whether they would retain some notion of being encoded Unicode data. If the former, how would character classes like `\u{...}` or `\p{...}` be used by the parser? And if the latter, what purpose does this serve beyond parsing the `UnicodeScalarView`?
```
milseman, line 351: They would consume a scalar, just like how `\X` consumes a grapheme cluster in scalar semantic mode. This is how code-unit based engines can consume and decode scalars, and it allows for the expression of cross-encoding regexes.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,354 @@
...
+
+For these reasons, regular expressions will target only the `StringProtocol`-conforming types — `String` and `Substring`. The other string views can be parsed using the more general `Collection`-based pattern matching, without the regular expression-specific features described in this proposal.
+
```
milseman, line 354: We can say that API discussion, how modes switch, etc., are out of scope of this semantics-clarifying pitch. But I think it's suspect to make this decision in a pitch when corresponding API doesn't exist yet nor is being pitched or discussed here.

milseman (commented):
Thanks for getting this pitch up! It's good to have a concrete basis for more discussion.

My main point of feedback is that we should be clear what the scope for discussion is. It seems like pitch as-is makes firm decisions about things we're considering outside the scope of this pitch (non-character-class API), is omitting what we were talking about as being the basis of our approach (API on `Character` and `Unicode.Scalar`), and doesn't mention some things that we want to be part of this discussion.

I also have specific questions as to the specific classes, but I apologize I wasn't able to refresh on UTS\#18 prior. I apologize for the brief responses, they're more so conversation starters and clarifying questions.

milseman:
## Introduction

[Declarative String Processing Overview][overview] presents regex-powered matching broadly, without details concerning syntax and semantics, leaving clarification to subsequent pitches. [Regular Expression Literals][literals] presents more details on regex _syntax_ such as delimiters and PCRE-syntax innards, but explicitly excludes discussion of regex _semantics_. This pitch and discussion aims to address a targeted subset of regex semantics: definitions of character classes. We propose a comprehensive treatment of regex character class semantics in the context of existing and newly proposed API directly on `Character` and `Unicode.Scalar`.

Character classes in regular expressions include metacharacters like `\d` to match a digit, `\s` to match whitespace, and `.` to match any character. Individual literal characters can also be thought of as character classes, as they at least match themselves, and, in case-insensitive matching, their case-toggled counterpart. For the purpose of this work, then, we consider a *character class* to be any part of a regular expression literal that can match an actual component of a string.

## Motivation

(Much abbreviated motivation. Too much example code using String API clouds the scope of this pitch. The comparison with other engines is great!)

(If you want to focus on the definitions, I can work on this too)

## Proposed Solution

This pitch is narrowly scoped to Swift definitions of character classes found in regexes. For each character class, we propose:

- A name for use in API
- A `Character` API, by extending Unicode scalar definitions to grapheme clusters
- A `Unicode.Scalar` API with modern Unicode definitions
- If applicable, a `Unicode.Scalar` API for notable standards like POSIX (or JS?)

We're proposing what we believe to be the Swiftiest definitions, referencing Unicode's [UTS\#18][uts18], [PCRE][pcre], [perl][perl], [Raku][raku], [Rust][rust], [Python][python], [C\#][csharp], [`NSRegularExpression` / ICU][icu], [POSIX][posix], [Oniguruma][oniguruma], (grep?), [Go][go], [C++][cplusplus], [RE2][re2], [Java][java] (kotlin if different?),  ... (or whatever your set is :-)


To extend scalar semantics to grapheme clusters, we're using (?algebra and?) rationale from [SE-0221: Character Properties][charpropsrationale].

## Detailed Design

` for (meta, name, bestStandard, rationale) in natesBrain { `

### `meta`

We propose `meta` be named `name` with the following definitions:

```swift
extension Character {
  /// \(Grapheme-semantic definition)
  public var isName: Bool { get }    
}

extension Unicode.Scalar {
  /// \(Scalar-semantic definition)
  public var isName: Bool { get }

  /// \(POSIX-semantic definition, if applicable)
  public var isPOSIXName: Bool { get }

  ...
}
```

_<details><summary>Rationale</summary>_

We picked `\(bestStandard)`'s definition (or mixture?) for `Unicode.Scalar`. (If not obvious, reasons).

We used `\(rationale)` for `Character`. (If not obvious, reasons).

</details>

(If the choice, especially grapheme, isn't obvious or we're not strongly convinced, make note of that. Just be straight forward and simple "We're unsure whether ... and would like more discussion with the community")

### Custom character class ranges

(Will we be comfortable making a recommendation soon, or are we going to say this is outside scope?)

Documentation/Evolution/CharacterClasses.md
Azoy: It would be useful to know which properties are only allowed at the scalar level and which ones are allowed at the grapheme cluster level.

milseman:
Hey @natecook1000, I rewrote the pitch [here](https://gist.github.com/milseman/c89c8b37c8bfa7734bf35bf8fd6afc5e)

It has `/// TODO` for you to plug in your class definitions and make sure to update the rationale disclosure triangles. It also has a list of evaluated engines/languages that should be updated to the real list (or the real list extended to include those :-).

I didn't commit it directly to this PR because I didn't want to clobber anything you had or were in the process of updating. I added myself as a second author (if that's ok with you), since a major part of this is extending scalar semantics to grapheme-clusters and you'll want someone to take the flak for that.

natecook1000:
@milseman Thanks so much — this is a really fantastic reformulation! I'm working on integrating my changes from before the weekend, and will update here when I've got those in so that we can pitch today. <3

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+  /// TODO   
+  public static var backspace: Unicode.Scalar { get }
+
+  /// TODO
+  public static func control(_ x: Unicode.Scalar): Unicode.Scalar { }
+}
+```
+
+**TODO**: What about `\r\n` in grapheme semantic mode?
```
milseman, line 315: **TODO**

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+}
+// ... same for Character ...
+```
+
+`Unicode.Scalar.isASCII` already exists and `Character.isASCII` and can satisfy `[:ascii:]`.
+
+Alternatively, we could introduce an option-set-like `POSIXCharacterClass` and `func isPOSIX(_:POSIXCharacterClass)` since POSIX is a fully defined standard. This would cut down on the amount of API noise directly visible on `Character` and `Unicode.Scalar` significantly.
+
+We'd like some more discussion with the community here, and it's possible this will become clearer as more of the string processing story takes shape.
+
```
milseman, line 409: Maybe we just strike the `public` additions to `Unicode.Scalar`. Let's give a treatment to which of these are not trivial compositions of `isFoo && isASCII` and which are meaningfully different. Then we can talk about adding `isPOSIXFoo` vs the option set approach. The option set approach might be `POSIXCharacterClass` or `ASCIICharacterClass`, ideally the latter if we want to bless ASCII-mode matching (which we mostly likely will want to do anyways).

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+
+### More classes or custom classes
+
+Future API might express custom classes or need more built-in classes. This pitch aims to establish rationale and precedent for a large number of character classes in Swift, serving as a basis that can be extended.
+
+
+## Alternatives Considered
+
+**NOTE** I say we leave this blank and fill this in as part of the discussion / second pitch. Or, we could reiterate that we'd like feedback on various things (e.g. alternate formulation for POSIX character classes).
```
milseman, line 452: Other people's thoughts?

milseman (commented):
Coming along!

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+  /// A Boolean value indicating whether this scalar is considered 
+  /// a digit.
+  ///
+  /// Any Unicode scalar in the general category `Nd`/`Decimal_Number`
+  /// is considered a digit. This includes the digits from the ASCII
+  /// range, from the _Halfwidth and Fullwidth Forms_ Unicode block,
+  /// as well as digits in some scripts, like `DEVANAGARI DIGIT NINE`
+  /// (U+096F).
+  public var isDigit: Bool { get }
+}
```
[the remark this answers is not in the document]
natecook1000: According to https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedNumericType.txt, those two ways of identifying scalars are identical.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+  /// along with the following control characters, are considered
+  /// whitespace:
+  ///
+  /// - `CHARACTER TABULATION` (U+0009)
+  /// - `LINE FEED (LF)` (U+000A)
+  /// - `LINE TABULATION` (U+000B)
+  /// - `FORM FEED (FF)` (U+000C)
+  /// - `CARRIAGE RETURN (CR)` (U+000D)
+  /// - `NEWLINE (NEL)` (U+0085)
+  public var isWhitespace: Bool { get }
```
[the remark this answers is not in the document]
natecook1000: They are indeed the same! I'll rewrite it in terms of that.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+extension Unicode.Scalar {
+  /// A Boolean value indicating whether this scalar is considered 
+  /// horizontal whitespace.
+  ///
+  /// All Unicode scalars with the general category 
+  /// `Zs`/`Space_Separator`, along with the control character 
+  /// `CHARACTER TABULATION` (U+0009), are considered horizontal 
+  /// whitespace.
+  public var isHorizontalWhitespace: Bool { get }
+}
```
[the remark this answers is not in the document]
natecook1000: I believe we are GC-based + the one control character here.

Documentation/Evolution/CharacterClasses.md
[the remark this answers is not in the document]
milseman: Sounds good. Let’s pick a documentation style or otherwise state they’re equivalent then.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+
+`\D` matches the inverse of `\d`.
+
+_<details><summary>Rationale</summary>_
+
+We chose the Unicode recommendation as the basis for `Unicode.Scalar`, which is to derive digit matching from the Unicode general category `Decimal_Number`. This behavior matches `NSRegularExpression` and the ICU regular expression specification, along with some languages with regular expressions in Unicode mode (yes: Perl, Python; no: ECMAScript, Java). For details, see [Unicode derived numeric types][derivednumeric].
+
+We chose to treat any grapheme cluster that leads with a Unicode scalar "digit" as a digit as well. This is compatible with the existing `Character.isNumber` property, which only checks the first scalar's numeric type. It does, on the other hand, diverge from the `isWholeNumber` and `isHexDigit` properties, which require that the `Character` comprises a single Unicode scalar.
+
+**TODO:** This is the right kind of matching for the rest of what we're doing, but it will yield substrings that will fail `Int.init?(_:radix:)`. Do we want to add another initializer that can understand this whole breadth of characters?
```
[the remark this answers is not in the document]
milseman: I think it is worth highlighting under future directions that we are interested in getting values out, but that’s a different discussion. It does affect some choices of naming though (e.g. isDecimalDigit)

Documentation/Evolution/CharacterClasses.md
milseman: We probably want a static var for crlf and it would probably be in that list on Characters.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,481 @@
...
+}
+```
+
+`\D` matches the inverse of `\d`.
+
+_<details><summary>Rationale</summary>_
+
+We chose the Unicode recommendation as the basis for `Unicode.Scalar`, which is to derive digit matching from the Unicode general category `Decimal_Number`. This behavior matches `NSRegularExpression` and the ICU regular expression specification, along with some languages with regular expressions in Unicode mode (yes: Perl, Python; no: ECMAScript, Java). For details, see [Unicode derived numeric types][derivednumeric].
+
+We chose to treat any grapheme cluster that leads with a Unicode scalar "digit" as a digit as well. This is compatible with the existing `Character.isNumber` property, which only checks the first scalar's numeric type. It does, on the other hand, diverge from the `isWholeNumber` and `isHexDigit` properties, which require that the `Character` comprises a single Unicode scalar.
```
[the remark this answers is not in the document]
milseman: I think we should actually go with `decimalDigit` if Unicode defines `\d` to be restricted to _decimal_ digits. We should also add a `var decimalDigitValue: Int?`. That is, we have restrictive semantics.

If there's any Unicode definition of digit distinct from numbers and decimal digits, we could add a permissive `isDigit` in the future.

I also wonder whether choice of restrictive/permissive will be relevant to future API.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,415 @@
...
+- A `Unicode.Scalar` API with modern Unicode definitions
+- If applicable, a `Unicode.Scalar` API for notable standards like POSIX (or JS?)
+
+We're proposing what we believe to be the Swiftiest definitions, referencing Unicode's [UTS\#18][uts18], [PCRE][pcre], [perl][perl], [Raku][raku], [Rust][rust], [Python][python], [C\#][csharp], [`NSRegularExpression` / ICU][icu], [POSIX][posix], [Oniguruma][oniguruma], (grep?), [Go][go], [C++][cplusplus], [RE2][re2], [Java][java], and [ECMAScript][ecmascript].
+
+To extend scalar semantics to grapheme clusters, we're using algebra and rationale from [SE-0221: Character Properties][charpropsrationale].
+
+## Detailed Design
+
+In general, we choose the Unicode recommendations in [UTS\#18][uts18] as the basis for matching character classes to `Unicode.Scalar`s. This is similar to the behavior of several languages' regular expressions when in Unicode mode (such as Perl and Python), as well as ICU-based regular expressions like `NSRegularExpression`.
```
milseman, line 58: If that's the case, we can rework/streamline some of the rationale. Do we at least have the superset of character classes across these engines?

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,415 @@
...
+
+To extend this matching behavior to `Character`s, we treat a `Character` as belonging to a character class if its leading Unicode scalar belongs to that class.
```
milseman, line 60: This is very much a case-by-case decision and I don't think it makes sense to declare a particular rationale unilaterally. Even the existing Character properties, restricted as they were to what was easier to express and less controversial, had two different ways.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,415 @@
...
+
+`\N{NAME}` matches a Unicode scalar value with the specified name. `\p{PROPERTY}` and `\p{PROPERTY=VALUE}` match a Unicode scalar value with the given Unicode property (and value, if given). 
+
+While most Unicode-defined properties can only match at the Unicode scalar level, some are defined to match an extended grapheme cluster. For example, `/\p{RGI_Emoji_Flag_Sequence}/` will match any flag emoji character, which are composed of two Unicode scalar values.
+
+`\P{...}` matches the inverse of `\p{...}`.
+
+Most of this functionality is already provided inside `Unicode.Scalar.Properties`, and we propose to round out Swift's current support with API to access the Unicode script(s) for a `Unicode.Scalar` or `Character`. (API TBD)
+
+**TODO**: Check with Alejandro that the code size impact is reasonable
```
milseman, line 326: **TODO** :-)

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,415 @@
...
+
+### Unicode named values and properties: `\N`, `\p`, `\P`
+
+`\N{NAME}` matches a Unicode scalar value with the specified name. `\p{PROPERTY}` and `\p{PROPERTY=VALUE}` match a Unicode scalar value with the given Unicode property (and value, if given). 
+
+While most Unicode-defined properties can only match at the Unicode scalar level, some are defined to match an extended grapheme cluster. For example, `/\p{RGI_Emoji_Flag_Sequence}/` will match any flag emoji character, which are composed of two Unicode scalar values.
+
+`\P{...}` matches the inverse of `\p{...}`.
+
+Most of this functionality is already provided inside `Unicode.Scalar.Properties`, and we propose to round out Swift's current support with API to access the Unicode script(s) for a `Unicode.Scalar` or `Character`. (API TBD)
```
milseman, line 324: **TODO** :-)

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,415 @@
...
+extension Unicode.Scalar {
+  /// Same as above...
+}
+```
+
+**TODO**: What about `\r\n` in grapheme semantic mode?
+
+_<details><summary>Rationale</summary>_
+
+This approach simplifies the use of some common control characters, while making the rest available through a method call.
```
milseman, line 310: We can omit the rationale triangle if vacuous.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -49,15 +49,21 @@ This pitch is narrowly scoped to Swift definitions of character classes found in
...
+We're proposing what we believe to be the Swiftiest definitions using [Unicode's guidance][uts18] for `Unicode.Scalar` and extending this to grapheme clusters using `Character`'s existing [rationale][charpropsrationale].
 
-To extend scalar semantics to grapheme clusters, we're using algebra and rationale from [SE-0221: Character Properties][charpropsrationale].
+<details><summary>Broad language/engine survey</summary>
 
-## Detailed Design
+For these definitions, we cross-referenced Unicode's [UTS\#18][uts18] with [PCRE][pcre], [perl][perl], [Raku][raku], [Rust][rust], [Python][python], [C\#][csharp], [`NSRegularExpression` / ICU][icu], [POSIX][posix], [Oniguruma][oniguruma], (grep?), [Go][go], [C++][cplusplus], [RE2][re2], [Java][java], and [ECMAScript][ecmascript].
+
+We found that ... **TODO** these are all subsets of UTS18, right? This is where to mention "ICU", NSRegularExpression, etc., but _definitely_ also perl/PCRE and Raku. Wait, does C\# do scalars or just UTF-16 code units? I feel like we've done a lot of research, so it would be nice to present some of it. This could be a table or just a couple of sentences.
```
milseman, line 58: @natecook1000 can you plug in your research here? I'll work on the other parts

Documentation/Evolution/CharacterClasses.md
kylemacomber: I think we can just delete this TODO. It seems like we need to provide this functionality so we'll just need to strike the right code-size/runtime efficiency balance in implementing it.

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,421 @@
...
+Regardless of any syntax or underlying formalism, classifying characters is a worthy and much needed addition to the Swift standard library. We believe our thorough treatment of every character class found across many popular regex engines gives Swift a solid semantic basis.
+
+## Proposed Solution
+
+This pitch is narrowly scoped to Swift definitions of character classes found in regexes. For each character class, we propose:
+
+- A name for use in API
+- A `Character` API, by extending Unicode scalar definitions to grapheme clusters
+- A `Unicode.Scalar` API with modern Unicode definitions
+- If applicable, a `Unicode.Scalar` API for notable standards like POSIX (or JS?)
```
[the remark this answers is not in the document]
milseman: Noting that we don't do this in the current pitch

Documentation/Evolution/CharacterClasses.md
```diff
@@ -0,0 +1,452 @@
...
+
+`\W` matches the inverse of `\w`.
+
+_<details><summary>Rationale</summary>_
+
+The Unicode recommendation is to derive word character matching from Unicode properties and general categories as described above
+
+We chose to treat any grapheme cluster that leads with a Unicode scalar word character as a word character as well.
+
+**TODO** How does this relate to `isLetter`? Why don't we have `alphaNumeric`? Let's establish proper subsetting here.
```
natecook1000, line 189: `isLetter` means any scalar that is part of the Letter parent category (Lu, Ll, Lt, Lm, Lo). That's a subset of `Alphabetic`, which also includes the Nl / Letter Numbers category and `Other_Alphabetic`, which is defined in PropList.txt.

I don't see `alphaNumeric` defined as a Unicode property, but it looks like it generally is what we've defined as a word character without the punctuation, marks, and connectors.

PR #19, opened 2021-10-19
More algorithms

timvermeulen:
Adds the `replacing(_:with:)` and `split(_:)` algorithms.

Also changes `CollectionSearcher` from

```swift
public protocol CollectionSearcher {
  associatedtype Searched: Collection
  associatedtype State = ()

  func initialState(_ searched: Searched) -> State
  func search(
    _ searched: Searched,
    subrange: Range<Searched.Index>,
    _ state: inout State
  ) -> Range<Searched.Index>?
}
```

to

```swift
public protocol CollectionSearcher {
  associatedtype Searched: Collection
  associatedtype State = Searched.Index
  
  func state(startingAt index: Searched.Index, in searched: Searched) -> State
  func search(_ searched: Searched, _ state: inout State) -> Range<Searched.Index>?
}
```

So now the state is responsible for maintaining the current index. This is helpful because it makes it impossible to search from a particular index that does not correspond to the current search state. `BidirectionalCollectionSearcher` does need more design work now so it's left out for the time being.

---

More changes:

* Added regex overloads for all of our algorithms (not yet documented).
* Added some regex tests for `ranges`, `replacing`, and `split`.
* Empty matches are now handled correctly. When finding an empty match, the searcher is responsible for advancing its state past it, which `StatelessCollectionSearcher` will do automatically. `ranges(of: subsequence)` wraps the `TwoWaySearcher` in a new `PatternOrEmpty` type that produces the correct empty ranges when the pattern is empty.
* `CollectionConsumer.Consumed` and `CollectionSearcher.Searched` are now both required to be their own `SubSequence`, and algorithms such as `ranges(of:)` now require that the provided searcher operates on `SubSequence` instead of `Self`. This makes using these algorithms way more pleasant, e.g. allowing `self.trimmingPrefix(searcher).trimmingSuffix(searcher)` without needing two separate searchers.

### DIFF
--- a/Sources/Algorithms/Consumers/RegexConsumer.swift
+++ b/Sources/Algorithms/Consumers/RegexConsumer.swift
@@ -26,10 +26,10 @@ public struct RegexConsumer: CollectionConsumer {
     _ consumed: Substring, from index: String.Index
   ) -> String.Index? {
     let result = vm.execute(
-      input: consumed, in: index..<consumed.endIndex, .partialFromFront)
+      input: consumed.base, in: index..<consumed.endIndex, .partialFromFront)
     assert(result?.range ==
            referenceVM.execute(
-            input: consumed, in: index..<consumed.endIndex, .partialFromFront
+            input: consumed.base, in: index..<consumed.endIndex, .partialFromFront
            )?.range)
 
     return result?.range.upperBound
--- a/Sources/Regex/CharacterClass.swift
+++ b/Sources/Regex/CharacterClass.swift
@@ -39,7 +39,7 @@ public struct CharacterClass: Hashable {
   
   /// Returns the end of the match of this character class in `str`, if
   /// it matches.
-  public func matches(in str: Substring, at i: String.Index) -> String.Index? {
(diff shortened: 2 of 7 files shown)


timvermeulen:
> What kinds of options or behaviors do you think we want for split collection? E.g. the ability to access the separators, behavior on empty-but-successful matches, etc?

Thanks for bringing this up, I think the behavior around empty matches will boil down to how `ranges` treats those. And I suppose that in turn will boil down to whatever the particular `CollectionSearcher` does? Given that the signature of `CollectionSearcher.search` is now

```swift
func search(_ searched: Searched, _ state: inout State) -> Range<Searched.Index>?
```

the searcher can advance to some other index than the upper bound of the matched range, if needed. I figured this could be useful for skipping an element that the searcher has already looked at, but it hadn't crossed my mind that this could also come in handy for dealing with empty matches.

I haven't put much thought into maybe wanting some way to access the separators, maybe that would be a different algorithm entirely. For `split` I think it would make sense to match the behavior of the existing `split` methods in the standard library.

PR #20, opened 2021-10-20
Support typed captures and capture transforms in regex engine and DSL.

rxwei:
This patch makes typed captures usable and adds support for transforming a capture with a closure passed to `capture(_:)`. To make this work, the regex engine is modified to store structured capture information, aka. `Capture`, which will be converted to statically inferred capture types after matching.

Changes in the regex engine:
- AST: Make `.captureGroup` store an optional transformation.
- IR: Add instructions for capturing substrings ("atoms"), optionals, and arrays.
- Compiler: Emit appropriate instructions for forming captures and capture groups.
- VM: Store a stack of groups and an array of top-level captures, which gets accumulated as the VM executes. When forming capture groups, use Swift type metadata to form arbitrarily sized tuples.

Note:
- Captures of containers (e.g. collections and optionals) currently produce `Substring` instead of the container. For example:
  ```swift
  OneOrMore("x").capture { c in ... // c has type `Substring`, not `[Substring]`.
  ```
  We will need to decide whether this is the optimal behavior.


rxwei:
@swift-ci please test

rxwei:
> Is this ready to hook up to the Exercises target?
Yep, it's ready to use!

natecook1000:
> Captures of containers (e.g. collections and optionals) currently produce Substring instead of the container. For example:
> ```
> OneOrMore("x").capture { c in ... // c has type `Substring`, not `[Substring]`.
> ```
> We will need to decide whether this is the optimal behavior.

This definitely seems like the correct behavior — `OneOrMore` expands the size of the match, while `Repeat` multiplies it to be several matches.

PR #22, opened 2021-10-20
Implement the pitched Character and Unicode.Scalar class APIs

natecook1000:
This includes all the APIs covered in the first round of the character classes Evolution pitch, along with Boolean properties on Unicode.GeneralCategory that correspond with the Unicode general category groups (e.g. `Letter` and `Mark`).

### DIFF
--- a/Sources/Util/CharacterClasses.swift
+++ b/Sources/Util/CharacterClasses.swift
@@ -1,0 +1,380 @@
+extension Character {
+  internal var isSingleScalar: Bool {
+    unicodeScalars.index(after: unicodeScalars.startIndex)
+      == unicodeScalars.endIndex
+  }
+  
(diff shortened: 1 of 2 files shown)


Sources/Util/CharacterClasses.swift
```diff
@@ -0,0 +1,380 @@
...
+  /// "Other" general category group.
+  public var isOther: Bool {
+    switch self {
+    case .control, .format, .surrogate, .privateUse, .unassigned:
+      return true
+    default:
+      return false
+    }
+  }
+}
```
milseman, line 105: Nice!

milseman:
What's the status of this PR? Should we go forwards or close it?

natecook1000:
These will need to be the subject of a future evolution proposal.

PR #26, opened 2021-10-27
Bidirectional searching

timvermeulen:
Bidirectional searching WIP.

Current approach:
* If a searcher produces the same ranges forward and backward then it can implement `BidirectionalCollectionSearcher`. If not, only `CollectionSearcher` and/or `BackwardCollectionSearcher`.
* The `ReversedRangesCollection` type traverses the ranges from the back of the collection and is generic over a `BackwardCollectionSearcher`, so it makes no guarantees about the ranges matching the forward direction. Can be obtained via `BidirectionalCollection.rangesFromBack(of searcher: Searcher)`.
* If the searcher of a `RangesCollection` is a bidirectional searcher then a `reversed()` method is available which returns a `ReversedRangesCollection`, and vice versa. If the searcher is not bidirectional then `reversed()` will use the default `Sequence` overload instead.

This approach does not give us `BidirectionalCollection` conformance. This can be achieved by having the index store the forward state as well as the backward state, both optional, and generating the necessary state from scratch if it's unavailable. This shouldn't result in any extra computations when just going one way, but it does require an extra `BidirectionalRangesCollection` type which may not carry its weight.


timvermeulen:
* Moved everything into the `_StringProcessing` module.
* Have `CollectionConsumer` and `CollectionSearcher` take an index range instead of a single index to start from.

timvermeulen:
@swift-ci Please test

Sources/_StringProcessing/Algorithms/Consumers/PredicateConsumer.swift
```diff
@@ -0,0 +1,42 @@
+public struct PredicateConsumer<Consumed: Collection> {
```
milseman, line 1: Why is this public? It serves as an interesting example of a trivial conformer and a nice way to test that composition and layering work, but IIUC I wouldn't expect people to explicitly make one or for it to appear in API.

Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
```diff
@@ -0,0 +1,68 @@
+import _MatchingEngine
+
+public struct RegexConsumer<Consumed: BidirectionalCollection> where Consumed.SubSequence == Substring {
```
milseman, line 3: Does this need to be public?

milseman (commented):
Before we merge this, let's vet what needs to be public.

Sources/_StringProcessing/Algorithms/Consumers/PredicateConsumer.swift
timvermeulen: It's currently used in the return type of a couple `split` overloads, so I had to make this type public or wrap those return types into new public types and make this one private. I went with the easy route initially but I can certainly wrap it if we think that's worth it. More powerful opaque return types will solve this too, so that is mainly why I didn't bother.

Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
timvermeulen: Also used in some return types of public methods, in this case `.ranges(of: /.../)` and `.split(by: /.../)`.

timvermeulen:
@swift-ci Please test linux

timvermeulen:
@swift-ci Please test linux

timvermeulen:
@swift-ci Please test linux

Sources/_StringProcessing/Algorithms/Consumers/PredicateConsumer.swift
milseman: Makes sense. Opaque return types will help, and we'll see what ends up in the `Swift` default import vs outside.

rxwei (commented):
Do we want to set a column limit for this repo? The existing code mostly uses an 80-character limit. We could also use the 100-character limit as specified in the [Google Swift Style Guide](https://google.github.io/swift/).

kylemacomber:
We should probably hue to the stdlib style guide

timvermeulen:
@swift-ci Please test linux

timvermeulen:
@swift-ci Please test linux

PR #30, opened 2021-10-29
Flexible classes

natecook1000:
This changes the regex parse stage to build grapheme clusters out of a character followed by a Unicode scalar literal or a series of Unicode scalar literals that comprise a grapheme cluster. This preserves the order of the Unicode scalars, so that if Unicode scalar matching semantics are required, the characters can be "exploded" into sequences of scalars.

This also moves the matching level up into `REOptions` instead of staying within each character class instance.

Still to do:
- match characters at the right level — `/e\{301}/` should match both `"é"` and `"e\u{301}"` at the grapheme cluster level, but only `"e\u{301}"` at the unicode scalar and UTF8 levels.
- figure out the right behavior for custom classes at non-grapheme-cluster levels, if they're always ranges of characters

### DIFF
--- a/Sources/Regex/CharacterClass.swift
+++ b/Sources/Regex/CharacterClass.swift
@@ -7,9 +7,9 @@ public struct CharacterClass: Hashable {
   var isInverted: Bool = false
 
   public enum Representation: Hashable {
-    /// Any single elements
+    /// Any single element, re the current matching semantics
     case any
-    /// Any grapheme cluster, even in non-grapheme-cluster mode
+    /// Any grapheme cluster, even when in a non-grapheme-cluster mode
     case anyGraphemeCluster
     /// Character.isDigit
     case digit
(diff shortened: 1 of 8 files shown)


Sources/Regex/CharacterClass.swift
```diff
@@ -63,14 +72,57 @@ public struct CharacterClass: Hashable {
...
-    public func matches(_ character: Character) -> Bool {
-      switch self {
-      case .character(let c): return c == character
-      case .range(let range): return range.contains(character)
-      case .characterClass(let custom):
-        let str = String(character)
-        return custom.matches(in: str, at: str.startIndex) != nil
-      case .setOperation(let op): return op.matches(character)
+    public func matches(in str: String, at i: String.Index, options: REOptions) -> String.Index? {
+      guard #available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) else { fatalError() }
```
milseman, line 76: We can also look at adding a version requirement to the package or a target

milseman:
Ping, what should we do with this PR?

PR #35, opened 2021-11-04
[Pitch] Strongly typed regex captures

rxwei:
Add Swift Evolution pitch: Strongly typed regex captures. Feedback welcome!


Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+
+A non-capturing group's capture type is the capture type of its underlying pattern.
+
+```
+pattern ::= '(' '?' ':' pattern ')'
+```
+
+```swift
+let graphemeBreakLowerBound = /([0-9A-F]+)(?:\.\.([0-9A-F]+))?/
+// => `Regex<Void>`
```
xAlien95, line 240:
```suggestion
// => `Regex<(Substring, Substring?)>`
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+#### Nested capturing group
+
+When capturing group is nested within another capturing group, they count as two
+distinct captures in the order their left parenthesis first appears in the
+regular expression literal. This is consistent with PCRE and allows us to use
+backreferences (e.g. `\2`) with linear indices.
+
+```swift
+let graphemeBreakPropertyData = /(([0-9A-F]+)(\.\.([0-9A-F]+)))\s*;\s(\w+).*/
+// Positions in capture tuple: 0 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~    4 ^~~~~
```
xAlien95, line 262: In PCRE, capture group indices start from 1, `\0` (or `$0` when substituting) refers to the whole match.
```swift
let graphemeBreakPropertyData = /(([0-9A-F]+)(\.\.([0-9A-F]+)))\s*;\s(\w+).*/
// Positions in capture tuple:  2 ^~~~~~~~~~~   4 ^~~~~~~~~~~      5 ^~~~~
//                                         3 ^~~~~~~~~~~~~~~~~
//                             1 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The result builder DSL equivalent is missing a `.capture()`:
```swift
//     let graphemeBreakPropertyData = Regex {
//         Regex {
//             OneOrMore(CharacterClass.hexDigit).capture() // (2)
//             Regex {
//                 ".."
//                 OneOrMore(CharacterClass.hexDigit).capture() // (4)
//             }.capture() // (3)
//         }.capture() // (1)
//         Repeat(CharacterClass.whitespace)
//         ";"
//         CharacterClass.whitespace
//         OneOrMore(CharacterClass.word).capture() // (5)
//         Repeat(CharacterClass.any)
//     }
```

kylemacomber: Ya, in what we've laid out here there will be an inconsistency between back reference numbering `\1` and tuple number `.0`. I don't think it's as intuitive for new users, but we could always include the whole match in the capture type and put it at position `0` to align with PCRE.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+
+## Motivation
+
+Across a variety of programming languages, many established regular expression
+libraries present captures as a collection of captured content to the caller
+upon a successful match
+[[1](https://developer.apple.com/documentation/foundation/nsregularexpression)][[2](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.capture)].
+However, to know the structure of captured contents, programmers often need to
+read the regular expression, which may be a literal or a series of API calls, or
+run the regular expression on some input to find out. Because regular
```
kylemacomber, line 88: I found this sentence a little hard to parse. Maybe:
```suggestion
However, to know the structure of captured contents, programmers often need to
carefully read the regular expression or run the regular expression on some
input to find out. Because regular
```

Documentation/Evolution/StronglyTypedCaptures.md
kylemacomber: @natecook1000 and I have been talking about this. In the current draft the match algorithm is implicitly conceptualized as:

```swift
extension String {
  func match<R: RegexProtocol>(_ regex: R) -> (match: Substring, captures: R.Captures)?
}
```

If we flattened the captures into the resulting tuple, then the index based access would align with the backreference numbering:

```swift
// longhand spelling
extension String {
  func match<R: RegexProtocol>(_ regex: R) -> (match: Substring)? where R.Captures == Void
  func match<R: RegexProtocol, C0>(_ regex: R) -> (match: Substring, C0)? where R.Captures == C0
  func match<R: RegexProtocol, C0, C1>(_ regex: R) -> (match: Substring, C0, C1)? where R.Captures == (C0, C1)
  ...
}

// variadic generics spelling
extension String {
  func match<R: RegexProtocol, T...>(_ regex: R) -> (match: Substring, T...)? where R.Captures == (T...)
}
```

The main avenue I see this foreclosing is returning a richer type than a tuple from `match` that we can hang additional operations off of.

kylemacomber (commented):
We should be consistent in this document with using the method name `firstMatch` vs `match`. I think we're more likely to go with `firstMatch`, so I think we should start using that.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+    .flatMap {
+        if let binary = $0.captures.binary {
+            return Int(binary, radix: 2)
+        } else if let decimal = $0.captures.decimal {
+            return Int(decimal, radix: 10)
+        } else if let hex = $0.captures.hex {
+            return Int(hex, radix: 16)
+        } else {
+            fatalError("unreachable")
+        }
```
kylemacomber, line 500: We could eliminate some this `.captures` boilerplate by using `@dynamicMemberLookup` on the `Alternation` type:

```swift
if let binary = $0.binary {
    return Int(binary, radix: 2)
} else if let decimal = $0.decimal {
    return Int(decimal, radix: 10)
} else if let hex = $0.hex {
    return Int(hex, radix: 16)
} else {
    fatalError("unreachable")
}
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+  based on matched ranges could minimize reference counting operations on
+  `Substring` and reduce allocations. However, array in capture types would make
+  a much cleaner type signature.
+- For repetitions or an exact count, e.g. `[a-z]{2}`, it would slightly improve
+  type safety to make its capture type be a homogeneous tuple instead of an
+  array. However, because large homogeneous tuples are difficult to work with
+  and because fixed-count repetitions are syntactically very similar to
+  dynamic-count repetitions which produce an array, an tuple capture may come as
+  a surprise. An array on the other is a better choise for both ease of
+  use and consistency.
```
natecook1000, line 377: Do we want a note that lazy quantifiers don't affect the capture type? 

Maybe something like:
```
- Lazy quantifiers have the same capture type semantics as their greedy counterparts.
  If the first example literal from above were instead written as `([0-9a-fA-F]+)+?`,
  it would yield the same `Regex<[Substring]>` type.
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+
+let input = "007F..009F    ; Control # Cc  [33] <control-007F>..<control-009F>"
+// Match result for `input`:
+// ("007F..009F", "007F", "..009F", "009F", "Control")
+```
+
+
+#### Quantification
+
+A quantifier's capture type depends on its underlying pattern.
```
natecook1000, line 292: Should we note that this is a big difference from most implementations, where you only have access to the last match of a repeated capture group?

Documentation/Evolution/StronglyTypedCaptures.md
xAlien95: Thank you for the internal discussion. A flattened match+captures tuple is what I thought of too initially. As for "returning a richer type", I guess it could still be done. With ~~variadic generics~~ tuple splatting spelling
```swift
struct Match<Captures> {
  let captures: (Substring, Captures...)
  var match: Substring { captures.0 }
}
```
You would be able to access the full match by either using `.match` or `.captures.0`. Without tuple splatting, you could either declare types on the lines of
```swift
struct Match0 {
  let captures: Substring
  var match: Substring { captures }
}
struct Match1<C1> {
  let captures: (Substring, C1)
  var match: Substring { captures.0 }
}
struct Match2<C1, C2> {
  let captures: (Substring, C1, C2)
  var match: Substring { captures.0 }
}
...

// longhand spelling
extension String {
  func match<R: RegexProtocol>(_ regex: R) -> Match0? where R.Captures == Void
  func match<R: RegexProtocol, C1>(_ regex: R) -> Match1<C1>? where R.Captures == C1
  func match<R: RegexProtocol, C1, C2>(_ regex: R) -> Match2<C1, C2>? where R.Captures == (C1, C2)
  ...
}
```
Or, more concisely, declare a provisional `Match_` type having `(Substring, T...)` as its generic parameter:
```swift
// Captures == (Substring, T...)
struct Match_<Captures> {
  let captures: Captures
  var match: Substring { /* unsafely point to `captures.0` and bound it to `Substring` */ }
}

// longhand spelling
extension String {
  func match<R: RegexProtocol>(_ regex: R) -> Match_<Substring>? where R.Captures == Void
  func match<R: RegexProtocol, C1>(_ regex: R) -> Match_<(Substring, C1)>? where R.Captures == C1
  func match<R: RegexProtocol, C1, C2>(_ regex: R) -> Match_<(Substring, C1, C2)>? where R.Captures == (C1, C2)
  ...
}
```
I expect the `Match_` type to not have public initializers, so it's safe to assume that its generic parameter will always be a tuple (or directly `Substring`) having `Substring` as its first element.

Documentation/Evolution/StronglyTypedCaptures.md
natecook1000: For example, Python only gives you access to the last group of this dash-separated string:

```python
rep = re.compile('(?:([0-9a-f]+)-?)+')
match = rep.match("1234-5678-9abc-def0")
print(match.group(1))
# Prints "def0"
```

Whereas the proposed version will provide access to all four sub-matches:

```swift
// Note: Syntax / API still in discussion
let pattern = /(?:([0-9a-f]+)-?)+/
if let match = "1234-5678-9abc-def0".firstMatch(of: pattern) {
    print(match.captures)
}
// Prints ["1234", "5678", "9abc", "def0"]
```

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Good point.

Documentation/Evolution/StronglyTypedCaptures.md
kylemacomber: Ya, actually as I mention below with `Alternation` we can even add `@dynamicMemberLookup` to the `Match` type to allow referencing captures by index or name directly. The only thing we'd lose with a nominal type is tuple destructuring... which isn't nothing.

I don't think we want the "match" to be part of the captures generic type, because when we get around the result builder syntax it won't compose as well.

So maybe it could look something like:

```swift
@dynamicMemberLookup
struct Match<Captures...> {
  let _storage: (Substring, Captures...)
  let match: Substring
  let captures: (Captures...)

  subscript<Value>(dynamicMember keyPath: KeyPath<(Substring, Captures...), Value>) -> Value {
    _storage[keyPath: keyPath]
  }
}

extension String {
  func match<R: RegexProtocol>(_ regex: R) -> Match<R.Captures>?
}
```

This would allow you to index on the match type where `0` is the match and `1` is the first capture which aligns with PCRE syntax. It would also allow you to name based access. And you could read `captures` if you wanted to use tuple de-structuring. The only weird thing is that once you access `captures`, the indices on the tuple will have shifted down by one.

Documentation/Evolution/StronglyTypedCaptures.md
kylemacomber: Should these notes be moved to subsections of Alternatives Considered?

Documentation/Evolution/StronglyTypedCaptures.md
xAlien95: With `@dynamicMemberLookup`, a captured group labeled `match` would conflict with the `Match` property named `match`. In that regard, `.captures` seems unavoidable.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,560 @@
...
+```
+
+Which would support:
+
+```swift
+let number = line
+    .firstMatch(/([01]+)|([0-9]+)|([0-9A-F]+)/)?
+    .captures
+    .flatMap {
+        switch $0.captures {
```
rxwei, line 472: I think we should go with `optionalProjection` as we discussed earlier, or some better name instead of `captures` for this property. Because `Alternation` is already captures. It's not immediately clear what we are doing by calling `captures` on something that already represents captures.

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Added!

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,621 @@
...
+
+Note that capturing collections of repeated captures like this is a departure
+from most regular expression implementations, which only provide access to the
+_last_ match of a repeated capture group. For example, Python only captures the
+last group in this dash-separated string:
+
+```python
+rep = re.compile(‘(?:([0-9a-f]+)-?)+’)
+match = rep.match(“1234-5678-9abc-def0”)
+print(match.group(1))
```
kylemacomber, line 376:
```suggestion
rep = re.compile('(?:([0-9a-f]+)-?)+')
match = rep.match("1234-5678-9abc-def0")
print(match.group(1))
# Prints "def0"
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,621 @@
...
+
+By contrast, the proposed Swift version captures all four sub-matches:
+
+```swift
+// Note: Syntax / API still in discussion
+let pattern = /(?:([0-9a-f]+)-?)+/
+if let match = “1234-5678-9abc-def0".firstMatch(of: pattern) {
+    print(match.captures)
+}
+// Prints [“1234”, “5678", “9abc”, “def0"]
```
kylemacomber, line 387:
```suggestion
let pattern = /(?:([0-9a-f]+)-?)+/
if let match = "1234-5678-9abc-def0".firstMatch(pattern) {
    print(match.captures)
}
// Prints ["1234", "5678", "9abc", "def0"]
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,621 @@
...
+With variadic generics we think we can define the following `Alternation` type.
+
+```swift
+@dynamicMemberLookup
+public struct Alternation<Captures> { ... }
+
+extension<Option...> Alternation where Captures == (Option...) {
+    public var options: (Option?...) { get }
+    
+    public subscript<T>(dynamicMember keyPath: KeyPath<(T?...), T>) -> T {
```
kylemacomber, line 476:
```suggestion
    public subscript<T>(dynamicMember keyPath: KeyPath<(Option?...), T>) -> T {
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,621 @@
...
+optionals. The Standard Library will provide the runtime guarantee that only one
+element of the `options` tuple will be non-nil. The programmer can use a
+`switch` statement to pattern-match the tuple and handle each case, or directly
+access the tuple's properties via key path dynamic member lookup.
+
+```swift
+let number = line
+    .firstMatch(/([01]+)|([0-9]+)|([0-9A-F]+)/)?
+    .captures
+    .map {
```
kylemacomber, line 492: These were `flatMap` because `Int(_:radix:)` returns an `Int?`, so using just `map` on an `Optional` results in `Int??`

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,621 @@
...
+
+```swift
+/[a-z]{5}/     // => Regex<(5 x Substring)> (exact count)
+/[a-z]{5, 8}/  // => Regex<[Substring]>     (bounded count) 
+/[a-z]{5,}/    // => Regex<[Substring]>     (lower-bounded count)
+```
+
+However, this would cause an inconsistency between exact-count quantification
+and bounded quantification.  We believe that the proposed design will result in
+much less surprises as we associate the `{...}` quantifier syntax with `Array`.
```
kylemacomber, line 571:
```suggestion
fewer surprises as we associate the `{...}` quantifier syntax with `Array`.
```

kylemacomber (commented):
Great edits! Only minor nits.

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Ah, I had assume `init(_:radix:)` returned an non-optional. Didn't look it up.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,622 @@
...
+    .firstMatch(/([01]+)|([0-9]+)|([0-9A-F]+)/)?
+    .captures
+    .flatMap {
+        switch $0 {
+        case let .0(binary):
+            return Int(binary, radix: 2)
+        case let .1(decimal):
+            return Int(decimal, radix: 10)
+        case let .2(hex):
+            return Int(decimal, radix: 16)
```
xAlien95, line 444:
```suggestion
            return Int(hex, radix: 16)
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+```swift
+let pattern = /(?:([0-9a-f]+)-?)+/
+if let match = “1234-5678-9abc-def0".firstMatch(of: pattern) {
+    print(match.captures)
+}
+// Prints [“1234”, “5678", “9abc”, “def0"]
+```
+
+Despite the deviation from prior art, we believe that the proposed capture
+behavior leads to better consistency with the meaning of these quantifiers.
```
milseman, line 393: But we can also provide a compatibility projection. BTW, do we know why they only provide the last one? Do we want that kind of memory savings or does that simplify our execution strategy?

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+
+Nested captures follow the algebra previously described.
+
+```swift
+/([01]+|[0-9]+|[0-9A-F]+)/
+// => `Regex<Substring>`
+/(([01]+)|([0-9]+)|([0-9A-F]+))/
+// => `Regex<(Substring, Alternation<(Substring, Substring, Substring))>>`
+/(?<overall>(?<binary>[01]+)|(?<decimal>[0-9]+)|(?<hex>[0-9A-F]+))/
+// => `Regex<(overall: Substring, Alternation<(binary: Substring, decimal: Substring, hex: Substring))>>`
```
milseman, line 428: I suspect the name `Alternation` won't survive the first 5 posts in the discussion.

milseman (commented):
Quick drive-by comments

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+```
+pattern ::= '(' pattern ')'
+```
+
+```swift
+let graphemeBreakLowerBound = /([0-9a-fA-F]+)/ // => `Regex<Substring>`
+
+// Result builder DSL equivalent
+//     let graphemeBreakLowerBound = OneOrMore(CharacterClass.hexDigit).capture()
+//     // `.Captures == Substring`
```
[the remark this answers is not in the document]
rxwei: Result builder types won't return a `Regex`. They are their own types conforming to `RegexProtocol`.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+## Introduction
+
+Capturing groups are a commonly used component of regular expressions as they
+allow the programmer to extract information from matched input. A capturing
+group collects multiple characters together as a single unit that can be
+[backreferenced](https://www.regular-expressions.info/backref.html) within the
+regular expression and accessed in the result of a successful match. For
+example, the following regular expression contains the capturing groups `(cd*)`
+and `(ef)`.
+
```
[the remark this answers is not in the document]
rxwei: I have the following paragraph in the beginning of the pitch:

> This focus of this pitch is the structural properties of capture types and how
> regular expression patterns compose to form new capture types. The semantics of
> string matching, its effect on the capture types (i.e. `UnicodeScalarView` or
> `Substring`), or the matching API will be discussed in a separate pitch.

I'll mention the result builder DSL here, too.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+//         "ab"
+//         Regex {
+//             "c"
+//             Repeat("d")
+//         }.capture()
+//         "ef".capture()
+//         "gh"
+//     }
+
+if let match = "abcddddefgh".firstMatch(regex) {
```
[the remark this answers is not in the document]
rxwei: See https://github.com/apple/swift-experimental-string-processing/pull/35#discussion_r743879343

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+#### Capturing group
+
+In regular expression literals, a capturing group is a pattern wrapped by a pair
+of parentheses. A capturing group's capture type is reflected in the result
+type's `Capture` generic argument. A capturing group's corresponding capture type
+is `Substring`.
+
+```
+pattern ::= '(' pattern ')'
+```
```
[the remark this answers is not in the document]
rxwei: Now that I think about it, the grammar could have been included in the regex literals pitch. Is there a reason that was not included?

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+//     }
+//     // `.Captures == (Substring, Substring)`
+```
+
+#### Named capturing group
+
+A named capturing group in a pattern with multiple captures causes the resulting
+tuple to have a tuple element label at the corresponding capture type position.
+When the pattern has only one capture, there will be no tuple element label
+because there are no 1-element tuples.
```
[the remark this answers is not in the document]
rxwei: It is discussed as the future directions section.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+group collects multiple characters together as a single unit that can be
+[backreferenced](https://www.regular-expressions.info/backref.html) within the
+regular expression and accessed in the result of a successful match. For
+example, the following regular expression contains the capturing groups `(cd*)`
+and `(ef)`.
+
+```swift
+// Literal version.
+let regex = /ab(cd*)(ef)gh/
+// => `Regex<(Substring, Substring)>`
```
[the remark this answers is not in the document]
rxwei: I'm pretty sure we don't have to store `Substring` or any other views in the implementation. It can be all ranges until the point `.captures` is called to access the captures.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+let graphemeBreakLowerBound = /(?<lower>[0-9A-F]+)\.\.[0-9A-F]+/
+// => `Regex<Substring>`
+
+let graphemeBreakRange = /(?<lower>[0-9A-F]+)\.\.(?<upper>[0-9A-F]+)/
+// => `Regex<(lower: Substring, upper: Substring)>`
+```
+
+#### Non-capturing group
+
+A non-capturing group's capture type is the capture type of its underlying pattern.
```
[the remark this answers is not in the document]
rxwei: Reworded:
> A non-capturing group's capture type is the capture type of its underlying
> pattern. That is, it does not capture anything by itself, but transparently
> propagate its underlying pattern's captures.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,647 @@
...
+
+| Syntax               | Description           | Capture type                                                  |
+| -------------------- | --------------------- | ------------------------------------------------------------- |
+| `*` / `*?`           | 0 or more             | `Array` of the sub-pattern capture type                       |
+| `+` / `+?`           | 1 or more             | `Array` of the sub-pattern capture type                       |
+| `?` / `??`           | 0 or 1                | `Optional` of the sub-pattern capture type                    |
+| `{n}` / `{n}?`       | Exactly _n_           | `Array` of the sub-pattern capture type                       |
+| `{n, m}` / `{n, m}?` | Between _n_ and _m_   | `Array` of the sub-pattern capture type                       |
+| `{n,}` / `{n,}?`     | _n_ or more           | `Array` of the sub-pattern capture type                       |
+
```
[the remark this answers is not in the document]
rxwei: Hmm here I just wanted to give a clear association between the syntax and the corresponding typing.

Documentation/Evolution/StronglyTypedCaptures.md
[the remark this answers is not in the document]
kylemacomber: After further discussion, I do think we're going to elaborate on the type signature for firstMatch in this proposal.

Documentation/Evolution/StronglyTypedCaptures.md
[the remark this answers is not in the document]
kylemacomber: I think Michael's saying the addition of the lazy quantifiers is really emphasizing the regex syntax. It might be just as clear if in the table we omit the lazy quantifier variants.

Documentation/Evolution/StronglyTypedCaptures.md
kylemacomber: `Alternation` was used in the result builder syntax in the overview pitch. In that thread, someone suggested `OneOf` instead. Though there's probably needs to be a difference between the type that represents the capture and the type to create the scoped alternation in the result builder.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,684 @@
...
+```
+
+```swift
+let graphemeBreakLowerBound = /([0-9a-fA-F]+)/ // => `Regex<Substring>`
+
+// Result builder DSL equivalent
+//     let graphemeBreakLowerBound = OneOrMore(CharacterClass.hexDigit).capture()
+//     // `.Captures == Substring`
+```
+
```
milseman, line 175: How do you feel about this alternate formulation? (rendered version of above commit suggestion)

#### Capturing group: `(...)`

In regular expression literals, a capturing group saves the portion of the input matched by its contained regex. A capturing group's `Capture` type is `Substring`.

```swift
let graphemeBreakLowerBound = /([0-9a-fA-F]+)/ // => `Regex<Substring>`

// Result builder DSL equivalent
//     let graphemeBreakLowerBound = OneOrMore(CharacterClass.hexDigit).capture()
```

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,684 @@
...
+```
+pattern ::= '(' pattern ')'
+```
+
+```swift
+let graphemeBreakLowerBound = /([0-9a-fA-F]+)/ // => `Regex<Substring>`
+
+// Result builder DSL equivalent
+//     let graphemeBreakLowerBound = OneOrMore(CharacterClass.hexDigit).capture()
+//     // `.Captures == Substring`
```
milseman, line 173:
```suggestion
#### Capturing group: `(...)`

In regular expression literals, a capturing group saves the portion of the input matched by its contained regex. A capturing group's `Capture` type is `Substring`.

```swift
let graphemeBreakLowerBound = /([0-9a-fA-F]+)/ // => `Regex<Substring>`

// Result builder DSL equivalent
//     let graphemeBreakLowerBound = OneOrMore(CharacterClass.hexDigit).capture()
```

milseman (commented):
I'm really excited by the pitch and its content. I do want to help guide the reader to the crux of this pitch. The syntactic references are great, I find the result builder DSL comparison helpful, but I think a few presentation tweaks can help.

If this isn't trying to pitch the formal literal syntax, let's try a more casual syntax presentation style akin to what's in [pcresyntax](https://www.pcre.org/original/doc/html/pcresyntax.html#TOC1) or Oniguruma's [docs](https://github.com/kkos/oniguruma/blob/master/doc/RE#L212).

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Good suggestion!  I'll update it.

Documentation/Evolution/StronglyTypedCaptures.md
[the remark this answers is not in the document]
milseman: We needed to get the pitch out quickly, and the full grammar is actually pretty massive. It sufficed to say "interior syntax is PCRE" and vaguely imply we're also unioned with UTS18, ICU, and perhaps Oniguruma. We do need a followup or a second draft though.

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -0,0 +1,684 @@
...
+```
+
+| Syntax               | Description           | Capture type                                                  |
+| -------------------- | --------------------- | ------------------------------------------------------------- |
+| `*` / `*?`           | 0 or more             | `Array` of the sub-pattern capture type                       |
+| `+` / `+?`           | 1 or more             | `Array` of the sub-pattern capture type                       |
+| `?` / `??`           | 0 or 1                | `Optional` of the sub-pattern capture type                    |
+| `{n}` / `{n}?`       | Exactly _n_           | `Array` of the sub-pattern capture type                       |
+| `{n, m}` / `{n, m}?` | Between _n_ and _m_   | `Array` of the sub-pattern capture type                       |
+| `{n,}` / `{n,}?`     | _n_ or more           | `Array` of the sub-pattern capture type                       |
```
milseman, line 312: Quantification is more complex if we're talking about syntax or capabilities (e.g. possessive quantification, whether `{n}?` means anything anymore, etc.). The key point is that all of this is entirely irrelevant to this pitch.

---

#### Quantification: `*`, `+`, `?`, `{n}`, `{n,}`, `{n,m}`

A quantifier wraps its underlying pattern's capture type in either an `Optional` or `Array`. Zero-or-one quantification (`?`) produces an `Optional` and all others produce an `Array`. The kind of quantification, i.e. greedy vs reluctant vs possessive, is irrelevant to determining the capture type.

...

milseman (commented):
Quick drive-by cleanup of the quantifiers section. You can still have a reduced table or listing, but I definitely want to lessen the focus on syntax.

Documentation/Evolution/StronglyTypedCaptures.md
milseman: FWIW, the token representation from my [draft](https://github.com/apple/swift-experimental-string-processing/pull/40/files#diff-21e1fabf61ddc50761ff35e58d9df816058df5224058a64ffb93ad4f7ae7c58cR162):

```swift
  enum Quantifier {
    case zeroOrMore(Kind)   // *,     *?,     *+
    case oneOrMore(Kind)    // +,     +?,     ++
    case zeroOrOne(Kind)    // ?,     ??,     ?+
    case exactly(Int, Kind) // {n},   {n}?,   {n}+
    case nOrMore(Int, Kind) // {n,},  {n,}?,  {n,}+
    case upToN(Int, Kind)   // {,n},  {,n}?,  {,n}+
    case range(             // {n,m}, {n,m}?, {n,m}+
      atLeast: Int, atMost: Int, Kind)

    enum Kind {
      case greedy     //
      case reluctant  // ?
      case possessive // +
    }
  }
```

PCRE doesn't have `{,n}`, and reluctant or possessive quantification with `{n}` and variants gets weird pretty quickly. [Oniguruma](https://github.com/kkos/oniguruma/blob/master/doc/RE#L154) doesn't typically support `{n}?` as meaning reluctant, as that's kinda meaningless, unless you specify/configure with `ONIG_SYNTAX_JAVA` or `ONIG_SYNTAX_PERL`. Otherwise, the `?` changes meaning from reluctant to zero-or-one. Similarly for some of the possessives.

We really don't want to wade into this in this pitch :-). CC @hamishknight

Documentation/Evolution/StronglyTypedCaptures.md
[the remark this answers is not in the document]
rxwei: Resolved in Kyle's commit 44186a7.

Documentation/Evolution/StronglyTypedCaptures.md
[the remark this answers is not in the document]
rxwei: Removed formal grammar as suggested.

Documentation/Evolution/StronglyTypedCaptures.md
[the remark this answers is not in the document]
rxwei: Removed syntax details as suggested.

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Updated as suggested.  I initially thought about using `(...)` earlier but I didn't because it could be confused with the `.` metacharacter. Formal grammar probably is the clearest way to describe it. Nonetheless, I agree `(...)` helps reduce the emphasis on syntax.

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Updated as suggested.

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: Current I'm distinguishing between the builder type and the match result by scoping: `Alternation` and `RegexMatch.Alternation`.  But I think `OneOf` is a better name for the DSL, so I'm leaving the match result type as `Alternation` for now.

Documentation/Evolution/StronglyTypedCaptures.md
rxwei: I added a paragraph about adding a variant of quantifiers for smaller memory footprint as a future direction. d8ccc63

Documentation/Evolution/StronglyTypedCaptures.md
```diff
@@ -221,7 +222,7 @@ matched by its contained pattern. A capturing group's capture type is
 ```swift
 let graphemeBreakLowerBound = /([0-9a-fA-F]+)/ // => `Regex<Substring>`
 
-// Result builder DSL equivalent
+// Equivalent result builder syntax:
```
rxwei, line 225: I know that this should be pretty clear to the reader, but this one is not really using the result builder syntax...  That's why I was using "DSL" to refer to anything that involves types like `OneOrMore`.

rxwei:
In eaff534 I'd like to explore a major alternative design. Here instead of `Regex<Captures>`, we try to align the regex type with PCRE's backreference numbering, with `.0` representing the whole match followed by captures. We rename the generic parameter to `Match`, which represents the entire match result. A capturing group is considered as extending `Match` with an extra substring.
```swift
Regex<(Substring, Substring, foo: Substring, bar: Substring, ...)>
                  ^~~~~~~~~~~~~~~ Captures (`\1`, `\k<foo>`, `\k<bar>`, ...)
       ^~~~~~~~~ Whole matched substring (`\0`)

```

Pros:
- Avoids most (if not all) weirdness of `Void`.
- `firstMatch(of:)` is no longer variadic.
- When the generic parameter is a tuple, tuple element indices align exactly with PCRE backreference numbering.
- When there's only one named capture, the name will still be labeled.  No need for single-element tuples.
- No need to pick one between `Never` and `Void`.

Cons:
- For the result builder DSL, `buildBlock(_:)` will need to "drop first" from type sequences, e.g.
    ```swift
    func buildBlock<R...>(_ components: R...) -> Regex<Substring, #dropFirst(R.Match)...>
    ```
   I'm not sure whether variadic generics can support this easily.

PR #36, opened 2021-11-04
Support creating `Regex` from a string with dynamic captures.

rxwei:
- `DynamicCaptures` is an enum whose values represent a tree of captures at runtime.
- `Regex.init(_: String)` allows the user to create a regular expression from a string with `DynamicCaptures` as the capture type.


milseman (approved):
LGTM

PR #39, opened 2021-11-07
Some more minor lexer refactoring

milseman:


milseman:
@swift-ci please test

milseman:
@hamishknight @natecook1000 I went ahead and merged to make other diffs cleaner, but I'm still interested in reviews.

Sources/Util/MissingUnicode.swift
```diff
@@ -0,0 +1,983 @@
+// MARK: - Missing stdlib API
+
+extension Unicode {
+  public enum Script: String {
```
hamishknight, line 4: We may want to separate out the in-source spellings as UTS#18 has fuzzy matching rules and aliases for properties (e.g ignoring case, whitespace, and `_`, and e.g `Latn` being an alias for `Latin`), but that can be done when we implement the lexing/parsing for them.

hamishknight (commented):
LGTM

Sources/Util/MissingUnicode.swift
milseman: CC @natecook1000, resident expert on character classes and properties.

PR #40, opened 2021-11-07
Prototype a new token representation

milseman:
Just a prototype of a new token representation. This new one nearly supports a superset of PCRE, Oniguruma, and UTS\#18. Omissions are mentioned at the bottom.

The idea is to store and traffic in `TokenStorage` and project a `Token` for ease of use. 

I'm thinking of having the lexer continue to track whether it's in a custom character class, so that it can produce richer tokens. The actual logic of what token to project based on that bit can be separated out, see `projectToken()`.


Sources/Regex/RegexToken2.swift
```diff
@@ -0,0 +1,286 @@
...
+
+    // Oniguruma
+    case trueAnychar = "O"
+    case textSegnemt = "y"
+    case notTextSegment = "Y"
+
+  }
+
+  enum Prop {
+    case gc(Unicode.GeneralCategory)
```
hamishknight, line 143: Just realized this doesn't cover the meta-categories e.g `L` which matches all the letter categories, should we spin off a separate type to handle those?

milseman: I'm not sure how best to model/unify this. The `FlattendedOnigurumaUnicodeProperty` mega enum just puts them all flat. It makes sense to me that lexical analysis can resolve this, though.

Sources/Regex/RegexToken2.swift
```diff
@@ -0,0 +1,286 @@
...
+    case pcreSpecial(PCRESpecialCategory)
+    case script(Unicode.Script)
+    // ... normal properties?
```
natecook1000, line 146: Would you include a separate `Name: String` case here?

Also we'll want a Boolean test and a name/value here.

milseman: Could you elaborate? We might have a `case property(UniProperty, value: String)`, or we might do it based on type/kind. Suggestions welcome

milseman:
`Quantifier` was looking pretty good so I made it the AST representation. The AST here is like `NewToken`, that is it is ideally a simple projection on top of more sophisticated storage that includes source locations, etc. We'll have to see how the design pattern works with indirect enums.

milseman:
If at any point in time someone wants to start building on top of this, let me know and I can whip it into mergeable shape.

Sources/Regex/Capture.swift
```diff
@@ -42,11 +42,10 @@ extension AST {
...
     case let .alternation(child), let .concatenation(child):
       return child.any(\.hasCaptures)
-    case .capturingGroup, .group:
-      return true
-    case let .many(child), let .zeroOrOne(child), let .oneOrMore(child),
-         let .lazyMany(child), let .lazyOneOrMore(child),
-         let .lazyZeroOrOne(child):
+    case let .group(g, child), let .groupTransform(g, child, _):
+      return g.isCapturing || child.hasCaptures
+      || true // wait, why?
```
milseman, line 47: @rxwei I'm doing this to preserve behavior. Two questions:

1) Why does compilation wrap concatenations in non-capturing groups if any child captures?
2) Why does it do it for non-capturing children too?

Sources/Regex/AST.swift
```diff
@@ -0,0 +1,44 @@
...
+
+  /// ... ...
+  indirect case concatenation([AST])
+
+  /// (...)
+  indirect case group(Group, AST)
+
+  /// Group with a registered transform
+  indirect case groupTransform(
+    Group, AST, transform: CaptureTransform)
```
milseman, line 15: CC @rxwei, not sure how you want to best model transforms on the AST. This AST form would be unused by literals probably (unless we do something special with literal syntax).

Sources/Regex/Group.swift
```diff
@@ -0,0 +1,102 @@
+public struct Group: ASTParentEntity {
+  public let name: String?
+  public let kind: Kind
+
```
milseman, line 4: The idea is to also be able to easily store source ranges

Sources/Regex/AST.swift
rxwei: Ideally the pattern DSL should have its own extended AST.  I can refactor this out as when I work on the DSL next.

Sources/Regex/Capture.swift
rxwei: > Why does compilation wrap concatenations in non-capturing groups if any child captures?

It actually wraps concatenations in a _capturing_ group if any child captures, just so that it gets pushed onto the stack and starts accumulating each capturing child's captures.  When we know that it has no captures, we skip that step.

> Why does it do it for non-capturing children too?

Does the above answer answer this question?

milseman: I'm more confused. Prior to this refactoring, my understand is that `group` is non-capturing in the AST.

From the parser:
```swift
    switch lexer.peek()?.kind {
    case .leftParen?:
      lexer.eat()
      var isCapturing = true
      if lexer.tryEat(.question) {
        try lexer.eat(expecting: .colon)
        isCapturing = false
      }
      let child = try parse()
      try lexer.eat(expecting: .rightParen)
      return isCapturing ? .capturingGroup(child) : .group(child)
```

From the compiler:

```swift
    case .group(let child):
      instructions.append(.beginGroup)
      compileNode(child)
      instructions.append(.endGroup)
      return

    case .capturingGroup(let child, let transform):
      instructions.append(.beginCapture)
      compileNode(child)
      instructions.append(.endCapture(transform: transform))
      return

    case .concatenation(let children):
      let childrenHaveCaptures = children.any(\.hasCaptures)
      if childrenHaveCaptures {
        instructions.append(.beginGroup)
      }
      children.forEach { compileNode($0) }
      if childrenHaveCaptures {
        instructions.append(.endGroup)
      }
      return
```

But then in RECode we see something different:

```swift
    /// Begin a capture group.
    case beginGroup

    /// Ends a capture group.
    case endGroup

    /// Begin capturing a portion of the input string.
    case beginCapture

    /// End capturing a portion of the input string, transforming the substring with the specified
    /// transform.
    case endCapture(transform: CaptureTransform? = nil)
```

Sources/Regex/AST.swift
milseman: Ah, that makes sense to me. So would regex's AST be one of potentially many children in it? Should the AST contain side-band capture metadata?

Sources/Regex/Capture.swift
rxwei: Ah, that is a naming inconsistency.  `beginGroup` should have been named something like `beginTuple`. What it does is that it intercepts its children’s captures and joins them into a tuple. `beginCapture` on the other hand is more tied to what a literal capturing group does where it stores a substring matching the child.

Sources/Regex/AST.swift
rxwei: Yes, I think embedding the regex AST in the more general pattern AST would help us decide whether to use a specialized lowering path for a part of the pattern.  Re. capture metadata, I think yes for each leaf’s type information; the combined type structure can be inferred from the tree.

milseman:
@hamishknight can you do a pass over `Token2.swift` to see if there's anything in there that we don't currently support? Then I think I can close this  fully.

hamishknight:
I think we have everything from `Token2.swift`

PR #44, opened 2021-11-12
Add Quantifer and Group to AST

milseman:
Add in richer Quantifier and Group types to the AST.

Update the Lexer design (partially), as the parser most of
the time knows what kinds of tokens to be pulling in.

### DIFF
--- a/Sources/Regex/ASTPrinting.swift
+++ b/Sources/Regex/ASTPrinting.swift
@@ -1,0 +1,65 @@
+
+// MARK: - Printing
+
+/// AST entities can be pretty-printed or dumped
+///
+/// Alternative: just use `description` for pretty-print
+/// and `debugDescription` for dump
+public protocol _ASTPrintable:
+  CustomStringConvertible,
+  CustomDebugStringConvertible
+{
+  func _print() -> String
+  func _dump() -> String
+}
+extension _ASTPrintable {
+  public var description: String { _print() }
+  public var debugDescription: String { _dump() }
+}
+
+public protocol _ASTPrintableNested: _ASTPrintable {
+  func _printNested(_ child: String) -> String
+  func _dumpNested(_ child: String) -> String
+}
+extension _ASTPrintableNested {
+  public func _print() -> String { _printNested("") }
+  public func _dump() -> String { _dumpNested("") }
+}
+
+public protocol ASTEntity: _ASTPrintable, Hashable {
+}
+
+public protocol ASTParentEntity: ASTEntity, _ASTPrintableNested {
+  // TODO: variadic access to children?
+}
+
+// MARK: - AST's conformance
+
+extension AST {
+  public func _print() -> String {
+    // TODO: printable version?
+    _dump()
+  }
+
+  public func _dump() -> String {
+    switch self {
+    case let .alternation(rest): return ".alt(\(rest))"
+    case let .concatenation(rest): return ".concat(\(rest))"
+    case let .group(g, rest):
+      return g._dumpNested(rest._dump())
+    case let .groupTransform(g, rest, transform):
+      return g._dumpNested("""
+        \(rest._dump()), transform: \(
+          String(describing: transform)))
+        """)
+    case let .quantification(q, rest):
+      return q._dumpNested(rest._dump())
+
+    case .character(let c):       return c.halfWidthCornerQuoted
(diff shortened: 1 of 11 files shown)


milseman:
@swift-ci please test

Sources/Regex/Quantifier.swift
```diff
@@ -0,0 +1,103 @@
...
+    case exactly(Int) // {n}
+    case nOrMore(Int) // {n,}
+    case upToN(Int)   // {,n}
+    case range(       // {n,m}
+      atLeast: Int, atMost: Int)
+  }
+
+  public enum Kind: String, ASTEntity {
+    case greedy     = ""
+    case reluctant  = "?"
```
natecook1000, line 25: "reluctant" is much better than "lazy" 👏🏻

natecook1000 (approved):

Sources/Regex/Capture.swift
```diff
@@ -42,11 +42,10 @@ extension AST {
...
     case let .alternation(child), let .concatenation(child):
       return child.any(\.hasCaptures)
-    case .capturingGroup, .group:
-      return true
-    case let .many(child), let .zeroOrOne(child), let .oneOrMore(child),
-         let .lazyMany(child), let .lazyOneOrMore(child),
-         let .lazyZeroOrOne(child):
+    case let .group(g, child), let .groupTransform(g, child, _):
+      return g.isCapturing || child.hasCaptures
+      || true // WIP: preserves old behavior
```
rxwei, line 47: When does `g.isCapturing || child.hasCaptures` not give you the old behavior?

Sources/Regex/Quantifier.swift
```diff
@@ -0,0 +1,103 @@
...
+extension Quantifier {
+  public enum Amount: ASTEntity {
+    case zeroOrMore   // *
+    case oneOrMore    // +
+    case zeroOrOne    // ?
+    case exactly(Int) // {n}
+    case nOrMore(Int) // {n,}
+    case upToN(Int)   // {,n}
+    case range(       // {n,m}
+      atLeast: Int, atMost: Int)
```
rxwei, line 20: Why not use a `ClosedRange`?

Sources/Regex/ASTPrinting.swift
```diff
@@ -0,0 +1,65 @@
+
+// MARK: - Printing
+
+/// AST entities can be pretty-printed or dumped
+///
+/// Alternative: just use `description` for pretty-print
+/// and `debugDescription` for dump
+public protocol _ASTPrintable:
```
rxwei, line 8: If this module is only visible to the compiler (and not to the end developer), any reason we need an underscore for this?

Sources/Regex/Capture.swift
milseman: For non-capturing groups that don't contain capturing groups, that will return false. The old behavior returns true for those. E.g. `a(?:b)c`.

Sources/Regex/ASTPrinting.swift
milseman: It's available to a number of clients:

- The compiler
- The stdlib
- SourceKit (in some TBD form)
- SwiftSyntax (in some TBD form)
- Future: LSP
- Future: 3rd party tools and libraries

Basically, we're writing compilers, which are better designed as libraries. The AST is the frontend's API (or more specifically AST and a corresponding ASTAction protocol of some sort).

It's also underscored because I know for sure this isn't a great design long-term but I like it a little better than `Custom[Debug]StringConvertible` for now. It's easier to change and adapt over time. It's a utility/tooling thing, but we lack access-controlled protocol conformances. This certainly isn't API in the sense that AST is, but might be in the future for pretty-printing and dumping, interface TBD.

I'm good going either way, but I want to start refining our use of `public` in the `Regex` module.

milseman: I haven't thought this through too deeply, so take this with a grain of salt. We're looking at converting this into a functional prototype really soon, integrated with the compiler and everything. I want to start getting clearer on what's a useful shortcut for prototyping and what's explicitly being prototyped or demonstrated. I'm definitely not prototyping AST printing, but I _am_ prototyping/demonstrating our AST. I wonder if a leading underscore convention for this distinction would be helpful.

milseman:
Landing this, I have more stuff to base on top.

PR #46, opened 2021-11-13
Introduce ASTAction protocol, represent source locations in AST, refactoring

milseman:
In order to produce an AST, the parser calls a number of `.init()` or static functions, which are just functions. So, the idea behind `ASTAction` is to eventually parameterize the parser over that set of functions.

Potential conformers:

- AST (obviously)
- Something that emits `builder.buildFoo()` calls
- Something that vends syntax highlighting info
- Something that emits or accumulates diagnostics alongside the resultant AST


milseman:

milseman:
Merging to unblock, still interested in review.

Sources/Regex/AST.swift
```diff
@@ -23,21 +24,11 @@ public enum AST: ASTEntity {
...
-    _ kind: Quantifier.Kind, _ a: AST
-  ) -> AST {
-    .quantification(.oneOrMore(kind), a)
-  }
-  public static func zeroOrOne(
-    _ kind: Quantifier.Kind, _ a: AST
-  ) -> AST {
-    .quantification(.zeroOrOne(kind), a)
-  }
+// Note that we're not yet an ASTEntity, would need to be a struct.
```
hamishknight, line 27: `ASTStorage` is a struct, is this comment outdated?

hamishknight (commented):
LGTM!

Sources/Regex/AST.swift
milseman: Its outdated, yes. There's a big rift between having natural nested enums vs retained source info and storage. On the other hand, it's possible this rift could be handled at the `ASTAction` level and we don't have storage types, just a different backend. WDYT?

PR #48, opened 2021-11-16
Matching engine backend for regex.

rxwei:
Matching engine changes:
- Change predicate closure type to be index-based to support matching unicode scalars.
- Support full match and prefix match modes.
- Minor refactoring for clarity.

Regex changes:
- Lower regex AST to the matching engine bytecode for execution. Add `Compiler` and `Executor` types.
- Change `RECode.withMatchLevel(_:)` to an AST transformation.

Next steps:
- Add support for structured captures.
- Add general purpose ISA (integer ops, comparison) to support count-based quantification.


rxwei:
@swift-ci please test

Sources/MatchingEngine/Builder.swift
```diff
@@ -1,12 +1,12 @@
 import Util
 
-extension Program where Element: Hashable {
-  public typealias Predicate = (Element) -> Bool
+extension Program where Input.Element: Hashable {
+  public typealias Predicate = (Input, Input.Index) -> Input.Index?
```
milseman, line 4: This is the consumer interface. Did you think about `(Input, in: Range<Input.Index>) -> Input.Index?`? That would allow us to avoid slicing problems, at least here.

Sources/MatchingEngine/Registers.swift
```diff
@@ -10,7 +10,7 @@ extension Processor {
     var bools: Array<Bool> // TODO: bitset
 
     // currently, these are static readonly
-    var predicates: Array<(Element) -> Bool>
+    var predicates: Array<(Input, Input.Index) -> Input.Index?>
```
milseman, line 13: Probably want to use a shared typealias, or maybe even a nominal type.

Sources/Regex/ASTTransform.swift
```diff
@@ -0,0 +1,33 @@
+extension AST {
+  public func withMatchLevel(_ level: CharacterClass.MatchLevel) -> AST {
```
milseman, line 2: I think this might be ok for now, but ideally compilation would be parameterized over matching level rather than doing a bunch of tree transforms. Also makes it harder to update or refactor the AST.

Alternatively, we define walkers.

milseman (commented):
Brief pass over, will review compilation in more depth.

Sources/Regex/ASTTransform.swift
rxwei: Agreed. The change in this PR was just refactoring the original `RECode.withMatchLevel(_:)` to work for both the legacy and the matching engine backends.

Sources/MatchingEngine/Builder.swift
rxwei: Good point. Change applied.

Sources/MatchingEngine/Registers.swift
rxwei: Done.

rxwei:
@swift-ci please test

milseman:

Sources/MatchingEngine/Processor.swift
```diff
@@ -115,6 +130,20 @@ extension Processor {
     callStack.removeLast(callStack.count - stackEnd)
   }
 
+  mutating func tryAccept() {
+    switch (currentPosition, matchMode) {
+    // When reaching the end of the match bounds or when we are only doing a
+    // prefix match, transition to accept.
+    case (bounds.upperBound, _), (_, .prefix):
+      state = .accept
+
```
milseman, line 139: For this, do we need to differentiate empty matches here? BTW, how are you thinking about modeling that?

Sources/Regex/Compiler.swift
```diff
@@ -0,0 +1,132 @@
...
+      for component in components.dropLast() {
+        let next = builder.createAddress()
+        builder.buildSave(next)
+        emit(component)
+        builder.buildBranch(to: done)
+        builder.label(next)
+      }
+      emit(components.last!)
+      builder.label(done)
+
```
milseman, line 40: For the more complicated cases, could you provide a comment showing what the generate code looks like?

milseman (approved):
Overall LGTM. We'll want to figure out consistent terminology across components. What you're calling "predicate" we've been calling "consume" as in collection-consumers (and `consumeHook`, but I'm open to better names), rather than something returning a bool. Predicate sounds more like the generalization of a character class membership check, whose result might drive assertions or matching. It might not be worth having _both_ custom predicates and custom consumers though.

Either way, we eventually want to avoid those call-outs on our fast paths, except for when we need to call across a compilation barrier.

rxwei:
> What you're calling "predicate" we've been calling "consume" as in collection-consumers (and consumeHook, but I'm open to better names), rather than something returning a bool.

Thanks for the feedback.  The name "predicate" was already there originally—I just changed its signature to make it support unicode scalar matching, with the intention to rename things in a separate patch.  I now renamed it to `ConsumeFunction`.

Sources/Regex/Compiler.swift
rxwei: Added.

Sources/MatchingEngine/Processor.swift
rxwei: I haven't thought about that yet. Do you have an example test case? I can address this in a separate patch.

rxwei:
@swift-ci please test Linux

milseman:
> Yes, specialized matching is on my list to tackle in the near future.

More fundamentally, I suspect we'll roll out a RISC core _and_ try to make things like any quantification of a single element (`a*+` / `ZeroOrMore("a", .possessive))`), a single instruction. We want to give ourselves the ability to compiler other languages to this byte code and give ourselves extensibility, while acknowledging that matching has a special place in all of this. So too might Swift runtime function calls and metatype lookups, etc.

milseman:
> `+ public struct Program<Input: Collection> where Input.Element: Equatable`

One other note is that I was trying to defer binding the `Collection` type as late as possible. This change is probably necessary, but if we're looking at extending to async streams, swift runtime calls, or other things where a collection doesn't exist, we might end up rearchitecting it a bit. But, that's aspirational at this point.

Binding a collection type would be necessary for consumer-style-extension, but is not necessary for predicate-style extension since predicates are `(Element) -> Bool`. Binding the collection type would also not be necessary for a monitor-style extension, `(inout MatchingEngineState) -> Bool`, where the extension code is capable of interacting with the matching engine and state but doesn't necessarily require a notion of position.

Sources/Util/Formatting.swift
```diff
@@ -62,8 +62,34 @@ extension TracedProcessor {
...
      }
      return "\(registers)\n"
-   }
-  public func formatInput() -> String{
+  }
+
+  public func formatInput() -> String {
+    // String override for printing sub-character information.
+    if !input.indices.contains(currentPosition),
+       let input = input as? StringProtocol {
```
milseman, line 70: In swift 5.5, I get an error on this line:

> error: protocol 'StringProtocol' can only be used as a generic constraint because it has Self or associated type requirements

milseman: This fixes it (warning, commit is from a very broken branch, but should be cherry-pickable for anyone blocked on this if I'm not able to land it myself soon):

https://github.com/milseman/swift-experimental-string-processing/commit/15b3112c8217943db30e6232cfa0c30ab7b6e3fc

rxwei: I think I'm using a newer compiler. Strange that CI didn't catch it.

rxwei: https://github.com/apple/swift-experimental-string-processing/pull/51

Sources/Regex/Legacy/VirtualMachine.swift
```diff
@@ -21,24 +21,29 @@ public struct MatchResult {
   }
 }
 
-/// VMs load RECode and run over Strings.
 public protocol VirtualMachine {
-  /// Declare this VM's motto and general life philosophy
+  associatedtype Program
```
milseman, line 25: So this actually breaks the existential VirtualMachines in Swift 5.5. The only use I found was in exercises, where it's easy enough to just name it TortoiseVM

PR #49, opened 2021-11-17
Quants, Quotes, Caps, and Comments

milseman:
I went ahead and finished up groups and quantification with associated syntax options. I added in quotes and comments too.

This demonstrates some the brokenness of the lexer's current approach, beyond brokenness from custom character classes. It also shows some of the absurdity of trying to model with nested `enum`s.

I have this as a draft PR because it probably makes sense to fix the tokens and lexer instead of piling on hacks.

### DIFF
--- a/Sources/Regex/AST.swift
+++ b/Sources/Regex/AST.swift
@@ -24,7 +24,7 @@ public enum AST: ASTValue, ASTAction {
   case any
   case empty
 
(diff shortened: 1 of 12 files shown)


milseman:
This is superseded by https://github.com/apple/swift-experimental-string-processing/pull/53

PR #51, opened 2021-11-20
Fix 'TracedProcessor.formatInput()' to compile with Swift 5.5.

rxwei:
Swift 5.5 doesn't support PATs so we cannot use `_openExistential` here.


milseman:

milseman (approved):
LGTM

PR #52, opened 2021-11-22
Expand builtin character class parsing

hamishknight:
Move classification of builtin character classes into TerminalClassification.swift, and handle more builtin character class kinds.


milseman:
(merging now, I want to rebase some stuff on top)

PR #53, opened 2021-11-22
Lexer and parser overhaul

milseman:
This is a WIP branch with a total overhaul of lexical analysis, source range tracking, diagnostics reporting, and much of the parser. It includes an AST model of custom character classes, comments, quotes, ranges, etc.

I don't know how long it will be before I have this merge-worthy (hopefully pretty soon), but I want to keep @hamishknight informed on the direction. @hamishknight another possibility is for me to start to pull stuff off into mergeable PRs, or you cn start working on top of this branch. Let me know what makes sense to you.

### DIFF
--- a/Sources/Regex/AST.swift
+++ b/Sources/Regex/AST.swift
@@ -27,8 +27,7 @@ public enum AST: ASTValue, ASTAction {
   // type produced from an AST node.
   case characterClass(CharacterClass)
 
-  case customCharacterClass(
-    CustomCharacterClass.Start, CustomCharacterClass)
+  case customCharacterClass(CustomCharacterClass)
 }
 
 extension AST {
@@ -81,6 +80,26 @@ extension AST {
         return $0.filter(f)
       }
     }
+    func filt(_ cc: CustomCharacterClass) -> CustomCharacterClass {
+      CustomCharacterClass(start: cc.start, members: filt(cc.members))
+    }
+    typealias CCCMember = CustomCharacterClass.Member
+    func filt(_ children: [CCCMember]) -> [CCCMember] {
+      children.compactMap {
+        switch $0 {
+        case let .custom(cc):
+          return .custom(filt(cc))
+        case .range(let lhs, let rhs):
+          guard let filtLHS = f(.atom(lhs)) ? lhs : nil else { return nil }
+          guard let filtRHS = f(.atom(rhs)) ? rhs : nil else { return nil }
+          return .range(filtLHS, filtRHS)
+        case let .atom(atom):
+          return f(.atom(atom)) ? .atom(atom) : nil
+        case let .setOperation(lhsMembers, op, rhsMembers):
+          return .setOperation(filt(lhsMembers), op, filt(rhsMembers))
+        }
+      }
+    }
     switch self {
     case let .alternation(children):
       return .alternation(filt(children))
@@ -88,7 +107,8 @@ extension AST {
     case let .concatenation(children):
       return .concatenation(filt(children))
 
-    case .customCharacterClass: fatalError("TODO")
+    case let .customCharacterClass(cc):
+      return .customCharacterClass(filt(cc))
 
     case let .group(g, child):
       guard let c = child.filter(f) else { return nil }
--- a/Sources/Regex/AST.swift
+++ b/Sources/Regex/AST.swift
@@ -28,6 +28,8 @@ public enum AST: ASTValue, ASTAction {
   case characterClass(CharacterClass)
 
   case customCharacterClass(CustomCharacterClass)
+
+  case empty
 }
 
 extension AST {
@@ -121,7 +123,7 @@ extension AST {
       guard let c = child.filter(f) else { return nil }
       return .quantification(q, c)
 
-    case .characterClass, .any, .trivia, .quote, .atom:
+    case .characterClass, .any, .trivia, .quote, .atom, .empty:
       return f(self) ? self : nil
     }
   }
--- a/Sources/Regex/AST.swift
+++ b/Sources/Regex/AST.swift
@@ -17,14 +17,45 @@ public enum AST: ASTValue, ASTAction {
 
   indirect case quantification(Quantifier, AST)
 
-  case character(Character)
-  case unicodeScalar(UnicodeScalar)
-  case characterClass(CharacterClass)
-  case any
-  case empty
+  case quote(String)
 
-  case trivia
+  case trivia // TODO: track comments
+
+  case atom(Atom)
+
+  // FIXME: This doesn't belong in the AST. It could be a model
+  // type produced from an AST node.
+  case characterClass(CharacterClass)
+
+  case customCharacterClass(
+    CustomCharacterClass.Start, CustomCharacterClass)
 }
+
+extension AST {
+  static var any: AST {
+    .atom(.any)
+  }
+}
+
+extension AST {
+  /// If this has a character class representation, whether built-in or custom, return it.
+  ///
+  /// TODO: Not sure if this the right model type, but I suspect we'll want to produce
+  /// something like this on demand
+  var characterClass: CharacterClass? {
+    switch self {
+    case .customCharacterClass:
+      fatalError("TODO")
+    case let .atom(a): return a.characterClass
+
+    // TODO: remove
+    case let .characterClass(c): return c
+
+    default: return nil
+    }
+  }
+}
+
 
 // Note that we're not yet an ASTEntity, would need to be a struct.
 // We might end up with ASTStorage which projects the nice AST type.
@@ -57,6 +88,8 @@ extension AST {
     case let .concatenation(children):
       return .concatenation(filt(children))
 
+    case .customCharacterClass: fatalError("TODO")
+
     case let .group(g, child):
       guard let c = child.filter(f) else { return nil }
       return .group(g, c)
@@ -69,8 +102,7 @@ extension AST {
       guard let c = child.filter(f) else { return nil }
       return .quantification(q, c)
 
-    case .character, .unicodeScalar, .characterClass,
-        .any, .empty, .trivia:
+    case .characterClass, .any, .trivia, .quote, .atom:
       return f(self) ? self : nil
     }
   }
--- a/Sources/Regex/AST.swift
+++ b/Sources/Regex/AST.swift
@@ -43,8 +43,7 @@ extension AST {
   /// something like this on demand
   var characterClass: CharacterClass? {
     switch self {
-    case .customCharacterClass:
-      fatalError("TODO")
+    case let .customCharacterClass(cc): return cc.modelCharacterClass
     case let .atom(a): return a.characterClass
 
     // TODO: remove
--- a/Sources/Regex/ASTTransform.swift
+++ b/Sources/Regex/ASTTransform.swift
@@ -14,19 +14,23 @@ extension AST {
       return .groupTransform(group, recurse(component), transform: transform)
     case .quantification(let quantifier, let component):
       return .quantification(quantifier, recurse(component))
-    case .character(let c):
-      return .character(c)
-    case .unicodeScalar(let u):
-      return .unicodeScalar(u)
+    case .atom(.char(let c)):
+      return .atom(.char(c))
+    case .atom(.scalar(let u)):
+      return .atom(.scalar(u))
     case .characterClass(var cc):
       cc.matchLevel = level
       return .characterClass(cc)
-    case .any:
-      return .any
-    case .empty:
-      return .empty
-    case .trivia:
-      return .trivia
+
+    case .any, .trivia, .quote: return self
+
+    // FIXME: Do we need to do anything here? Match level is
+    // fundamental to the interpretation of atoms, but not
+    // their representation.
+    case .atom: return self
+
+    case .customCharacterClass:
+      fatalError("TODO")
     }
   }
 }
--- a/Sources/Regex/ASTTransform.swift
+++ b/Sources/Regex/ASTTransform.swift
@@ -5,6 +5,7 @@ extension AST {
       child.withMatchLevel(level)
     }
     switch self {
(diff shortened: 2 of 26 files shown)


Sources/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -0,0 +1,456 @@
...
+        return nil
+      }
+      // TODO: Store customCC in the atom, if that's useful
+
+      // POSIX named set
+      if src.tryEat(sequence: "[[:") {
+        // ...
+        try src.expect(sequence: ":]]")
+        fatalError("TODO: posix named sets")
+      }
```
hamishknight, line 428: POSIX char class atoms are more like `[:xxx:]`, it's just that they're only allowed in custom character classes, but you can have e.g `[ab[:space:]\d[:upper:]cd]`

Sources/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -0,0 +1,456 @@
...
+
+      // Escaped
+      case "\\": return try src.expectEscaped(
+          isInCustomCharacterClass: customCC).value
+
+      // TODO: backreferences et al here?
+
+      case "]":
+        assert(!customCC, "parser should have prevented this")
+        fallthrough
```
hamishknight, line 447: PCRE lets you have e.g `]]]` which are treated as literal, is it expected the parser will guard against this, or should we produce `.char` here?

Sources/Regex/Parse/Parse.swift
```diff
@@ -2,158 +2,190 @@
...
+            | 'u' HexDigit{4}
+            | 'x{' HexDigit{1, 8}
+            | 'x' HexDigit{2}
+            | 'U' HexDigit{8}
+
+
+ // TODO: Hamish, can you look over this?
+ CustomCharClass -> '[' '^'? CCCComponent+ ']'
+ CCCComponent    -> CharSet SetOp CharSet | Atom
+ CCCBinOp        ->
```
hamishknight, line 56: Is this meant to be part of the `CustomCharClass` definition above?

Sources/Regex/Parse/Parse.swift
```diff
@@ -183,19 +215,56 @@ extension Parser {
...
-    let isInverted = lexer.tryEat(.caret)
-    var components: [CharacterSetComponent] = []
-    while !lexer.tryEat(.rightSquareBracket) {
+  mutating func parseCustomCharacterClass(
+    _ start: CustomCharacterClass.Start
+  ) throws -> CustomCharacterClass {
+    typealias Member = CustomCharacterClass.Member
+    var members: Array<Member> = []
+
+    // TODO: Is this a correct/sane associativity? Precedence?
```
hamishknight, line 233: Left association by default makes sense to me (especially for subtraction). We may want to give some operators precedence over others, though it may be worth getting users to spell it out explicitly

Sources/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -0,0 +1,456 @@
...
+  ///
+  mutating func lexQuote() throws -> Value<String>? {
+    try recordLoc { src in 
+      if src.tryEat(sequence: #"\Q"#) {
+        return try src.expectQuoted(endingWith: #"\E"#).value
+      }
+      if src.modernQuotes, src.tryEat("\"") {
+        // TODO: escaped `"`, etc...
+        return try src.expectQuoted(endingWith: "\"").value
+      }
```
hamishknight, line 228: Out of curiosity, would we allow modern quotes in custom character classes? `\Q...\E` are currently allowed and just escape things so `[\Qabc\E]` is equivalent to `[abc]`, but it may be confusing to allow `["abc"]` to be interpreted as `[abc]`.

Sources/Regex/Parse/Parse.swift
```diff
@@ -2,158 +2,190 @@
...
+ Group           -> GroupStart Regex ')'
+
+Custom character classes are a mini-language to their own. We
+support UTS#18 set operators and nested character classes. The
+meaning of some atoms, such as `\b` changes inside a custom
+chararacter class. Below, we have a grammar "scope", that is we say
+"SetOp" to mean "CustomCharactetClass.SetOp", so we don't have to
+abbreviate/obfuscate/disambiguate with ugly names like "CCCSetOp".
+
+TODO: Hamish, can you look this over?
```
hamishknight, line 20: Looks good to me!

Sources/Regex/Parse/Parse.swift
```diff
@@ -183,19 +215,56 @@ extension Parser {
...
+  mutating func parseCustomCharacterClass(
+    _ start: CustomCharacterClass.Start
+  ) throws -> CustomCharacterClass {
+    typealias Member = CustomCharacterClass.Member
+    var members: Array<Member> = []
+
+    // TODO: Is this a correct/sane associativity? Precedence?
+    while true {
+      try source.expectNonEmpty()
+      try parseCCCMembers(into: &members)
```
hamishknight, line 236: I believe this may be at a binary operator token if the previous iteration already lexed a binary operator

hamishknight (approved):
Direction looks good to me!

Sources/Regex/Parse/LexicalAnalysis.swift
milseman: 🤯

Sources/Regex/Parse/LexicalAnalysis.swift
milseman: But those are outside custom char classes, right? Otherwise the first one terminates it, as PCRE doesn't do nesting.

Sources/Regex/Parse/Parse.swift
milseman: Whoops, I think I just ended up doing it instead 😅. That was an older version of `CustomCharClass` above.

Sources/Regex/Parse/Parse.swift
milseman: Alrighty, I will consider this concern deferred to you :-)

Sources/Regex/Parse/LexicalAnalysis.swift
hamishknight: Ah whoops sorry I totally misread the assertion here :)

Sources/Regex/Parse/LexicalAnalysis.swift
milseman: Hmm, another interpretation could be that they spell a single Character, like `"a\u{301}"`.

Outside of a character class, adjacency means concatenation and inside it's listing set members, so PCRE's quote behavior does make sense. When we add in grapheme-semantics, these things get a lot murkier.

(noting that the "modern syntax" stuff is just experimental) I wouldn't be surprised if we try to restrict it to Character literals, and since we're in Swift and can do grapheme breaking, we could finally issue a static error if it's not a single grapheme cluster.

Sources/Regex/Parse/Parse.swift
milseman: Ah, I was thinking it would work out if `parseCCCMembers` was like `(!SetOp .)*`, that is it could consume zero or more non-set-operators. But I think I can simplify the loop logic a bit, since once we have a set op we're only ever iterating with the RHS.

hamishknight: I went ahead and simplified this a bit, could you double check it's what you had in mind?

Sources/Regex/Atom.swift
```diff
@@ -404,19 +404,15 @@ extension Atom: _ASTPrintable {
...
-
     case .decimalDigit:    return .digit
     case .notDecimalDigit: return .digit.inverted
 
     case .horizontalWhitespace: return .horizontalWhitespace
     case .notHorizontalWhitespace:
       return .horizontalWhitespace.inverted
 
-    case .notNewline: fatalError("TODO")
+    case .notNewline: return .newlineSequence.inverted
```
milseman, line 411: Just to double check, there's `\R`, `\n`, and `\N`. Do we handle all 3? Is `\N` the same as `\R` inverted?

Sources/Regex/AST.swift
```diff
@@ -81,14 +80,35 @@ extension AST {
         return $0.filter(f)
       }
     }
+    func filt(_ cc: CustomCharacterClass) -> CustomCharacterClass {
```
milseman, line 82: FWIW, I think we're going to want to get away from tree transformations and treat the AST as just a nice view of what's there.

Sources/Regex/Atom.swift
```diff
@@ -450,4 +450,22 @@ extension Atom {
...
   }
+
+  /// Retrieve the character value of the atom if it represents a literal
+  /// character or unicode scalar, nil otherwise.
+  var literalCharacterValue: Character? {
+    switch self {
+    case .char(let c):
+      return c
+    case .scalar(let s):
+      return Character(s)
```
milseman, line 457: I'm a little worried this won't work in grapheme semantic mode. We want to permit a sequence of literal scalars that identify a single grapheme cluster.

Sources/Regex/CustomCharClass.swift
```diff
@@ -1,14 +1,18 @@
...
 
   public enum Member: Hashable {
+    /// A nested custom character class `[[ab][cd]]`
     case custom(CustomCharacterClass)
+    /// A character range `a-z`
     case range(Atom, Atom)
+    /// A single character or escape
     case atom(Atom)
+    /// A binary operator applied to sets of members `abc&&def`
+    case setOperation([Member], SetOp, [Member])
```
milseman, line 15: This all looks much nicer! It's also easy to extend with source ranges, I like it!

Sources/Regex/Atom.swift
hamishknight: Yes we handle all 3, and I believe `\N` is indeed `\R` inverted, or at least it is if we're recognizing all the unicode line endings as characters not matched by `.` when not in multi-line mode

Sources/Regex/Atom.swift
hamishknight: Ah good point

Sources/Regex/Atom.swift
milseman: But that's contingent on semantic matching mode as well as options, right? Is `\N` the inverse of `\R` or `\n`?

Sources/Regex/Atom.swift
milseman: Later, this is a semantic concern so it's weird for it to have any presence in the AST or parser. An AST can produce a model type given a set of parameters. It's possible those parameters come in from the literal itself as options, in which case the parser stores set of options, or it's parameters to the function that produces the model type. The latter is the more general formulation anyways, so I'd go that way.

milseman:
Ok, all but 2 tests are now passing. Those are some of the recent legacy tests over the new matching engine, and it seems like we parse it identically. So not sure what's going on there. We can comment/XFAIL and merge at any point.

Sources/Regex/Atom.swift
hamishknight: Hmm yeah it seems the newlines we recognize would be affected by the `(*CR)`, `(*LF)`, `(*CRLF)`, `(*ANYCRLF)`, `(*ANY)` and `(*NUL)` sequences. PCRE also appears to default to recognizing only `\n`, at least if a convention isn't provided to `pcre2_compile`. Really we'd want to track the newline convention we're matching against for the `any` character class, though perhaps for now we should just leave it as the inverse of `\n`?

milseman: We will have options that govern what is in a character class, at a semantic level, but that shouldn't affect parsing or really code in the parser (outside of us parsing the options and/or faithfully ferrying them along). IIRC, `\N` is normally the inverse of `\n`, but we can treat inverses as distinct classes syntactically.

Sources/Regex/ASTTransform.swift
```diff
@@ -14,19 +15,31 @@ extension AST {
...
       cc.matchLevel = level
       return .characterClass(cc)
-    case .any:
-      return .any
-    case .empty:
-      return .empty
-    case .trivia:
-      return .trivia
+
+    case .any, .trivia, .quote, .empty: return self
```
milseman, line 26: @rxwei would there be any difference for `.any` to return `self` vs a `.any`? I'm not sure how this was switching any-grapheme to any-scalar.

I'm failing inside `testLegacyMatchLevel`, where both the grapheme-semantic AST and the scalar-semantic AST are identical because it's just the `..` regex. . The match level is otherwise not associated with the program AFAICT, and thus the compiled programs are identical between scalar and grapheme-semantic levels.

milseman: Ok, it's because it's above the below check to convert character classes....

milseman:

milseman:
@hamishknight  I believe this is mergeable (even though there's still some fatal errors, existing tests don't hit them). WDYT?

Sources/Regex/Atom.swift
```diff
@@ -0,0 +1,474 @@
...
+
+    /// \y
+    case textSegment
+
+    /// \Y
+    case notTextSegment
+  }
+}
+
+extension Atom.EscapedBuiltin {
```
milseman, line 191: I'm not convinced we want this nesting vs just flattening this into `Atom` and have a computed property for `isEscaped`. Flatter enums are generally better, and I like having the `atom` case in the AST to contrast with the `indirect` cases, etc.

milseman (commented):
We'll also want to drop the old lexer and convert tests over.

hamishknight:
> I believe this is mergeable (even though there's still some fatal errors, existing tests don't hit them). WDYT?

Good with me!

milseman:

rxwei:
IMHO it would've make a cleaner commit history on `main` if these commits were squashed before merging. WDYT?

PR #54, opened 2021-11-29
Fix compilation with Swift 5.5.

rxwei:
Not use `VirtualMachine` as an existential in `RegexConsumer` given that it's being initialized as a concrete `TortoiseVM` anyway.


rxwei:
@swift-ci please test linux

PR #56, opened 2021-12-02
Remove CharacterClass from AST

milseman:


Sources/RegexDSL/DSL.swift
```diff
@@ -23,7 +23,10 @@ extension CharacterClass: RegexProtocol {
   public typealias Capture = Empty
 
   public var regex: Regex<Capture> {
-    .init(ast: .characterClass(self))
+    guard let ast = self.makeAST() else {
+      fatalError("FIXME: extended AST?")
```
milseman, line 27: @rxwei this works for prototyping, but at some point we want a result builder that can run custom code. Have you had thoughts about how this would work? Extend the AST, or an extended AST, or something else entirely?

Sources/Regex/CharacterClass.swift
```diff
@@ -242,3 +242,54 @@ extension CharacterClass: CustomStringConvertible {
     return "\(isInverted ? "not " : "")\(cc)"
   }
 }
+
+extension CharacterClass {
+  public func makeAST() -> AST? {
```
rxwei, line 247: Should it ever return nil?

Sources/RegexDSL/DSL.swift
rxwei: I think the builder DSL will have it's own AST, with embedded Regex AST as a case when applicable.

Sources/Regex/CharacterClass.swift
milseman: Hmm, yes, I think it could be a total function. Perhaps it makes sense to stick this with the DSL code, because it's mostly just a way to map to the "regular" (cough) AST. I'm not sure how short-term this mapping is, but in theory we could map or reconstruct custom char classes.

Given that the custom char class representation is kinda the same as the AST's, what do @natecook1000 and @hamishknight think about just storing the AST inside there and interpreting it?

milseman:
Ah, this is actually blocking some of the AST refactorings. Merging now. We'll want to consider the DSL AST more.

milseman:

PR #58, opened 2021-12-07
Parse \N{...} and \P{...}

hamishknight:
Parse character name and character property syntax. This handles the following kinds of character properties:
- General category e.g `Lu`, `C`
- Binary properties e.g `Alphabetic`
- Script and script extensions e.g `Greek`
- `Any`, `Assigned` & `Ascii`

This covers the properties required by [UTS#18 Level 1][Level1], and uses the loose matching specified by [UAX44-LM3][MatchingRules], including handling of aliases defined by Unicode.

This also covers a good chunk of the properties supported by Oniguruma, with the exception of the `In_...` properties it supports. For now, I've left it as a separate case that gets parsed without loose matching, but we may want to eventually fold them into some of the other cases as aliases. There are also a couple of PCRE special properties such as `Xps` that get parsed this way.

For any properties that aren't currently recognized, they are parsed into an `other` case, as it's possible we may support them in the future.

[Level1]: http://www.unicode.org/reports/tr18/#RL1.2
[MatchingRules]: https://www.unicode.org/reports/tr44/#UAX44-LM3

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/Atom.swift
+++ b/Sources/_MatchingEngine/Regex/AST/Atom.swift
@@ -12,7 +12,7 @@ public enum Atom: Hashable {
   /// A Unicode property, category, or script
   ///
   /// \p{...}, \p{^...}, \P
-  case property(Prop, inverted: Bool)
+  case property(CharacterProperty)
 
   /// A built-in escaped character
   ///
@@ -306,20 +306,51 @@ extension Atom.EscapedBuiltin {
 }
 
 extension Atom {
-  // TODO: Hamish, I believe you have a formulation of this and have
-  // thought through the parsing a whole lot more. This is just what
-  // I have at the time, but let's get something better for the AST
-  // and parser support.
-  public enum Prop: Hashable {
-    case gc(Unicode.GeneralCategory)
-    case pcreSpecial(PCRESpecialCategory)
+  public struct CharacterProperty: Hashable {
+    public var kind: Kind
+    public var isInverted: Bool
+
+    public init(_ kind: Kind, isInverted: Bool) {
+      self.kind = kind
+      self.isInverted = isInverted
+    }
+  }
+}
+
+extension Atom.CharacterProperty {
+  public enum Kind: Hashable {
+    /// Matches any character, equivalent to Oniguruma's '\O'.
+    case any
+
+    // The inverse of 'Unicode.ExtendedGeneralCategory.unassigned'.
+    case assigned
+
+    /// All ascii characters U+00...U+7F
+    case ascii
+
+    /// A general category property.
+    case generalCategory(Unicode.ExtendedGeneralCategory)
+
+    /// Binary character properties. Note that only the following are required
+    /// by UTS#18 Level 1:
+    /// - Alphabetic
+    /// - Uppercase
+    /// - Lowercase
+    /// - White_Space
+    /// - Noncharacter_Code_Point
+    /// - Default_Ignorable_Code_Point
+    case binary(Unicode.BinaryProperty, value: Bool = true)
+
+    /// Character script and script extensions.
     case script(Unicode.Script)
+    case scriptExtension(Unicode.Script)
 
-    // TODO: replace me
-    case propCheck(PropName, PropValue)
+    /// Some special properties implemented by PCRE and Oniguruma.
+    case pcreSpecial(PCRESpecialCategory)
+    case onigurumaSpecial(OnigurumaSpecialProperty)
 
-    // TODO: yuk, let's make sure other cases are a superset of this
-    case oniguruma(FlattendedOnigurumaUnicodeProperty)
+    /// Unhandled properties.
+    case other(key: String?, value: String)
 
     // TODO: erm, separate out or fold into something? splat it in?
     public enum PCRESpecialCategory: String, Hashable {
@@ -329,9 +360,6 @@ extension Atom {
       case universallyNamed = "Xuc"
       case perlWord         = "Xwd"
     }
-
-    public enum PropName: Hashable {}
-    public enum PropValue: Hashable {}
   }
 }
 
--- a/Sources/_MatchingEngine/Regex/Parse/CharacterPropertyClassification.swift
+++ b/Sources/_MatchingEngine/Regex/Parse/CharacterPropertyClassification.swift
@@ -1,0 +1,422 @@
+extension Source {
+  typealias PropertyKind = Atom.CharacterProperty.Kind
+
+  static private func withNormalizedForms<T>(
+    _ str: String, match: (String) -> T?
+  ) -> T? {
+    // This follows the rules provided by UAX44-LM3, including trying to drop an
+    // "is" prefix, which isn't required by UTS#18 RL1.2, but is nice for
+    // consistency with other engines and the Unicode.Scalar.Properties names.
+    let str = str.filter { !$0.isWhitespace && $0 != "_" && $0 != "-" }
+                 .lowercased()
+    if let m = match(str) {
+      return m
+    }
+    if str.hasPrefix("is"), let m = match(String(str.dropFirst(2))) {
+      return m
+    }
+    return nil
+  }
+
+  static private func classifyGeneralCategory(
+    _ str: String
+  ) -> Unicode.ExtendedGeneralCategory? {
+    // This uses the aliases defined in
+    // https://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt.
+    withNormalizedForms(str) { str in
+      switch str {
+      case "c", "other":                   return .other
+      case "cc", "control", "cntrl":       return .control
+      case "cf", "format":                 return .format
+      case "cn", "unassigned":             return .unassigned
+      case "co", "privateuse":             return .privateUse
+      case "cs", "surrogate":              return .surrogate
+      case "l", "letter":                  return .letter
+      case "lc", "casedletter":            return .casedLetter
+      case "ll", "lowercaseletter":        return .lowercaseLetter
+      case "lm", "modifierletter":         return .modifierLetter
+      case "lo", "otherletter":            return .otherLetter
+      case "lt", "titlecaseletter":        return .titlecaseLetter
+      case "lu", "uppercaseletter":        return .uppercaseLetter
+      case "m", "mark", "combiningmark":   return .mark
+      case "mc", "spacingmark":            return .spacingMark
+      case "me", "enclosingmark":          return .enclosingMark
+      case "mn", "nonspacingmark":         return .nonspacingMark
+      case "n", "number":                  return .number
+      case "nd", "decimalnumber", "digit": return .decimalNumber
+      case "nl", "letternumber":           return .letterNumber
+      case "no", "othernumber":            return .otherNumber
+      case "p", "punctuation", "punct":    return .punctuation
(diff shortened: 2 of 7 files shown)


hamishknight:
@swift-ci please test

Sources/Regex/Parse/CharacterPropertyClassification.swift
```diff
@@ -0,0 +1,392 @@
...
+      return .binary(prop, value: isTrue)
+    }
+
+    // This uses the aliases defined in
+    // https://www.unicode.org/Public/UCD/latest/ucd/PropertyAliases.txt.
+    switch normalizeForCharacterPropertyMatching(key) {
+    case "script", "sc":
+      guard let script = classifyScriptProperty(value) else {
+        throw ParseError.invalidScriptKind(value)
+      }
```
hamishknight, line 376: @milseman Any thoughts on whether we should throw an error here, or just fallback to parsing as an `.other` in case new scripts (and maybe general categories?) get added in the future?

Sources/Regex/Atom.swift
```diff
@@ -305,20 +308,51 @@ extension Atom.EscapedBuiltin {
...
+
+    /// Character script and script extensions.
     case script(Unicode.Script)
+    case scriptExtension(Unicode.Script)
 
-    // TODO: replace me
-    case propCheck(PropName, PropValue)
+    /// Some special properties implemented by PCRE and Oniguruma.
+    case pcreSpecial(PCRESpecialCategory)
+    case onigurumaSpecial(OnigurumaSpecialProperty)
```
milseman, line 352: Does Oniguruma have anything special beyond Unicode?

Sources/Regex/Parse/CharacterPropertyClassification.swift
```diff
@@ -0,0 +1,392 @@
+import Util
+
+extension Source {
+  static private func normalizeForCharacterPropertyMatching(_ str: String) -> String {
+    // This follows the rules provided by UAX44-LM3, except the dropping
+    // of an "is" prefix, which UTS#18 RL1.2 states isn't a requirement.
+    return str.filter { !$0.isWhitespace && $0 != "_" && $0 != "-" }
+              .lowercased()
```
milseman, line 8: FWIW, it would be nice to drop the `is` too, since `Unicode.Scalar.Properties` uses `is` and so does API on Character. Not sure if we'd have perfect parity with the other parts of the name, but I don't think it's harmful

Sources/Regex/Parse/CharacterPropertyClassification.swift
milseman: Run time can have a different version of the spec which could be newer. I don't know how we'll handle that kind of availability... We may need to communicate back to the parser and our enum/struct or whatever has availability on it.

milseman (approved):
Overall LGTM

milseman:
You'll need to rebase on the big module overhaul from https://github.com/apple/swift-experimental-string-processing/pull/59.

Sources/Regex/Atom.swift
```diff
@@ -305,20 +308,51 @@ extension Atom.EscapedBuiltin {
...
+    case any
+
+    // The inverse of 'Unicode.ExtendedGeneralCategory.unassigned'.
+    case assigned
+
+    /// All ascii characters U+00...U+7F
+    case ascii
+
+    /// A general category property.
+    case gc(Unicode.ExtendedGeneralCategory)
```
rxwei, line 334: Should we use a more descriptive name, e.g. `general` or `generalCategory`, for readability?  I came across this while reading test cases.

Sources/Regex/Atom.swift
hamishknight: I don't think so, I'm pretty sure the only properties remaining are those for the `Block` Unicode property (though Oniguruma prefixes them with `In_`, presumably as it would otherwise cause ambiguities with the script properties), but I need to go through and double check that. If that is the case, I think we should probably parse out the block properties along with their Unicode aliases and Oniguruma `In_` prefixes (and decide when you need to disambiguate with `Block=`), but for now I figured we should at least parse out the bits that Oniguruma accepts.

hamishknight:
@swift-ci please test Linux

Sources/Regex/Atom.swift
hamishknight: Yeah, changed to `generalCategory`, thanks!

Sources/Regex/Parse/CharacterPropertyClassification.swift
hamishknight: Okay yeah, I've updated to parse unknown cases as `other` for now

Sources/Regex/Parse/CharacterPropertyClassification.swift
hamishknight: Updated to drop `is`, how does it look?

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/AST/Atom.swift
```diff
@@ -303,20 +306,51 @@ extension Atom.EscapedBuiltin {
...
-    // TODO: replace me
-    case propCheck(PropName, PropValue)
+    /// Some special properties implemented by PCRE and Oniguruma.
+    case pcreSpecial(PCRESpecialCategory)
+    case onigurumaSpecial(OnigurumaSpecialProperty)
 
-    // TODO: yuk, let's make sure other cases are a superset of this
-    case oniguruma(FlattendedOnigurumaUnicodeProperty)
+    /// Unhandled properties.
+    case other(key: String?, value: String)
```
milseman, line 353: I kept thinking we'd make the RHS of the `Key=Value` bit optional since that's what's omitted when we're doing a membership test. But I can see it being the other way around. Thoughts on this?

Sources/_MatchingEngine/Regex/Parse/CharacterPropertyClassification.swift
```diff
@@ -0,0 +1,395 @@
...
+  static private func normalizeForCharacterPropertyMatching(
+    _ str: String
+  ) -> String {
+    // This follows the rules provided by UAX44-LM3, including the dropping of
+    // an "is" prefix, which isn't required by UTS#18 RL1.2, but is nice for
+    // consistency with other engines and the Unicode.Scalar.Properties names.
+    let str = str.filter { !$0.isWhitespace && $0 != "_" && $0 != "-" }
+                 .lowercased()
+    if str.hasPrefix("is") {
+      return String(str.dropFirst(2))
```
milseman, line 11: Do we have to check count and/or other things here? Otherwise there might be some conflict with scripts that have `IS` as a prefix.

Sources/_MatchingEngine/Regex/Parse/CharacterPropertyClassification.swift
```diff
@@ -0,0 +1,395 @@
...
+    case "ps", "openpunctuation": return .openPunctuation
+    case "s", "symbol": return .symbol
+    case "sc", "currencysymbol": return .currencySymbol
+    case "sk", "modifiersymbol": return .modifierSymbol
+    case "sm", "mathsymbol": return .mathSymbol
+    case "so", "othersymbol": return .otherSymbol
+    case "z", "separator": return .separator
+    case "zl", "lineseparator": return .lineSeparator
+    case "zp", "paragraphseparator": return .paragraphSeparator
+    case "zs", "spaceseparator": return .spaceSeparator
```
milseman, line 59: Hmmm, I'm not a huge fan of this generally, but I think this source formatting would be improved if the `returns` for the cases (not necessarily for `default`) lined up on the same column. WDYT?

Tests/RegexTests/ParseTests.swift
```diff
@@ -242,7 +242,80 @@ extension RegexTests {
...
+    parseTest(#"\p{isAlphabetic}"#, .atom(prop(.binary(.alphabetic))))
+    parseTest(#"\p{isAlpha=isFalse}"#, .atom(prop(.binary(.alphabetic, value: false))))
+
+    parseTest(#"\p{In_Runic}"#, .atom(prop(.onigurumaSpecial(.inRunic))))
+
+    parseTest(#"\p{Xan}"#, .atom(prop(.pcreSpecial(.alphanumeric))))
+    parseTest(#"\p{Xps}"#, .atom(prop(.pcreSpecial(.posixSpace))))
+    parseTest(#"\p{Xsp}"#, .atom(prop(.pcreSpecial(.perlSpace))))
+    parseTest(#"\p{Xuc}"#, .atom(prop(.pcreSpecial(.universallyNamed))))
+    parseTest(#"\p{Xwd}"#, .atom(prop(.pcreSpecial(.perlWord))))
```
milseman, line 318: I'll be overhauling the AST soon to put source locations everywhere, which would otherwise have huge impact on test cases constructing AST nodes. Try to migrate to https://github.com/apple/swift-experimental-string-processing/blob/main/Sources/_MatchingEngine/Regex/AST/ASTBuilder.swift (and expand upon it if needed).

When the AST settles down, and result builders's usage is clearer, that file can be moved to a test target (or made internal thanks to `@testable`).

Tests/RegexTests/SyntaxOptionsTests.swift
```diff
@@ -12,12 +12,16 @@ private let dotAST = AST.concatenation([
 extension RegexTests {
 
   func testSemanticWhitespace() {
+    func concat(_ asts: AST...) -> AST { return .concatenation(asts) }
+
```
milseman, line 16:
```suggestion
```

milseman (approved):
Overall LGTM!

Sources/_MatchingEngine/Regex/AST/Atom.swift
hamishknight: Hmm I'm not sure, if you have `\p{X}`, `X` is generally a value for which a key can be inferred. The only exception to this is for binary properties where `= True` can be inferred, but IMO it makes more sense in general for the LHS to be optional.

Sources/_MatchingEngine/Regex/Parse/CharacterPropertyClassification.swift
hamishknight: Good catch! There are only a couple of cases of properties beginning with `is` in the spec, and none of them affect the cases implemented so far, but I think we should probably change this to first try matching without dropping the prefix, and then try again with it dropped.

Tests/RegexTests/ParseTests.swift
hamishknight: To be clear, is this comment aimed at at this patch, or just generally for future? I did add a `func prop` to ASTBuilder.swift, should there be more functions that e.g produce `AST` directly?

hamishknight:
@swift-ci please test Linux

hamishknight:
@milseman How does this look? Also dropped the logic that threw an error on an invalid bool spelling, as I suppose it's possible a future spec could add new spellings.

Tests/RegexTests/ParseTests.swift
milseman: Ah, I see. I saw the `.atom` constructing a raw AST node, but it's true that I haven't added atom builders and it doesn't make sense for you to have to do so either. LGTM.

milseman:
SGTM for now, thought I suspect we'll consider adding that error back in specifically. Either way shouldn't hold up this PR.

milseman:

PR #57, opened 2021-12-07
Track source locations through the AST

milseman:
- Source range tracking for groups and quantification
- Overhaul AST to be enum of structs and use protocols for common operations

Still need to track way more source locations, but I think it's important to merge this soon so we can build on it.

PR #59, opened 2021-12-07
Reorganize our modules

milseman:
Ready for review. I'll squash this all into a single commit; rebasing on top of this is probably best done in one (painful) shot.

- `_MatchingEngine` contains the compile time libraries
    - Matching engine, byte code, etc
    - Most utility functionality (e.g. typed ints)
    - Regex parser and AST
- `_StringProcessing` contains runtime libraries, depends on `_MatchingEngine`
    - Model types like `CharacterClass`
    - Compilation (could be moved to `_MatchingEngine` in future)
    - DSL


milseman:
@hamishknight here, and in [the AST source location plumbing PR](https://github.com/apple/swift-experimental-string-processing/pull/57), trying to keep the old Lexer alive and up-to-date is a bit of work. This PR removes the old lexer and those tests.

I believe you were looking at porting the tests over, or rather the few that might need porting.

milseman:

Sources/_StringProcessing/Utility/TypeConstruction.swift
```diff
@@ -99,6 +99,13 @@ public func tupleType<ElementTypes: BidirectionalCollection>(
   }
 }
 
+// FIXME: namespaces are hard
+internal func _tuple<Elements: BidirectionalCollection>(
+  of elements: __owned Elements
+) -> Any where Elements.Element == Any {
+  tuple(of: elements)
+}
+
```
rxwei, line 108: Any reason we need an additional `_tuple(of:)`, instead of using `tuple(of:)` defined below?

Sources/_StringProcessing/CharacterClass.swift
```diff
@@ -284,12 +286,124 @@ extension CharacterClass {
...
+  }
+}
+
+extension CharacterClass {
+  public func withMatchLevel(
+    _ level: CharacterClass.MatchLevel
+  ) -> CharacterClass {
+      var cc = self
+      cc.matchLevel = level
+      return cc
```
rxwei, line 320:
```suggestion
    var cc = self
    cc.matchLevel = level
    return cc
```

Sources/_MatchingEngine/Utility/Peek.swift
```diff
@@ -1,6 +1,6 @@
 // MARK: - _Peekable
 
-protocol _Peekable {
+public protocol _Peekable {
```
rxwei, line 3: It's only used within this module.   Does it need to be public?

Sources/_StringProcessing/Utility/TypeConstruction.swift
milseman: The call site originally stated `Util.tuple` because it was ambiguous and/or shadowed by `Capture.tuple`. Now that it's the same module, we don't have that disambiguation option. It might not be necessary here, I had tried doing this change when the module was still named `Regex` which was particularly unhelpful due to `Regex<Capture>`.

milseman:
```
StringProcessing/Sources/_StringProcessing/Capture.swift:26:33: error: cannot convert value of type '[Any]' to expected argument type '[Capture]'
      return tuple(of: elements.map(\.value))
                                ^
```

rxwei: Also, `tupleType(of:)` and `tuple(of:)` should be made internal since they aren't user-facing.

Sources/_StringProcessing/Capture.swift
```diff
@@ -1,4 +1,6 @@
-import Util
+import _MatchingEngine
+
+// TODO: what here should be in the compile-time module?
 
 public enum Capture {
```
rxwei, line 5: Let's make this internal as well as it's not user-facing.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -1,7 +1,7 @@
-import MatchingEngine
+import _MatchingEngine
 
 public struct RegexProgram {
```
rxwei, line 3: Internal too, as the compiler and executor aren't user-facing.

Sources/_StringProcessing/Utility/TypeConstruction.swift
rxwei: > The call site originally stated Util.tuple because it was ambiguous and/or shadowed by Capture.tuple. Now that it's the same module, we don't have that disambiguation option.

Maybe move `tuple(of:)` and `tupleType(of:)` as static methods under an empty enum `TypeConstruction`?

rxwei: Re. error, we should make `Capture` internal.

Sources/_MatchingEngine/Utility/Peek.swift
milseman: Ah, the `Lexer` was conforming before I deleted that. Will remove

Sources/_StringProcessing/Compiler.swift
milseman: But `Executor` is, and `Compiler` is used by the regex consumer conformance

rxwei:
Remaining public symbols that should probably be made internal:
`TortoiseVM`, `HareVM`, `compile(_: AST, options: REOptions)`, `RECode`, `MatchResult`, `VirtualMachine`

Sources/_StringProcessing/Compiler.swift
milseman: Alternatively, we pull that stuff in to this module too, or else have simple public functions to do compilation etc.

rxwei: Oh... Then I guess we can leave them public for now until we find a way to better organize the algorithms.

Sources/_StringProcessing/Utility/TypeConstruction.swift
milseman: Done

Sources/_StringProcessing/Compiler.swift
milseman: I went ahead and dropped the whole public VirtualMachine thing and TortoiseVM used as a reference. We'll just have to actually test the matching engine going forwards :-)

Sources/Algorithms/Consumers/RegexConsumer.swift
```diff
@@ -15,14 +15,9 @@ public struct Regex {
...
   let vm: Executor
-  let referenceVM: TortoiseVM
 
   public init(regex: Regex) {
-    let ast = try! parse(regex.string, .traditional)
-    let program = Compiler(ast: ast).emit()
-    self.vm = Executor(program: program)
-    let legacyProgram = try! compile(ast, options: regex.options)
-    self.referenceVM = TortoiseVM(program: legacyProgram)
+    self.vm = _compileRegex(regex.string)
```
rxwei, line 20: Can this use the public API (e.g. `firstMatch`) instead?

rxwei: Currently it is `RegexProtocol.match(in:)`. You can add a range parameter to it.

milseman: What would I use for types? It seems like `DynamicCaptures` will let me parse the string, but will trap at runtime:

```swift
  let regex: _StringProcessing.Regex<(DynamicCaptures)>
```

```
Could not cast value of type '()' (0x7fff815b12b8) to '_StringProcessing.DynamicCaptures' (0x10f67b618).
```

rxwei: Ah, this needs the following change:
```diff
--- a/Sources/_StringProcessing/RegexDSL/Core.swift
+++ b/Sources/_StringProcessing/RegexDSL/Core.swift
@@ -85,7 +85,13 @@ extension RegexProtocol {
     guard let result = executor.execute(input: input) else {
       return nil
     }
-    return RegexMatch(range: result.range, captures: () as! Capture)
+    let convertedCapture: Capture
+    if Capture.self == DynamicCaptures.self {
+      convertedCapture = DynamicCaptures.tuple([]) as! Capture
+    } else {
+      convertedCapture = () as! Capture
+    }
+    return RegexMatch(range: result.range, captures: convertedCapture)
   }
 }
```

rxwei: Address this later sounds good too

milseman: Scratch that, I later also get a 

```
Could not cast value of type 'Swift.Array<Swift.Substring>' (0x7fff815ac3d8) to '()' (0x7fff815b12b8).
```

rxwei (approved):

milseman:

PR #60, opened 2021-12-08
Some AST construction and DSL refactoring

milseman:
- Add in `PatternTree`, which holds a formula over an algebra of patterns
- Minor refactorings

Doing the [AST overhaul](https://github.com/apple/swift-experimental-string-processing/pull/57) involves _heavily_ modifying the DSL code, and I realize much of this overhaul can be refactored out and developed in parallel / incrementally. Trying to do it all at once is proving troublesome.

This does some cosmetic changes and starts us down the path towards a pattern algebra. It still falls back to regex AST. I also didn't bother to rename anything fundamental (like `Pattern` vs `Regex`). This is meant to be the minimal changes that help unblock AST reworking.

Next step will be to get `groupTransform` cleanly off the AST and into PatternTree. Compilation should compile a pattern tree. I don't know how or if we want to do a lot of nesting here...

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/ASTBuilder.swift
+++ b/Sources/_MatchingEngine/Regex/AST/ASTBuilder.swift
@@ -1,0 +1,156 @@
+/*
+
+These functions are temporary AST construction helpers. As
+the AST gets more and more source-location tracking, we'll
+want easier migration paths for our parser tests (which
+construct and compare location-less AST nodes) as well as the
+result builder DSL (which has a different notion of location).
+
+Without real namespaces and `using`, attempts at
+pseudo-namespaces tie the use site to being nested inside a
+type. So for now, these are global, but they will likely be
+namespaced in the future if/when clients are weaned off the
+AST.
+
+*/
+
+public let _fakeLoc = "".startIndex
+public let _fakeRange = _fakeLoc ..< _fakeLoc
+
+public func alt(_ asts: [AST]) -> AST {
+  .alternation(asts)
+}
+public func alt(_ asts: AST...) -> AST {
+  alt(asts)
+}
+
+public func concat(_ asts: [AST]) -> AST {
+  .concatenation(asts)
+}
+public func concat(_ asts: AST...) -> AST {
+  concat(asts)
+}
+
+public func group(
+  _ kind: Group.Kind, _ child: AST
+) -> AST {
+  .group(Group(kind, _fakeRange), child)
+}
+public func capture(
+  _ child: AST
+) -> AST {
+  group(.capture, child)
+}
+public func nonCapture(
+  _ child: AST
+) -> AST {
+  group(.nonCapture, child)
+}
+public func namedCapture(
+  _ name: String,
+  _ child: AST
+) -> AST {
+  group(.namedCapture(name), child)
+}
+public func nonCaptureReset(
+  _ child: AST
+) -> AST {
+  group(.nonCaptureReset, child)
+}
+public func atomicNonCapturing(
+  _ child: AST
+) -> AST {
+  group(.atomicNonCapturing, child)
+}
+public func lookahead(_ child: AST) -> AST {
+  group(.lookahead, child)
+}
+public func lookbehind(_ child: AST) -> AST {
+  group(.lookbehind, child)
+}
+public func negativeLookahead(_ child: AST) -> AST {
+  group(.negativeLookahead, child)
+}
+public func negativeLookbehind(_ child: AST) -> AST {
+  group(.negativeLookbehind, child)
+}
+
+public var any: AST { .atom(.any) }
+
+public func quant(
+  _ amount: Quantifier.Amount,
+  _ kind: Quantifier.Kind = .greedy,
+  _ child: AST
+) -> AST {
+  .quantification(Quantifier(amount, kind, _fakeRange), child)
+}
+public func zeroOrMore(
+  _ kind: Quantifier.Kind = .greedy,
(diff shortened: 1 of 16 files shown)


Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -13,39 +11,47 @@ public protocol RegexProtocol {
...
     }()
     /// The program for execution with the matching engine.
-    lazy private(set) var loweredProgram = Compiler(ast: ast).emit()
+    lazy private(set) var loweredProgram = Compiler(
+      ast: regexAST! // FIXME
+    ).emit()
 
-    init(ast: AST) {
-      self.ast = ast
+    init(_ tree: PatternTree) {
```
rxwei, line 40: Nit: I’d prefer keeping an `ast:` or `tree:` argument label since it’s not exactly a value-preserving conversion and it’s better to be able to tell this is calling an internal API at call sites.

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -13,39 +11,47 @@ public protocol RegexProtocol {
...
     }
   }
 
   let program: Program
-  var ast: AST { program.ast }
 
-  init(ast: AST) {
-    self.program = Program(ast: ast)
+  var tree: PatternTree { program.tree }
+  var regexAST: AST? { program.regexAST }
```
rxwei, line 51: If we kept its name `ast` and made it non-optional (it okay to fail for now since it’s to be overhauled again), most of the changes to the gyb’d contents will not be necessary.

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -13,39 +11,47 @@ public protocol RegexProtocol {
...
-extension Optional: EmptyProtocol where Wrapped: EmptyProtocol {}
-
 /// A regular expression.
 public struct Regex<Capture>: RegexProtocol {
   /// A program representation that caches any lowered representation for
   /// execution.
   internal class Program {
-    /// The underlying AST.
-    let ast: AST
+    /// The underlying pattern tree
```
rxwei, line 19:
```suggestion
    /// The underlying pattern tree.
```

Sources/_StringProcessing/RegexDSL/Core.swift
milseman: When might it not be value-preserving? It's weird to have the type name as the label, if there's a value loss, is there a way to describe what could be lost?

Sources/_StringProcessing/RegexDSL/Core.swift
milseman: But we'd still need to change them when we stop using the AST, right?

Sources/_StringProcessing/RegexDSL/Core.swift
rxwei: I imagined there would eventually be invalid trees? If so it would not be a monomorphism, otherwise never mind the value-preserving conversion argument.

That said, IMO the main value of having an argument label is to distinguish between calls to internal API and calls to public API more easily in our tests. I'm imagining there will be call sites like `Pattern(.hexDigit)`, where which API being called would be less distinguishable between that and `Pattern(.regex(...))`. An argument label will describe the argument's role and surely doesn't have to repeat the type information. Something like `init(representing:)` would work, too.

milseman:

Sources/_MatchingEngine/Regex/AST/ASTBuilder.swift
```diff
@@ -0,0 +1,156 @@
...
+result builder DSL (which has a different notion of location).
+
+Without real namespaces and `using`, attempts at
+pseudo-namespaces tie the use site to being nested inside a
+type. So for now, these are global, but they will likely be
+namespaced in the future if/when clients are weaned off the
+AST.
+
+*/
+
```
rxwei, line 16: For namespacing, could you move these top-level symbols to an enum, e.g. `ASTBuilder`, as static methods?

rxwei (approved):

Sources/_MatchingEngine/Regex/AST/ASTBuilder.swift
milseman: Maybe eventually. I'm inclined not to for now because otherwise you end up either with a bunch of `ASTBuilder.concat(ASTBuilder.atom....)` in unit tests, or every client just reinvents these. Since the contextual type requested is AST and not ASTBuilder, they wouldn't be seen for lookup unless everything was typed as a builder instead of an AST. These don't work well hosted on AST because names can collide with case names.

Longer term, I don't think we'll be as enthusiastically creating AST nodes for every need. This is public in the "core" module right now (so visible to _StringProcessing but not its clients). Another option is to make these internal inside _StringProcessing and the test use case can use `@testable` anyways.

Also, we'll see if AST even stays as an enum or in its current form. We might need to overhaul it more for options tracking.

rxwei: >  Since the contextual type requested is AST and not ASTBuilder, they wouldn't be seen for lookup unless everything was typed as a builder instead of an AST. 

Hmm, these could also be static methods on `AST` itself though. I'm still concerned about polluting the global namespace.

> Another option is to make these internal inside _StringProcessing and the test use case can use `@testable` anyways.

I think that this approach works better; we only build location-less ASTs in  _StringProcessing module and its tests.

PR #61, opened 2021-12-10
Create Prototypes target for our prototypes

milseman:
Moves PEG under Prototypes, as it's a real prototype. Also moved in PTCaRet, Combinators (now actually being built!), restored the TourOfTypes (in case that's useful), etc.

So, we'll have:

- `_MatchingEngine`, eventually named something something "core", has compile-time components
- `_StringProcessing` pulls in `_MatchingEngine`  and defines run-time components
- `Prototypes` is where we put prototypes, at least "prototypical" compared to `_StringProcessing`'s status
- `Exercises` is where we put examples, benchmark fodder, etc (still needs more fleshing out!)

@timvermeulen can you do a pass over `Algorithms` and see if it's what we want as API for the holidays? Ideally we'd have `trim` etc available taking regex. We can make it a subdirectory under `_StringProcessing`. If there's anything that's of a more in-progress status that can be a new folder under `Prototypes`.


milseman:

PR #62, opened 2021-12-10
Parse more group, property, and scalar syntax

hamishknight:
Start parsing:

- `\N{U+hh...}`
- Octal sequences `\o{...}` and `\0...`. Sequences spelled as `'\' N` (where `N >= 1`) will be done when parsing backreferences.
- Alternative spellings for groups, including the spelled out PCRE2 names.

In addition, unify character property parsing such that it's shared between POSIX syntax and `\p`, allowing e.g `\p{alnum}` and `[:script=Greek:]`, which are accepted by ICU.


hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -12,143 +12,140 @@ API convention:
...
   }
 
   /// Eat a scalar off the front, starting from after the
   /// backslash and base character (e.g. `\u` or `\x`).
   ///
-  ///     UniScalar -> 'u{' HexDigit{1...8}
+  ///     UniScalar -> 'u{' HexDigit{1...} '}'
   ///                | 'u'  HexDigit{4}
-  ///                | 'x{' HexDigit{1...8}
+  ///                | 'x{' HexDigit{1...} '}'
```
milseman, line 114: CC @natecook1000 

We will likely also want fixits to convert everything to a single canonical notation

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -379,6 +375,43 @@ extension Source {
...
+        }
+        if src.tryEat(sequence: "sr:") || src.tryEat(sequence: "script_run:") {
+          return .scriptRun
+        }
+        if src.tryEat(sequence: "asr:") ||
+            src.tryEat(sequence: "atomic_script_run:") {
+          return .atomicScriptRun
+        }
+      }
+
```
milseman, line 414: In the future, I think this code will be an example of composing consumers into ordered choices.

Sources/_MatchingEngine/Utility/MissingUnicode.swift
```diff
@@ -174,21 +174,21 @@ extension Unicode {
...
-    case space = "space"
-    case upper = "upper"
     case word = "word"
     case xdigit = "xdigit"
+    // As per http://www.unicode.org/reports/tr18/#Compatibility_Properties,
+    // [:alpha:], [:lower:], [:upper:], and [:space:] are covered by binary
+    // properties. [:punct:], [:digit:], and [:cntrl:] are covered by general
+    // categories. [:ascii:] is covered by CharacterProperty.Kind.ascii.
+    // These may have different semantics depending on matching mode, but that
+    // should be left up to the matching engine.
```
milseman, line 191: But, people can still type them in and we'll still parse them right? Wouldn't we want to preserve how they specified it, especially since the interpretation varies based on semantic mode?

milseman (commented):
Just a quick look over, and some quick questions

Sources/_MatchingEngine/Utility/MissingUnicode.swift
hamishknight: Yes we still parse them, and we preserve whether they were written using `[:...:]` or `\p{...}` syntax, which could be used to vary behavior based on matching mode. It's just that by default they will be classified as general categories or binary properties, as Unicode defines aliases for them that cover the POSIX spellings. We could prefer to classify them as POSIX, but IMO it's nicer to have a complete Unicode syntax and add in the bits that it doesn't cover, what do you think? It's also complicated by the fact that the Unicode definitions have aliases for e.g `punct`, `punctuation`, and `p`. It may be that we need to preserve which alias was used, as e.g should only the first get the POSIX behavior?

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -297,7 +291,7 @@ extension Source {
 
   /// Try to consume a comment
   ///
-  ///     Comment -> '(?#.' [^')']* ')'
+  ///     Comment -> '(?#' [^')']* ')'
```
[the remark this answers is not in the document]
hamishknight: I think that's just there to denote the contents of the comment, http://pcre.org/current/doc/html/pcre2pattern.html#SEC24 states that:

> The sequence `(?#` marks the start of a comment that continues up to the next closing parenthesis

Sources/_MatchingEngine/Utility/MissingUnicode.swift
kylemacomber: The latest Nate and I talked, we were thinking that `[:...:]` character classes would be restricted to their traditional ASCII definition regardless of semantic mode, `\p{...}` character classes would be restricted to their traditional Unicode definition regardless of semantic mode, and `\w` style character classes would be more flexible binding to ASCII, Unicode, or grapheme cluster definitions depending on semantic mode.

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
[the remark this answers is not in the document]
milseman: Should we also start tracking the extended syntax support and options?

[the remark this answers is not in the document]
milseman: Hah, I think they have an extra dot in their docs:

>   (?#....)        comment (not nestable)

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
milseman: CC @natecook1000 . This could be similar to combining regexes into grammars of syntax highlighters.

Sources/_MatchingEngine/Utility/MissingUnicode.swift
milseman: Do we have any rationale why that particular syntax would deviate from _both_ the normal mechanisms for reasoning about semantic levels _and_ Unicode's explicit recommendations in [UTS-18 Annex C](https://www.unicode.org/reports/tr18/#Compatibility_Properties)?

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Utility/MissingUnicode.swift
kylemacomber: IMO the shortest spelling feels the most underspecified. I think we can defend it more at length in the proposal. FWIW It seems like this decision is precedented in Perl: https://perldoc.perl.org/perlre#/a-(and-/aa)

> **/a (and /aa)**
> 
> This modifier stands for ASCII-restrict (or ASCII-safe). This modifier may be doubled-up to increase its effect.
> 
> When it appears singly, it causes the sequences \d, \s, \w, and the Posix character classes to match only in the ASCII range.

milseman: By default, perl will follow the model we're outlining here. If you specify ASCII mode in that way (or posix mode in some engines), then the POSIX-syntactic classes will be restricted to ASCII. The remaining question is whether `\p` should as well, which I don't think is a question we'll be answering unilaterally, and there are contextual options controlling this:

https://github.com/kkos/oniguruma/blob/master/doc/RE#L270

```
  (?imxWDSPy-imxWDSP:subexp)  option on/off for subexp

                           i: ignore case
                           m: multi-line (dot (.) also matches newline)
                           x: extended form
                           W: ASCII only word (\w, \p{Word}, [[:word:]])
                              ASCII only word bound (\b)
                           D: ASCII only digit (\d, \p{Digit}, [[:digit:]])
                           S: ASCII only space (\s, \p{Space}, [[:space:]])
                           P: ASCII only POSIX properties (includes W,D,S)
                              (alnum, alpha, blank, cntrl, digit, graph,
                               lower, print, punct, space, upper, xdigit, word)

                           y{?}: Text Segment mode
                              This option changes the meaning of \X, \y, \Y.
                              Currently, this option is supported in Unicode only.

                              y{g}: Extended Grapheme Cluster mode (default)
                              y{w}: Word mode
                              See [Unicode Standard Annex #29]

```

Either way, this affects the AST representation to the extent we record these options and present API for reasoning. I don't think we have a good enough feel at this moment to decide whether it's a different syntactic node or not. Most of the AST API relevant to users is along a more semantic (and perhaps parameterized) level instead of derived via a parse tree.

milseman:
@hamishknight is this PR about ready to be merged?

hamishknight:
@swift-ci please test Linux

PR #65, opened 2021-12-10
Initial Unicode module

milseman:
This includes a port of `UnicodeHelpers.swift` from the standard library, and stubs for other stuff.


atrick:
I don't get why a public `decode` function would take anything other than a raw pointer. The type to decode is explicit in the signature. A typed pointer is never the right way to view a block of memory that's about to be decoded.

If this is about wanting to internally work with a Collection of CodeUnit, so subscripting does the scaling-by-stride for you, then I'd use a trivial UnicideCodeUnit wrapper over the raw pointer.

karwa:
> We need Unicode functionality without allocating a String as an intermediary
...
> We can't be bothered to consistently contribute to the stdlib because it's so painful

Preach it! 🙌

milseman:
@atrick good point, we should change that (or have both overloads). As you know, there is no contiguousRawStorageIfAvailable equivalent, so types may live on for a while. 

We’d want to migrate to a move only safe view as soon as that’s available. If anything, this module could be a prime example of why that’s so vital.

milseman:

milseman:
Merging to unblock some lexer work.

PR #67, opened 2021-12-12
Refactor/rename source locations, tracking, etc

milseman:
Track source locations for Atoms, quantifier sub-values (like ranges and modifiers), and other refactoring and changes.

Naming convention:

- `Source.Position`: A dimensionless point in the input
    - Corresponds to an index in the input
- `Source.Location`: The region of the input from which a parsed entity came
    - Corresponds to an index-range in the input, but separate type to hold API
    - `SourceLocation` typealias
- `AST.Loc` / `Source.Loc`: a located value, that is a value with a location.
    - Name is short because it's ubiquitous

AST builders and faking:

- Builder global functions internalized and sunk into the `_StringProcessing` module
- `_fake` globals removed in favor of static members or inits on `SourceLocation`
- `atom_m` makes the custom char set member variant instead of the AST variant

Changes to AST or parser functionality:

- Trivia now records the contents as a String
- Source is no longer a collection, and dropped the Peekable protocol.
- Atoms tracked
- Quantifier values (like the ranges or kinds) tracked

After this PR, there's only a handful of minor things to track:

- The `|` in alternations
- Potentially, even finer-grained pieces:
    - The `-` in ranges inside custom character classes
    - Such as the closing `)` tracked separately, though that's derivable from the group's range
    - Quote's `\Q` and `\E` distinct from the contents of the quote
    - The location of the `^` for inverting Unicode properties inside a `\p{^...}`
    - The portion covering the name in named back references (groups do have it though)
    - The location of the comma in a quantification `a{3,7}`
    - ...


hamishknight (approved):

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+  }
+
+  /// Located value: a value wrapped with a source range
+  ///
+  /// Note: source location is part of value identity, so that the same
+  /// e.g. `Character` appearing twice can be stored in a data structure
+  /// distinctly. To ignore source locations, use `.value` directly.
+  public struct Loc<T: Hashable>: Hashable {
```
rxwei, line 56: The existence of both `Source.Location` and `Source.Loc` makes it less readable as the latter could be misinterpreted as the former (esp. for people coming from the compiler repo, i.e. `SourceLoc`).  How about using the full name `Located<T>`

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
```diff
@@ -0,0 +1,135 @@
...
+}
+public typealias SourceLocation = Source.Location
+
+extension Source {
+  var currentPosition: Position { bounds.lowerBound }
+}
+
+extension Source {
+  /// An error with source location info
+  public struct Err<E: Error>: Error {
```
rxwei, line 36: Could we use a more descriptive name instead of an abbreviation that could be easily misinterpreted? Suggest `LocatedError<E>`.

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
```diff
@@ -0,0 +1,135 @@
...
+extension AST {
+  public typealias Loc = Source.Loc
+}
+
+extension Source {
+  // MARK: - recordLoc
+
+  /// Record source loc before processing and return
+  /// or throw the value/error with source locations.
+  mutating func recordLoc<T>(
```
rxwei, line 85:
```suggestion
  mutating func recordLocation<T>(
```

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
```diff
@@ -0,0 +1,135 @@
...
+    public init(faking v: T) {
+      self.init(v, .fake)
+    }
+
+    var start: Source.Position { location.start }
+    var endLoc: Source.Position { location.end }
+  }
+}
+extension AST {
+  public typealias Loc = Source.Loc
```
rxwei, line 77: For consistency, suggest `Located` instead of abbreviation.

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
```diff
@@ -0,0 +1,135 @@
...
+    public init(_ v: T, _ r: Range<Source.Position>) {
+      self.value = v
+      self.location = Location(r)
+    }
+    public init(faking v: T) {
+      self.init(v, .fake)
+    }
+
+    var start: Source.Position { location.start }
+    var endLoc: Source.Position { location.end }
```
rxwei, line 73: Nit: Suggest the consistent use of suffix "location" (either both or neither).  Here the "start location" variable name doesn't have "location" in it but "end location" does. Since the base name is `Located`, I think that "loc" is redundant, hence the following suggestion:
```suggestion
    var start: Source.Position { location.start }
    var end: Source.Position { location.end }
```

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
milseman: I agree for public API (even public for the parser).

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
milseman: Sounds good, though I think we're likely going to make this internal and formally use a ParseError struct with the location. These wrappers are trending towards being useful for lexical analysis and holding sourced/located values, as opposed to AST nodes themselves. Diagnostics will likely follow suit.

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
milseman: Actually, this shouldn't even be internal. I'll make it `fileprivate` to lexical analysis. Within fileprivate usage, I think the abbreviation is appropriate.

Sources/_MatchingEngine/Regex/Parse/SourceLocation.swift
milseman: Hah, I was meaning to remove those or renamed them. It slipped my other cleanups

milseman:

PR #68, opened 2021-12-13
'Regex<Captures>' -> 'Regex<Match>'

rxwei:
Changes `Regex` and result builder prototype to use `Match` as the generic parameter to make it consistent with the [Strongly Typed Regex Captures](https://forums.swift.org/t/pitch-strongly-typed-regex-captures/53391) pitch.

Introduces `Tuple<n>` structs in order to be able to express constraints on capture types (i.e. `Match` dropped first) while being able to filter out empty captures in concatenation. `Tuple<n>` is also needed to implement a prototype of the [proposed matching semantics](#64).

```swift
@frozen @dynamicMemberLookup
public struct Tuple{n}<_0, _1, ...> {
  public typealias Tuple = (_0, _1, ...)
  public var tuple: Tuple
  public init(_ tuple: Tuple)
  public init(_ _0: _0, _ _1: _1, ...)
  public subscript<T>(dynamicMember keyPath: WritableKeyPath<Tuple, T>) -> T { get _modify }
}
```

As coercion into `Tuple<n>` can no longer use runtime magic like native tuples do, we incorporate child capture type information into RECode's `captureNil` and `captureArray` instructions so that we will always get a concrete type when forming a nil or an empty array capture. The resulting existential tuple capture can then be opened and bitcast to a `Tuple<n>`.

### DIFF
--- a/Sources/_StringProcessing/RegexDSL/Core.swift
+++ b/Sources/_StringProcessing/RegexDSL/Core.swift
@@ -1,18 +1,23 @@
 import _MatchingEngine
 
-public struct RegexMatch<CapturedValue> {
+@dynamicMemberLookup
+public struct RegexMatch<Match> {
   public let range: Range<String.Index>
-  public let captures: CapturedValue
+  public let match: Match
+
+  public subscript<T>(dynamicMember keyPath: KeyPath<Match, T>) -> T {
+    match[keyPath: keyPath]
+  }
 }
 
 /// A type that represents a regular expression.
 public protocol RegexProtocol {
-  associatedtype Capture
-  var regex: Regex<Capture> { get }
+  associatedtype Match: MatchProtocol
+  var regex: Regex<Match> { get }
 }
 
 /// A regular expression.
-public struct Regex<Capture>: RegexProtocol {
+public struct Regex<Match: MatchProtocol>: RegexProtocol {
   /// A program representation that caches any lowered representation for
   /// execution.
   internal class Program {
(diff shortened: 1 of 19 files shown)


rxwei:
@kylemacomber This paves the way for `Submatch` (or `Unbound`, whatever it's called) :)

rxwei:
@swift-ci please test linux

rxwei:
@swift-ci please test linux

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -58,86 +63,109 @@ public struct Regex<Capture>: RegexProtocol {
...
+public struct EmptyCapture: EmptyCaptureProtocol {}
+extension Array: EmptyCaptureProtocol where Element: EmptyCaptureProtocol {}
+extension Optional: EmptyCaptureProtocol where Wrapped: EmptyCaptureProtocol {}
 
+public protocol MatchProtocol {
+  associatedtype Capture
+}
+extension Substring: MatchProtocol {
+  public typealias Capture = EmptyCapture
+}
```
milseman, line 171: I don't recall whether modern computing tolerates no newline at the end of a file.

milseman (approved):
Overall LGTM.

In general, switching to a nominal type makes sense to me. Our of curiosity, what necessitated this change? Also, do we need those generic parameters?

rxwei:
`buildBlock`s need to determine whether a component has no captures. A component has no captures when every capture conforms to `EmptyCaptureProtocol`.  For example, a `OneOrMore` quantifier's `Match` is a tuple of its whole match and an array of its capture. `[Component.Match.Capture]` conforms to `EmptyCaptureProtocol` when `Component.Match.Capture` (a `Tuple{n}`) conforms.

```swift
public struct OneOrMore<Component: RegexProtocol>: RegexProtocol {
  public typealias Match = Tuple2<Substring, [Component.Match.Capture]>
```

```swift
extension Tuple5: EmptyCaptureProtocol where _1: EmptyCaptureProtocol, _2: EmptyCaptureProtocol, _3: EmptyCaptureProtocol, _4: EmptyCaptureProtocol {}
```

Sources/_StringProcessing/RegexDSL/Core.swift
rxwei: Fixed. It's partly because I've been trying to use Windows to develop these days, trying Swift for Windows myself.

rxwei:
@swift-ci please test Linux

rxwei:
This apparently affects regex literal type inference in the compiler, so I'll put up a PR on the compiler side to test against this.

rxwei:
@swift-ci please test linux

rxwei:
@swift-ci please test Linux

PR #69, opened 2021-12-13
Add 'init(_regexString:)' for compiler integration.

rxwei:
Add 'init(_regexString:)' for compiler integration.

Also make `Regex` frozen to avoid it being an address-only type. I think it's not likely going to need resilience, but we can also change later if needed.

### DIFF
--- a/Sources/_StringProcessing/RegexDSL/Core.swift
+++ b/Sources/_StringProcessing/RegexDSL/Core.swift
@@ -12,9 +12,11 @@ public protocol RegexProtocol {
 }
 
 /// A regular expression.
+@frozen
 public struct Regex<Capture>: RegexProtocol {
   /// A program representation that caches any lowered representation for
   /// execution.
+  @usableFromInline
   internal class Program {
     /// The underlying AST.
     let ast: AST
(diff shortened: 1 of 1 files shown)


rxwei:
@swift-ci please test Linux

rxwei:
@swift-ci please test Linux

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -12,9 +12,11 @@ public protocol RegexProtocol {
 }
 
 /// A regular expression.
+@frozen
```
milseman, line 15: What's the rationale for this? I was expecting `Regex` to be resilient, and certainly not expecting to put the program class in our ABI

rxwei: When do you expect it to be resilient?  `Program` acts like the guts that caches the compiled program, and I think that expansion to `Regex` could be added to that type.  As mentioned in the PR description, I'd like to avoid it address-only type, but I can change if we need to.

milseman: I might be prematurely optimizing this, but I was thinking that if we froze the-type-that-can-be-executed, we'd boil it down to a `_NativeObject`-like reference to supports immortal references for literals. 

(Also, I don't know if we'll use the same type for a linked/executable program as the type used for program composition)

What are you worried about specifically if this is address only? Is there something you want or need to inline elsewhere to get around the generic parameters?

rxwei: > What are you worried about specifically if this is address only? Is there something you want or need to inline elsewhere to get around the generic parameters?

Just slightly more straightforward to emit the code in SILGen as well as file-checking it. But sure, let me change it back to non-frozen.

PR #70, opened 2021-12-13
Fix an issue with 'DynamicCaptures' where there is no captures.

rxwei:


rxwei:
@swift-ci please test Linux

rxwei:
Merging to unblock apple/swift#40531.

PR #71, opened 2021-12-13
Make 'Regex' resilient.

rxwei:
Drop `@frozen` as requested.


rxwei:
@swift-ci please test Linux

PR #72, opened 2021-12-13
Quick compiler interface mockup

milseman:
Quick mockup of full lexing in the package, where the Swift compiler hands off a pointer to us and we proceed from there, instead of the compiler parsing the delimiter for us.


milseman:
@hamishknight this should get you off and running. I suspect you'll want to also return the end pointer after the close (should be `current.advanced(by: 2)` or something like that, or else you can just consume those bytes inside the loop).

milseman:

Issue #73, opened 2021-12-14
Failing ParseableInterface test


rxwei:
_MatchingEngine is failing the ParseableInterface test in the Swift repo.

```console
/Volumes/Media/Development/Swift/swift-source/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/macosx/_MatchingEngine.swiftmodule/x86_64-apple-macos.swiftinterface:1460:1: error: type 'TypedIndex<C, 👻>' does not conform to protocol 'RangeReplaceableCollection'
extension _MatchingEngine.TypedIndex : Swift.RangeReplaceableCollection where C : Swift.RangeReplaceableCollection {
^
/Volumes/Media/Development/Swift/swift-source/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/macosx/_MatchingEngine.swiftmodule/x86_64-apple-macos.swiftinterface:1460:1: error: unavailable instance method 'replaceSubrange(_:with:)' was used to satisfy a requirement of protocol 'RangeReplaceableCollection'
extension _MatchingEngine.TypedIndex : Swift.RangeReplaceableCollection where C : Swift.RangeReplaceableCollection {
^
Swift.RangeReplaceableCollection:4:26: note: 'replaceSubrange(_:with:)' declared here
    public mutating func replaceSubrange<C>(_ subrange: Range<Self.Index>, with newElements: C) where C : Collection, Self.Element == C.Element
                         ^
Swift.RangeReplaceableCollection:4:19: note: requirement 'replaceSubrange(_:with:)' declared here
    mutating func replaceSubrange<C>(_ subrange: Range<Self.Index>, with newElements: __owned C) where C : Collection, Self.Element == C.Element
                  ^
/Volumes/Media/Development/Swift/swift-source/build/Ninja-ReleaseAssert/swift-macosx-x86_64/lib/swift/macosx/_MatchingEngine.swiftmodule/x86_64-apple-macos.swiftinterface:1464:14: error: no exact matches in call to instance method 'replaceSubrange'
    rawValue.replaceSubrange(rawRange, with: newElements)
             ^
Swift.RangeReplaceableCollection:4:19: note: candidate requires that the types 'C.Element' and 'C.Element' be equivalent (requirement specified as 'Self.Element' == 'C.Element')
    mutating func replaceSubrange<C>(_ subrange: Range<Self.Index>, with newElements: __owned C) where C : Collection, Self.Element == C.Element
[...]

PR #74, opened 2021-12-14
ParseableInterface test failure workaround (#73)

rxwei:
Disables `RangeReplaceableCollection` conformance for now until we find a way to declare a conformance without triggering #73.

However if there's a viable fix for #73, we should go with that.


rxwei:
@swift-ci please test linux

milseman (approved):
LGTM

PR #75, opened 2021-12-14
Match tests

kylemacomber:
Adding match tests that correspond to parse tests on top of #62

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,7 +46,7 @@ let package = Package(
             dependencies: []),
         .testTarget(
             name: "RegexTests",
-            dependencies: ["_StringProcessing"]),
+            dependencies: ["_StringProcessing", "Algorithms"]),
         .testTarget(
             name: "RegexDSLTests",
             dependencies: ["_StringProcessing"]),
--- a/Tests/RegexTests/MatchTests.swift
+++ b/Tests/RegexTests/MatchTests.swift
@@ -11,9 +11,9 @@ func matchTest(
   xfail: Bool = false
 ) {
   guard !xfail else {
-    XCTExpectFailure {
-      XCTFail()
-    }
+//    XCTExpectFailure {
+//      XCTFail()
+//    }
     return
   }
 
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -46,7 +46,7 @@ let package = Package(
             dependencies: []),
         .testTarget(
             name: "RegexTests",
-            dependencies: ["_StringProcessing"]),
+            dependencies: ["_StringProcessing", "Algorithms"]),
```
milseman, line 49: Related: @timvermeulen any progress on vetting these API for the `_StringProcessing` model?

Tests/RegexTests/MatchTests.swift
```diff
@@ -0,0 +1,403 @@
...
+      "a|b?c", input: "123bcxyz", match: "bc")
+    matchTest(
+      "(a|b)c", input: "123abcxyz", match: "bc")
+    matchTest(
+      "(.)*(.*)", input: "123abcxyz", match: "123abcxy", xfail: true)
+    matchTest(
+      #"abc\d"#, input: "xyzabc123", match: "abc1")
+
+    // MARK: Unicode scalars
+
```
milseman, line 70: I don't think it makes sense to duplicate parse tests below, but it would be nice to convert the below into end-to-end tests.

kylemacomber:
@swift-ci please test

kylemacomber:
@swift-ci please test Linux

Tests/RegexTests/MatchTests.swift
```diff
@@ -0,0 +1,467 @@
...
+      #"a(?#comment)b"#, input: "123abcxyz", match: "ab")
+    matchTest(
+      #"a(?#. comment)b"#, input: "123abcxyz", match: "ab")
+
+    // MARK: Quantification
+
+    matchTest(
+      #"a{1,2}"#, input: "123aaaxyz", match: "aa", xfail: true)
+    matchTest(
+      #"a{,2}"#, input: "123aaaxyz", match: "", xfail: true)
```
milseman, line 230: Why is the match empty? This would be eager by default, so I would expect the first "aa" to match.

milseman: (note that PCRE doesn't parse the `{,2}` as a range, but Python, Oniguruma, and others do)

milseman: Ugh, never mind, it successfully matches the empty range at the start.

PR #76, opened 2021-12-15
Delimiter lexing support for `'/` and `'|`

hamishknight:
Complete the delimiter lexing support such that the C++ compiler's lexer will be able to call into the implementation and allow us to lex the delimiters we want to support, and hand back the compiler a pointer from which to resume lexing.

In addition, expose a parsing entrypoint that expects a regex literal with delimiters, from which we can infer the syntax options. This will be used by both the C++ parser, and by a new `Regex(_regexString:version:)` initializer that the compiler will be switched over to calling. Adding a version parameter will allow us to more easily make versioned changes to the format of regex string we pass to the runtime.

Once the C++ logic is switched over to these new entrypoints, the regex delimiters will no longer be `'...'`, instead they will be `'/.../'` for traditional interior syntax, or `'|...|'` for modern syntax.

### DIFF
--- a/Sources/_MatchingEngine/Regex/Parse/Mocking.swift
+++ b/Sources/_MatchingEngine/Regex/Parse/Mocking.swift
@@ -37,71 +37,211 @@ struct SwiftLexer {
 
 // TODO: mock up multi-line soon
 
+enum Delimiter: Hashable, CaseIterable {
+  case regular
+  case modern
 
-enum LexError: Error {
-  case endOfString
-  case invalidUTF8 // TODO: better range reporting
-  case notRegexLiteralStart
+  var openingAndClosing: (opening: String, closing: String) {
+    switch self {
+    case .regular: return ("'/", "/'")
+    case .modern: return ("'|", "|'")
+    }
+  }
+  var opening: String { openingAndClosing.opening }
+  var closing: String { openingAndClosing.closing }
+
+  /// The default set of syntax options that the delimiter indicates.
+  var defaultSyntaxOptions: SyntaxOptions {
+    switch self {
+    case .regular: return .traditional
+    case .modern: return .modern
+    }
+  }
 }
 
-public func _lexRegex(
-  _ start: UnsafeRawPointer
-) throws -> (String, SyntaxOptions)? {
+struct LexError: Error, CustomStringConvertible {
+  enum Kind: Hashable {
+    case endOfString
+    case invalidUTF8 // TODO: better range reporting
+    case unknownDelimiter
+  }
+
+  var kind: Kind
+
+  /// The pointer at which to resume lexing.
+  var resumePtr: UnsafeRawPointer
+
+  init(_ kind: Kind, resumeAt resumePtr: UnsafeRawPointer) {
+    self.kind = kind
+    self.resumePtr = resumePtr
+  }
+
+  var description: String {
+    switch kind {
+    case .endOfString: return "unterminated regex literal"
+    case .invalidUTF8: return "invalid UTF-8 found in source file"
+    case .unknownDelimiter: return "unknown regex literal delimiter"
+    }
+  }
+}
+
+/// Drop a set of regex delimiters from the input string, returning the contents
+/// and the delimiter used. The input string must have valid delimiters.
+func droppingRegexDelimiters(_ str: String) -> (String, Delimiter) {
+  let utf8 = str.utf8
+  func stripDelimiter(_ delim: Delimiter) -> String? {
+    let prefix = delim.opening.utf8
+    let suffix = delim.closing.utf8
+    guard utf8.prefix(prefix.count).elementsEqual(prefix),
+          utf8.suffix(suffix.count).elementsEqual(suffix) else { return nil }
+
+    return String(utf8.dropFirst(prefix.count).dropLast(suffix.count))
+  }
+  for d in Delimiter.allCases {
+    if let contents = stripDelimiter(d) {
+      return (contents, d)
+    }
+  }
+  fatalError("No valid delimiters")
+}
+
+/// Attempt to lex a regex literal between `start` and `end`, returning either
+/// the contents and pointer from which to resume lexing, or an error.
+func lexRegex(
+  start: UnsafeRawPointer, end: UnsafeRawPointer
+) throws -> (contents: String, Delimiter, end: UnsafeRawPointer) {
+  precondition(start <= end)
   var current = start
 
   func ascii(_ s: Unicode.Scalar) -> UInt8 {
     assert(s.value <= 0x7F)
     return UInt8(truncatingIfNeeded: s.value)
   }
-  func load(offset: Int) -> UInt8 {
-    current.load(fromByteOffset: offset, as: UInt8.self)
+  func load(offset: Int) -> UInt8? {
+    guard current + offset < end else { return nil }
+    return current.load(fromByteOffset: offset, as: UInt8.self)
   }
-  func load() -> UInt8 { load(offset: 0) }
-  func advance() { current = current.successor() }
-  func eat() -> UInt8 {
-    defer { advance() }
-    return load()
-  }
-
-  guard eat() == ascii("'") else {
-    throw LexError.notRegexLiteralStart
+  func load() -> UInt8? { load(offset: 0) }
+  func advance(_ n: Int = 1) {
+    precondition(current + n <= end, "Cannot advance past end")
+    current = current.advanced(by: n)
   }
 
-  let syntax: SyntaxOptions
-  let delimiter = eat()
-  switch delimiter {
-  case ascii("|"): syntax = .modern
-  case ascii("/"): syntax = .traditional
-  default: return nil
+  func tryEat(_ utf8: String.UTF8View) -> Bool {
+    for (i, idx) in utf8.indices.enumerated() {
+      guard load(offset: i) == utf8[idx] else { return false }
+    }
+    advance(utf8.count)
+    return true
+  }
+
+  // Try to lex the opening delimiter.
+  guard let delimiter = Delimiter.allCases.first(
+    where: { tryEat($0.opening.utf8) }
+  ) else {
+    throw LexError(.unknownDelimiter, resumeAt: current.successor())
   }
 
   let contentsStart = current
-  let count: Int = try {
-    while true {
-      switch load() {
-      case 0: throw LexError.endOfString
-      case ascii("\\"):
-        // Skip next byte
-        advance()
-        advance()
-      case delimiter:
-        if load(offset: 1) == ascii("'") {
-          return current - contentsStart
-        }
-        advance()
+  while true {
+    switch load() {
+    case nil, ascii("\n"), ascii("\r"):
+      throw LexError(.endOfString, resumeAt: current)
 
-      default: advance()
+    case ascii("\\"):
+      // Skip next byte.
+      advance(2)
+
+    default:
+      // Try to lex the closing delimiter.
+      let contentsEnd = current
+      guard tryEat(delimiter.closing.utf8) else {
+        advance()
+        continue
       }
+
+      // Form a string from the contents and make sure it's valid UTF-8.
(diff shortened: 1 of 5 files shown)


hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
-      return lexSingleQuotedLiteral(
-        src[src.index(after: start)..<src.lastIndex(of: "'")!])
+public enum Delimiter: Hashable, CaseIterable {
+  case regular
+  case modern
+
+  var openingAndClosing: (opening: String, closing: String) {
+    switch self {
+    case .regular: return ("'/", "/'")
+    case .modern: return ("'|", "|'")
```
milseman, line 11: Out of curiosity, why?

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
+    case invalidUTF8 // TODO: better range reporting
+    case unknownDelimiter
   }
 
-}
+  var kind: Kind
 
-// TODO: mock up multi-line soon
+  /// The pointer at which to resume lexing.
+  public var resumePtr: UnsafeRawPointer
```
milseman, line 36: So we have lexing errors and parsing errors. I thought a lexing error would just send the Swift compiler down the diagnosing single quotes error path, or even better, a path that checks the experimental flag and tells you what delimiters to use.

What is the resumption design?

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
+    case .endOfString: return "unterminated regex literal"
+    case .invalidUTF8: return "invalid UTF-8 found in source file"
+    case .unknownDelimiter: return "unknown regex literal delimiter"
+    }
+  }
 }
 
+/// Drop a set of regex delimiters from the input string, returning the contents
+/// and the delimiter used. The input string must have valid delimiters.
+func droppingRegexDelimiters(_ str: String) -> (String, Delimiter) {
```
milseman, line 58: I don't understand the point of this function, why would we have a parallel implementation over String?

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
 
-  let syntax: SyntaxOptions
-  let delimiter = eat()
-  switch delimiter {
-  case ascii("|"): syntax = .modern
-  case ascii("/"): syntax = .traditional
-  default: return nil
+  // Try to lex the opening delimiter.
+  guard let delimiter = Delimiter.allCases.first(
+    where: { tryEat($0.opening.utf8) }) else {
```
milseman, line 108:
```suggestion
    where: { tryEat($0.opening.utf8) }
  ) else {
```

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
-    while true {
-      switch load() {
-      case 0: throw LexError.endOfString
-      case ascii("\\"):
-        // Skip next byte
-        advance()
+  while true {
+    switch load() {
+    case nil, ascii("\n"), ascii("\r"):
+      return .failure(LexError(.endOfString, resumeAt: current))
```
milseman, line 116: Not for this PR, but how do you think we should do multi-line support? I guess with the version argument we can add that at any time

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
+
+    case ascii("\\"):
+      // Skip next byte.
+      advance(2)
+
+    default:
+      // Try to lex the closing delimiter.
+      let preClosingDelimiter = current
```
milseman, line 124: `contentsEnd`?

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,142 @@
...
+  fatalError("No valid delimiters")
+}
+
+/// Attempt to lex a regex literal between `start` and `end`, returning either
+/// the contents and pointer from which to resume lexing, or an error.
 public func _lexRegex(
-  _ start: UnsafeRawPointer
-) throws -> (String, SyntaxOptions)? {
+  start: UnsafeRawPointer, end: UnsafeRawPointer
+) -> Result<(contents: String, Delimiter, end: UnsafeRawPointer), LexError> {
```
milseman, line 80: Why a `Result`? I think we can be throwing in everything except the C wrapper shim which can catch the result and pick a common handling strategy.

hamishknight: I really only added it so the error could be strongly typed as a `LexError` to make sure that libswift can pick up the resumption point in the case of an error. I'm happy to make it throwing and have libswift catch the LexError specifically though.

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
hamishknight: Yeah, as long as we can pick it up by e.g a different delimiter, we could change the logic to support multiple lines

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
hamishknight: It's used by both:

- The runtime to drop the delimiters off and get the syntax options to use
- The C++ parser as we don't currently track the contents range on the C++ `Token` so we need to drop off the delimiters again when we come to parse

We could have both of those go back through the lexing function, I don't feel too strongly about it. I agree it's not ideal having two implementations of effectively the same operation

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
hamishknight: > I thought a lexing error would just send the Swift compiler down the diagnosing single quotes error path

For errors like invalid UTF-8 and end-of-string, wouldn't we want to diagnose that instead?

> What is the resumption design?

After emitting the error, the C++ lexer resumes lexing at the given pointer, which lets us have better error handling for things like an unterminated literal, as we'd resume lexing at the newline.

milseman: Ah makes sense.

> For errors like invalid UTF-8 and end-of-string, wouldn't we want to diagnose that instead?

We could emit it, or we could fall back to the lexer to emit it. I thought the latter might be a simpler integration story for now, but either way.

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
milseman: The delimiter ideally would be something like 

```
let r = '''/
  regex
  '''/
```

or something.

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
milseman: The mapping logic should remain in the package. This isn't the function directly called by compiler, so it can throw. We can later look at refactoring or sharing logic with later down the pipeline.

hamishknight:
Moved the libswift entrypoints into the repo, how does this look @milseman?

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,199 @@
...
 
-  guard s.utf8.elementsEqual(contents) else {
-    throw LexError.invalidUTF8
+/// Public interface for libswift.
+///
+/// Lex a regular expression literal starting at `inputPtr`, making sure not to
+/// lex past `bufferEndPtr`. The pointer at which to resume lexing is returned,
+/// or nil if this is not a regex literal. The `errOut` parameter will be set
+/// if an error is encountered.
+public func libswiftLexRegexLiteral(
```
rxwei, line 157: Can this be `@usableFromInline internal` instead?

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,199 @@
...
+    fatalError("Should be a LexError")
   }
+}
 
-  return (s, syntax)
+/// Public interface for libswift.
+///
+/// Parse a regex string from `inputPtr`, which should be null-terminated
+/// C-string. `errOut` will be set if an error was encountered.
+public func libswiftParseRegexLiteral(
```
rxwei, line 185: Same question.

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
hamishknight: Ah good point! Updated

hamishknight:
@swift-ci please test Linux

rxwei (approved):

milseman (approved):
Overall LGTM and follow up can be in a subsequent PR.

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
```diff
@@ -1,107 +1,201 @@
...
+/// Lex a regular expression literal starting at `inputPtr`, making sure not to
+/// lex past `bufferEndPtr`. The pointer at which to resume lexing is returned,
+/// or nil if this is not a regex literal. The `errOut` parameter will be set
+/// if an error is encountered.
+@usableFromInline
+func libswiftLexRegexLiteral(
+  _ inputPtr: UnsafePointer<CChar>?,
+  _ bufferEndPtr: UnsafePointer<CChar>?,
+  _ errOut: UnsafeMutablePointer<UnsafePointer<CChar>?>?
+) -> UnsafePointer<CChar>? {
```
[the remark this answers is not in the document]
rxwei: Yeah, we need something like `@_cdecl("_libswift_lexRegexLiteral")`.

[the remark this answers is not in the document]
milseman: Since the initialize function in libSwift makes a closure to call this function, I believe it will get an automagic `@convention(c)` on it.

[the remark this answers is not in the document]
rxwei: If it's dynamically registering a closure, we may not need `@usableFromInline`.

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/Mocking.swift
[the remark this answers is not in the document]
hamishknight: Updated to drop `@usableFromInline`

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -47,6 +47,14 @@ public struct Regex<Capture>: RegexProtocol {
     self.init(ast: try! parse(pattern, .traditional))
   }
 
+  // Compiler interface. Do not change independently.
+  @usableFromInline
+  init(_regexString pattern: String, version: Int) {
+    // The version argument is passed by the compiler using the value defined
+    // in libswiftParseRegexLiteral.
```
milseman, line 54: Can we assert its equal to 1, at least for now? And/or make an internal global for current version number, and we'll expand on that if/when we need a new one

milseman (approved):
LGTM. Comments can be done after merge if you want.

milseman:

Sources/_StringProcessing/RegexDSL/Core.swift
hamishknight: How's this?

hamishknight:
@swift-ci please test Linux

PR #79, opened 2021-12-16
[gardening] Some test reorganization

milseman:
XCTest seems to spawn a process and wait per target, so we can consolidate our test targets like we did our code targets.


milseman:

milseman:
Merging to unblock

PR #80, opened 2021-12-16
AST Atom and custom char class consumers

milseman:
Generate consumers for AST nodes that are atoms or custom character classes.

Most of the char class and props tests are now passing, notably except for scripts (which the stdlib doesn't surface yet, cc @Azoy).

### DIFF
--- a/Sources/_StringProcessing/ConsumerInterface.swift
+++ b/Sources/_StringProcessing/ConsumerInterface.swift
@@ -354,25 +354,41 @@ extension Unicode.BinaryProperty {
     case .diacratic: // spelling?
       return consumeScalarProp(\.isDiacritic)
     case .emojiModifierBase:
-      break // availability
+      if #available(macOS 10.12.2, *) {
+        return consumeScalarProp(\.isEmojiModifierBase)
+      } else {
+        throw unsupported("isEmojiModifierBase on old OSes")
+      }
     case .emojiComponent:
       break
     case .emojiModifier:
-      break // availability
+      if #available(macOS 10.12.2, *) {
+        return consumeScalarProp(\.isEmojiModifier)
+      } else {
+        throw unsupported("isEmojiModifier on old OSes")
+      }
     case .emoji:
-      break // availability
+      if #available(macOS 10.12.2, *) {
+        return consumeScalarProp(\.isEmoji)
+      } else {
+        throw unsupported("isEmoji on old OSes")
+      }
     case .emojiPresentation:
-      break // availability
+      if #available(macOS 10.12.2, *) {
+        return consumeScalarProp(\.isEmojiPresentation)
+      } else {
+        throw unsupported("isEmojiPresentation on old OSes")
+      }
     case .extender:
       return consumeScalarProp(\.isExtender)
     case .extendedPictographic:
-      break
+      break // NOTE: Stdlib has this data internally
     case .fullCompositionExclusion:
       return consumeScalarProp(\.isFullCompositionExclusion)
     case .graphemeBase:
       return consumeScalarProp(\.isGraphemeBase)
     case .graphemeExtended:
-      break
+      return consumeScalarProp(\.isGraphemeExtend)
     case .graphemeLink:
       break
     case .hexDigit:
@@ -426,7 +442,9 @@ extension Unicode.BinaryProperty {
     case .radical:
       return consumeScalarProp(\.isRadical)
     case .regionalIndicator:
-      break
+      return consumeScalar { s in
+        (0x1F1E6...0x1F1FF).contains(s.value)
+      }
     case .softDotted:
       return consumeScalarProp(\.isSoftDotted)
     case .sentenceTerminal:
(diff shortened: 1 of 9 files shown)


Tests/RegexTests/MatchTests.swift
```diff
@@ -246,110 +256,155 @@ extension RegexTests {
...
+    matchTest(#"\p{gc=L}"#, input: "123abcXYZ", match: "a")
+    matchTest(#"\p{Lu}"#, input: "123abcXYZ", match: "X")
+
+    // TODO: diagnose
+    matchTest(
+      #"\P{Cc}"#, input: "123\n\n\nXYZ", match: "\n",
+      xfail: true)
+    matchTest(
+      #"\P{Z}"#, input: "123 XYZ", match: " ",
+      xfail: true)
```
kylemacomber, line 282: When I was writing these I think I missed that capital P was a negation. The following should work:

```suggestion
    // TODO: diagnose
    matchTest(
      #"\P{Cc}"#, input: "\n\n\nXYZ", match: "X")
    matchTest(
      #"\P{Z}"#, input: "   XYZ", match: "X")
```

kylemacomber (approved):
I only looked at the changes to the tests. This looks like great progress!

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -0,0 +1,602 @@
...
+    case .otherIDContinue:
+      break
+    case .otherIDStart:
+      break
+    case .otherLowercase:
+      break
+    case .otherMath:
+      break
+    case .otherUppercase:
+      break
```
Azoy, line 396: These properties are unfortunate because they get scooped up with their normal counterpart (`Alphabetic`) to match ICU's behavior when asking if a scalar `isAlphabetic`. If we need to make the distinction here, then that'll probably need to be a separate table.

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -0,0 +1,602 @@
...
+    case .patternWhitespace:
+      return consumeScalarProp(\.isPatternWhitespace)
+    case .prependedConcatenationMark:
+      break
+    case .quotationMark:
+      return consumeScalarProp(\.isQuotationMark)
+    case .radical:
+      return consumeScalarProp(\.isRadical)
+    case .regionalIndicator:
+      break
```
Azoy, line 408: This property is a simple range `0x1F1E6...0x1F1FF` if we wanted to expose this.

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -0,0 +1,602 @@
...
+    case .xidStart:
+      return consumeScalarProp(\.isXIDStart)
+    case .expandsOnNFC:
+      break
+    case .expandsOnNFD:
+      break
+    case .expandsOnNFKC:
+      break
+    case .expandsOnNFKD:
+      break
```
Azoy, line 434: These properties are all deprecated as of Unicode 6 (we're on 14 now). Do we need to support these?

Azoy (commented):
There's a lot of properties that we don't have yet that probably make sense to expose at some point. Regarding scripts, I'm going to need to store some information on scripts regardless because I need to implement the CLDR grapheme breaking rule to be consistent with ICU. So it probably makes sense to start storing all of scripts data anyway because it seems like we need that here. Should we start making a list of what properties we need to start exposing? Should we expose more and more Unicode properties?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -0,0 +1,602 @@
...
+    case .paragraphSeparator:
+      return consumeScalarGC(.paragraphSeparator)
+    case .spaceSeparator:
+      return consumeScalarGC(.spaceSeparator)
+    }
+
+
+  }
+
+}
```
rxwei, line 602:
```suggestion
    }
  }
}
```

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -0,0 +1,602 @@
...
+    switch kind {
+    case let .scalar(s):
+      return { input, bounds in
+        // TODO: bounds checking?
+        let low = bounds.lowerBound
+        guard input.unicodeScalars[low] == s else {
+          return nil
+        }
+        return input.unicodeScalars.index(after: low)
+      }
```
hamishknight, line 64: Could this use `consumeScalar`?

hamishknight (approved):

Sources/_StringProcessing/ConsumerInterface.swift
milseman: Let's expose any properties that lets you write grapheme breaking as a regular expression for testing purposes :-)

milseman:
Switched all the fatal errors to throws, to at least let us know reason more and better support xfails

milseman:

Sources/_StringProcessing/ConsumerInterface.swift
milseman: Hmm, IIRC they were because Unicode historically used general categories for these kind of queries, but it became apparent that a universal categorization is not great for this use. IIRC `isAlphabetic` is derived from categories and properties, so it might actually be possible to run that in reverse, that is we check `isAlphabetic` and excluding the others.

Sources/_StringProcessing/ConsumerInterface.swift
milseman: We might one day.... but for now we'll punt on these

milseman:

PR #82, opened 2021-12-16
Fix dumping logic for tests

hamishknight:
Make sure we dispatch to AST's children when dumping, implement base dumping for custom character class members, and exclude trivia from dumps for now.

This fixes an issue where tests would incorrectly pass when comparing incorrect AST expectations.


hamishknight:
@swift-ci please test Linux

milseman (approved):
LGTM

PR #86, opened 2021-12-17
Support xfail

milseman:
Make XFAIL .... fail... if the test actually passes.


milseman:

kylemacomber (approved):
Since the bots don't support `XCTExpectedFailure`, I think this is a good direction.

PR #88, opened 2021-12-17
Parse escaped backreferences and subpatterns

hamishknight:
Parse the escaped syntaxes for backreferences and subpatterns (the latter are so syntactically similar, it made sense to also parse them).

This doesn't yet handle the non-escaped syntax for either, in particular Python-style backreferences `(?P=...)`. These are syntactically similar to groups (despite being atoms), so will require some more thought on how to parse.

### DIFF
--- a/Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
@@ -234,7 +234,7 @@ extension Source {
   ///                | 'x'  HexDigit{2}
   ///                | 'U'  HexDigit{8}
   ///                | 'o{' OctalDigit{1...} '}'
-  ///                | '0'  OctalDigit{0...2}
+  ///                | OctalDigit{1...3}
   ///
   mutating func expectUnicodeScalar(
     escapedCharacter base: Character
@@ -257,12 +257,12 @@ extension Source {
         let str = try src.lexUntil(eating: "}").value
         return try Source.validateUnicodeScalar(str, .octal)
 
-      case "0":
+      case let c where c.isOctalDigit:
         // We can read *up to* 2 more octal digits per PCRE.
-        // FIXME: ICU can read up to 3 octal digits, we should have a parser
-        // mode to switch.
-        guard let str = src.tryEatPrefix(maxLength: 2, \.isOctalDigit)?.string
-        else { return Unicode.Scalar(0) }
+        // FIXME: ICU can read up to 3 octal digits if the leading digit is 0,
+        // we should have a parser mode to switch.
+        let nextDigits = src.tryEatPrefix(maxLength: 2, \.isOctalDigit)
+        let str = String(c) + (nextDigits?.string ?? "")
         return try Source.validateUnicodeScalar(str, .octal)
 
       default:
@@ -661,6 +661,10 @@ extension Source {
     }
   }
 
+  /// Try to lex an absolute or relative numbered reference.
+  ///
+  ///     NumberRef -> ('+' | '-')? <Decimal Number>
+  ///
   private mutating func lexNumberedReference(
   ) throws -> Located<Reference>? {
     try recordLoc { src in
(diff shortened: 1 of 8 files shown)


hamishknight:
@swift-ci please test Linux

milseman (approved):
Overall LGTM, though I suspect the parser will be the one to make the call whether `\n` is a backreference or an octal reference. Otherwise the AST needs API to help clients decide.

milseman:

hamishknight:
@swift-ci please test Linux

hamishknight:
Updated to disambiguate octal vs backreferences in the parser, and also extended the `\nnn` syntax to work more generally in custom character classes, how does this look @milseman?

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -633,14 +661,140 @@ extension Source {
...
+      // here.
+      if src.peek() != "0", let digits = src.peekPrefix(\.isNumber) {
+        // First lex out the decimal digits and see if we can treat this as a
+        // backreference.
+        let num = try Source.validateNumber(digits.string, Int.self, .decimal)
+        if num < 10 || digits.first == "8" || digits.first == "9" ||
+            num <= numberOfPriorGroups {
+          src.advance(digits.count)
+          return .backreference(.absolute(num))
+        }
```
milseman, line 784: Should we just return the octal literal?

milseman (approved):
LGTM

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -704,7 +861,8 @@ extension Source {
   ///     ExpGroupStart -> '(_:'
   ///
   mutating func lexAtom(
-    isInCustomCharacterClass customCC: Bool
+    isInCustomCharacterClass customCC: Bool,
+    numberOfPriorGroups: Int
```
rxwei, line 865: Naming convention nit: `...Count` is much more commonly used than `numberOf...`. Suggest `priorGroupCount:`.

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
hamishknight: We can do, though we'd need to call into `expectUnicodeScalar` as we can only read up to 3 digits of octal, and it wouldn't cover the `\0nn` case unless we also add a condition for that here

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
milseman: Where is that logic when we fall through?

hamishknight: In `expectUnicodeScalar`, called from `expectEscaped`

milseman: It seems more straight-forward to make the explicit call, especially if we have context-local constraints on the number of digits or other aspects of interpretation. Otherwise, would the general case have to reconstruct this logic?

Fall-through also means that any intermediary code may have to be concerned with or reason about this possibility. In general I'm a fan of local reasoning. But this can be done after this PR, and it's not a huge deal.

hamishknight: > Otherwise, would the general case have to reconstruct this logic?

I might not be following what you mean by this, but I don't *think* so. Both the backreference and unicode scalar cases have differences in what they accept, but there is a syntactic ambiguity between them. In general, we just need to make sure to try lex as a backreference before a unicode scalar. IMO if we change this logic to produce a scalar, we should also change the scalar logic to be able to produce a backreference so we no longer need to reason about which is called first. I'm happy to do that in a follow-up, but going to merge this for now to cleanup the option parsing PR

hamishknight:
@swift-ci please test Linux

PR #92, opened 2021-12-20
Produce capture structures for regex literal type inference.

rxwei:
Produce capture structures for regex literal type inference.

Introduce `CaptureStructure`, a tree representing the structure (i.e. not the value) of captures. This will be used by the compiler to infer the capture types of a regular expression literal.

`libswiftParseRegexLiteral(_:_:_:_:_:)` encodes a capture structure to a byte sequence to hand it off to C++. The Swift type checker deserializes it to form actual Swift types. The serialization format is as follows:

```
encode(〚`T`〛) ==> <version>, 〚`T`〛, .end
〚`T` (atom)〛 ==> .atom
〚`name: T` (atom)〛 ==> .atom, `name`, '\0'
〚`[T]`〛 ==> 〚`T`〛, .formArray
〚`T?`〛 ==> 〚`T`〛, .formOptional
〚`(T0, T1, ...)`〛 ==> .beginTuple, 〚`T0`〛, 〚`T1`〛, ..., .endTuple
```

### DIFF
--- a/Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
+++ b/Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
@@ -1,0 +1,258 @@
+// A tree representing the type of some captures.
+public enum CaptureStructure: Equatable {
+  case atom(name: String? = nil)
+  indirect case array(CaptureStructure)
+  indirect case optional(CaptureStructure)
+  indirect case tuple([CaptureStructure])
+
+  public static func tuple(_ children: CaptureStructure...) -> Self {
+    tuple(children)
+  }
+
+  public static var empty: Self {
+    .tuple([])
+  }
+}
+
+extension AST {
+  public var captureStructure: CaptureStructure {
+    // Note: This implementation could be more optimized.
+    switch self {
+    case .alternation(let alternation):
+      assert(alternation.children.count > 1)
+      return alternation.children
+        .map(\.captureStructure)
+        .reduce(.empty, +)
+        .map(CaptureStructure.optional)
+    case .concatenation(let concatenation):
+      assert(concatenation.children.count > 1)
+      return concatenation.children.map(\.captureStructure).reduce(.empty, +)
+    case .group(let group):
+      let innerCaptures = group.child.captureStructure
+      switch group.kind.value {
+      case .capture:
+        return .atom() + innerCaptures
+      case .namedCapture(let name):
+        return .atom(name: name.value) + innerCaptures
+      default:
+        return innerCaptures
+      }
+    case .quantification(let quantification):
+      return quantification.child.captureStructure.map(
+        quantification.amount.value == .zeroOrOne
+          ? CaptureStructure.optional
+          : CaptureStructure.array)
+    case .groupTransform:
+      fatalError("Unreachable. Case will be removed later.")
+    case .quote, .trivia, .atom, .customCharacterClass, .empty:
+      return .empty
+    }
+  }
+}
+
+// MARK: - Combination and transformation
+
+extension CaptureStructure {
+  /// Returns a capture structure by concatenating any tuples in `self` and
+  /// `other`.
+  func concatenating(with other: CaptureStructure) -> CaptureStructure {
+    switch (self, other) {
+    // (T...) + (U...) ==> (T..., U...)
+    case let (.tuple(lhs), .tuple(rhs)):
+      return .tuple(lhs + rhs)
+    // T + () ==> T
+    case (_, .tuple(let rhs)) where rhs.isEmpty:
+      return self
+    // () + T ==> T
+    case (.tuple(let lhs), _) where lhs.isEmpty:
+      return other
+    // (T...) + U ==> (T..., U)
+    case let (.tuple(lhs), _):
+      return .tuple(lhs + [other])
+    // T + (U...) ==> (T, U...)
+    case let (_, .tuple(rhs)):
+      return .tuple([self] + rhs)
+    // T + U ==> (T, U)
+    default:
+      return .tuple([self, other])
+    }
+  }
+
+  static func + (
+    lhs: CaptureStructure, rhs: CaptureStructure
+  ) -> CaptureStructure {
+    lhs.concatenating(with: rhs)
+  }
+
+  /// Returns a capture structure by transforming any tuple element of `self`
+  /// or transforming `self` directly if it is not a tuple.
+  func map(
+    _ transform: (CaptureStructure) -> CaptureStructure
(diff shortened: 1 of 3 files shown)


rxwei:
@swift-ci please test linux

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
```diff
@@ -0,0 +1,196 @@
...
+    case end           = 0
+    case atom          = 1
+    case namedAtom     = 2
+    case formArray     = 3
+    case formOptional  = 4
+    case beginTuple    = 5
+    case endTuple      = 6
+  }
+
+  /// Encode the capture structure to the given buffer as a serialized
```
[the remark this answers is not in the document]
milseman: (Sometimes also a special initial byte or byte sequence, which can be a fun parity-checking easter egg).

Sources/_MatchingEngine/Regex/Parse/Parse.swift
```diff
@@ -78,22 +88,28 @@ extension Parser {
   ///     Regex        -> '' | Alternation
   ///     Alternation  -> Concatenation ('|' Concatenation)*
   ///
-  mutating func parse() throws -> AST {
+  mutating func parse() throws -> ParseResult {
```
milseman, line 91: @hamishknight how would this impact things like isolated options, which kick off a recursive parse to end-of-group, but whose capture structure wouldn't nest because of  this?

milseman: Also, why are we complicating parsing for something that can be trivially derived from the AST?

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
```diff
@@ -0,0 +1,196 @@
...
+  /// Returns a capture structure by transforming any tuple element to an
+  /// optional thereof. This is used specifically for forming the capture type
+  /// of an alternation.
+  func alternating() -> CaptureStructure {
+    // (T...) ==> (T?...)
+    if case .tuple(let children) = self {
+      return .tuple(children.map(CaptureStructure.optional))
+    }
+    // T ==> T
+    return self
```
hamishknight, line 54: Shouldn't the capture for e.g `(\d)|a` become optional?

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
```diff
@@ -0,0 +1,196 @@
...
+        append(.atom)
+      // 〚`name: T` (atom)〛 ==> .atom, `name`, '\0'
+      case .atom(name: let name?):
+        append(.namedAtom)
+        let nameCString = name.utf8CString
+        let nameSlot = UnsafeMutableRawBufferPointer(
+          rebasing: buffer[offset ..< offset+nameCString.count])
+        nameCString.withUnsafeBytes(nameSlot.copyMemory(from:))
+        offset += nameCString.count
+      // 〚`[T]`〛 ==> 〚`T`〛, .formOptional
```
hamishknight, line 123:
```suggestion
      // 〚`[T]`〛 ==> 〚`T`〛, .formArray
```

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
```diff
@@ -0,0 +1,196 @@
...
+    }
+    // T ==> T
+    return self
+  }
+
+  /// Returns a capture structure by transforming any tuple element of `self`
+  /// or transforming `self` directly if it is not a tuple.
+  func compactMap(
+    _ transform: (CaptureStructure) -> CaptureStructure
+  ) -> CaptureStructure {
```
hamishknight, line 61: IMO this seems closer to a regular `map` than a `compactMap`, what do you think?

Sources/_MatchingEngine/Regex/Parse/Parse.swift
hamishknight: I think this should work fine with isolated options as they effectively become implicit non-capturing groups, which shouldn't affect the capture structure AFAICT.

And yeah, IMO it would be nicer if the capture structure computation was done on the AST after parsing. I'm happy for that to be done as a follow-up tho.

rxwei: Well, I initially did perform this on the AST but I didn't want this trivial operation to have to traverse the AST.  It's just a tradeoff between separation of concerns and efficiency.  I don't have a strong opinion either way and am happy to change this (on vacation right now; maybe later tonight).

milseman: For efficiency, I wonder if it would be possible to emit the byte code directly from the AST, rather than build up a tree of individually allocated and ref-counted indirect nodes.

(Also, I expect all the AST enum stuff to dwarf these costs, unless we get off of enums for ASTs too. Either way, I'd rather spend the complexity budget on run-time perf)

rxwei: > I wonder if it would be possible to emit the byte code directly from the AST

This, too, is how I implemented it initially but it made things less readable. For example, the mapping behavior of quantification needs top-down contextual information for emission. With an enum tree the logic is easy to read, which IMO is especially important as we rapidly evolve the prototype.

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
rxwei: Done and tested.

Sources/_MatchingEngine/Regex/Parse/Parse.swift
rxwei: Reverted to an AST pass. I would prefer not to prematurely optimize this flow until the user-facing prototype is mature.

rxwei:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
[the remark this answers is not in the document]
rxwei: Added a 16-bit version (same as Swift module format version).

rxwei:
@swift-ci please clean test Linux

rxwei:
@swift-ci please clean test Linux

rxwei:
@swift-ci please test Linux

rxwei:
@swift-ci please test Linux

hamishknight (approved):

milseman (approved):

PR #95, opened 2021-12-23
Positive and negative look-ahead assertions

milseman:
Compiler support by using save points.

### DIFF
--- a/Sources/_StringProcessing/Compiler.swift
+++ b/Sources/_StringProcessing/Compiler.swift
@@ -79,9 +79,21 @@ class Compiler {
     case .trivia, .empty:
       break
 
-    // FIXME: This can't be right...
     case .group(let g):
-      try emit(g.child)
+      if let lookaround = g.lookaroundKind {
+        try emitLookaround(lookaround, g.child)
+        return
+      }
+
+      switch g.kind.value {
+      case .lookahead, .negativeLookahead,
+          .lookbehind, .negativeLookbehind:
+        fatalError("unreachable")
+
+      default:
+        // FIXME: This can't be right...
+        try emit(g.child)
+      }
 
     case .quantification(let quant):
       try emitQuantification(quant)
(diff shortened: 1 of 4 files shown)


Sources/_StringProcessing/Compiler.swift
```diff
@@ -92,11 +100,69 @@ class Compiler {
...
+      save(restoringAt: success)
+      save(restoringAt: intercept)
+      <sub-pattern>    // failure restores at intercept
+      clearSavePoint   // remove intercept
+      <if negative>:
+        clearSavePoint // remove success
+      fail             // positive->success, negative propagates
+    intercept:
+      <if positive>:
+        clearSavePoint // remove result
```
milseman, line 138:
```suggestion
        clearSavePoint // remove success
```

Sources/_StringProcessing/Compiler.swift
```diff
@@ -101,6 +109,60 @@ class Compiler {
     }
   }
 
+  func compileLookaround(
```
rxwei, line 112:
```suggestion
  func emitLookaround(
```
Nit: How about starting with "emit" as existing subroutines called from `emit(_:)` are named "emit...".

Sources/_StringProcessing/Compiler.swift
```diff
@@ -101,6 +109,60 @@ class Compiler {
...
+
+    case .lookahead, .negativeLookahead:
+      break
+
+    default: fatalError("unreachable")
+    }
+
+    let positive = kind == .lookahead
+    assert(positive || kind == .negativeLookahead)
+    _ = """
```
rxwei, line 128: Can we use a comment instead? What's the benefit of using a string literal?

Sources/_StringProcessing/Compiler.swift
```diff
@@ -79,9 +79,17 @@ class Compiler {
...
       break
 
-    // FIXME: This can't be right...
     case .group(let g):
-      try emit(g.child)
+      switch g.kind.value {
+
+      case .lookahead, .negativeLookahead,
+          .lookbehind, .negativeLookbehind:
+        try compileLookaround(g.kind.value, g.child)
```
rxwei, line 87: We could use a `Lookaround` enum in the AST to group these four cases. Then `compileLookaround(_:_:)` would not need to handle unreachable cases.

rxwei (approved):

Sources/_StringProcessing/Compiler.swift
milseman: Looks prettier in the editor (and sometimes better indentation than a multi-line comment), but I can use comments

Sources/_StringProcessing/Compiler.swift
milseman: Added a `var lookaroundKind: (forwards: Bool, positive: Bool)?` to `Group`. This is probably a little more useful than an enum, which we'd want to immediately destructure or add these APIs to anyways. If we have more uses for more semantic info here, we can make the enum.

milseman:

PR #96, opened 2021-12-23
Generalized bidirectional assertion support

milseman:
And support for most built-in anchors.

### DIFF
--- a/Sources/_MatchingEngine/Engine/Engine.swift
+++ b/Sources/_MatchingEngine/Engine/Engine.swift
@@ -1,6 +1,6 @@
 // Currently, engine binds the type and consume binds an instance.
 // But, we can play around with this.
-public struct Engine<Input: Collection> where Input.Element: Hashable {
+public struct Engine<Input: BidirectionalCollection> where Input.Element: Hashable {
 
   var program: Program<Input>
 
--- a/Sources/_StringProcessing/Compiler.swift
+++ b/Sources/_StringProcessing/Compiler.swift
@@ -108,8 +108,83 @@ class Compiler {
       // We stick quoted content into read-only constant strings
       builder.buildMatchSequence(q.literal)
 
+    case .atom(let a) where a.assertionKind != nil:
+      try emitAssertion(a.assertionKind!)
+
     case .customCharacterClass, .atom:
       throw unsupported(node._dumpBase)
+    }
+  }
+
+  func emitAssertion(_ kind: AST.Atom.AssertionKind) throws {
+    // FIXME: Depends on API model we have... We may want to
+    // think through some of these with API interactions in mind
+    //
+    // This might break how we use `bounds` for both slicing
+    // and things like `firstIndex`, that is `firstIndex` may
+    // need to supply both a slice bounds and a per-search bounds.
+    switch kind {
+    case .startOfSubject:
+      builder.buildAssert { (input, pos, bounds) in
+        pos == input.startIndex
+      }
+
+    case .endOfSubjectBeforeNewline:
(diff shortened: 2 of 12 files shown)


Sources/_MatchingEngine/Engine/Engine.swift
```diff
@@ -1,6 +1,6 @@
 // Currently, engine binds the type and consume binds an instance.
 // But, we can play around with this.
-public struct Engine<Input: Collection> where Input.Element: Hashable {
+public struct Engine<Input: BidirectionalCollection> where Input.Element: Hashable {
```
milseman, line 3: (please ignore until you're back from break) 

@rxwei, I had to make this Bidi for easier anchor or assertions like `\b`. I still want to make some kind of capabilities system where you can use ME on non-bidi or even non-collections provided you don't use those instructions, but that will probably have to be added later.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -108,11 +108,86 @@ class Compiler {
...
   }
 
+  func emitAssertion(_ kind: AST.Atom.AssertionKind) throws {
+    // FIXME: Depends on API model we have... We may want to
+    // think through some of these with API interactions in mind
+    //
+    // This might break how we use `bounds` for both slicing
+    // and things like `firstIndex`, that is `firstIndex` may
+    // need to supply both a slice bounds and a per-search bounds.
+    switch kind {
```
milseman, line 126: (please ignore until you're back from break)

@timvermeulen @natecook1000 . We have a bounds for getting around the slicing problem, and we have a bounds in which we search, but we might need to make those two separate concepts for anchors like `^`, which match at the beginning of the input. If we're always advancing our bounds as part of `firstIndex`, then `^` will _always_ match if it's key-ed off of that.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -108,11 +108,86 @@ class Compiler {
...
+
+    case .firstMatchingPositionInSubject:
+      // TODO: We can probably build a nice model with API here
+      builder.buildAssert { (input, pos, bounds) in
+        pos == bounds.lowerBound
+      }
+
+    case .textSegment:
+      // This we should be able to do!
+      throw unsupported(#"\y (text segment)"#)
```
milseman, line 156: (please ignore until you're back from break)

@Azoy, here's where we could use some simple on-the-fly grapheme breaking. By default, `\y` will match in-between two grapheme clusters.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -108,11 +108,86 @@ class Compiler {
...
+      builder.buildAssert { (input, pos, bounds) in
+        pos == input.endIndex || input[pos].isNewline
+      }
+
+    case .wordBoundary:
+      // TODO: May want to consider Unicode level
+      builder.buildAssert { (input, pos, bounds) in
+        // TODO: How should we handle bounds?
+        CharacterClass.word.isBoundary(
+          input, at: pos, bounds: bounds)
```
milseman, line 178: @Azoy, for [UTS-18](https://www.unicode.org/reports/tr18/#Default_Word_Boundaries), I think that we will want to have different semantic levels for word boundaries, perhaps including the generic Unicode algorithm or even some CLDR-like data.

We'll have to explore the tradeoffs here, but this is a chance for Swift to continue to be Unicode-forward in its design.

milseman:
(merging to unblock other stuff, but still want to keep the reviews and discussion alive)

milseman:

PR #98, opened 2022-01-04
Source locations for alternation pipes and char class ranges

hamishknight:
Add source location info to the AST for alternation pipes, and for the `-` character in custom char class ranges. In addition, add some extra test cases for alternations, including some that test the expected source locations are parsed.

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/AST.swift
+++ b/Sources/_MatchingEngine/Regex/AST/AST.swift
@@ -92,11 +92,20 @@ extension AST {
 
   public struct Alternation: Hashable, _ASTNode {
     public let children: [AST]
-    public let location: SourceLocation
+    public let pipes: [SourceLocation]
 
-    public init(_ mems: [AST], _ location: SourceLocation) {
+    public init(_ mems: [AST], pipes: [SourceLocation]) {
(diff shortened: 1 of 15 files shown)


Sources/_MatchingEngine/Regex/AST/AST.swift
```diff
@@ -88,11 +92,20 @@ extension AST {
...
+      // An alternation must have at least two branches (though the branches
+      // may be empty AST nodes), and n - 1 pipes.
+      precondition(mems.count >= 2)
+      precondition(pipes.count == mems.count - 1)
+
       self.children = mems
-      self.location = location
+      self.pipes = pipes
+      self.location = SourceLocation(
+        mems.first!.location.start ..< mems.last!.location.end
```
milseman, line 107: Could this be a computed property instead, then?

milseman (approved):
LGTM

Sources/_MatchingEngine/Regex/AST/AST.swift
hamishknight: Ah yes fair point

hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

PR #101, opened 2022-01-10
Correct \x lexing

hamishknight:
As per PCRE, we eat up to 2 digits, with a fun behavior where no digits is interpreted as 0.


hamishknight:
@swift-ci please test Linux

milseman (approved):
LGTM

PR #103, opened 2022-01-11
Enable resilience for _MatchingEngine and _StringProcessing.

rxwei:
_MatchingEngine and _StringProcessing are built with resilience enabled in Swift core libs. Currently the compiler build produces lots of warnings of missing `@unknown default`. We enable resilience in this package so that we can catch these issues early.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,13 +31,19 @@ let package = Package(
         // Targets can depend on other targets in this package, and on products in packages which this package depends on.
         .target(
             name: "_MatchingEngine",
-            dependencies: [/*"_Unicode"*/]),
+            dependencies: [/*"_Unicode"*/],
+            swiftSettings: [
+                .unsafeFlags(["-enable-library-evolution"])
+            ]),
         .testTarget(
             name: "MatchingEngineTests",
             dependencies: ["_MatchingEngine"]),
         .target(
             name: "_StringProcessing",
-            dependencies: ["_MatchingEngine"]),
+            dependencies: ["_MatchingEngine"],
+            swiftSettings: [
+                .unsafeFlags(["-enable-library-evolution"])
+            ]),
         .target(
             name: "_Unicode",
             dependencies: []),
--- a/Sources/_MatchingEngine/Regex/AST/AST.swift
+++ b/Sources/_MatchingEngine/Regex/AST/AST.swift
@@ -1,4 +1,5 @@
 /// A regex abstract syntax tree
+@frozen
 public indirect enum AST:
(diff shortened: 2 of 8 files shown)


rxwei:
@swift-ci please test Linux

rxwei:
:(

```console
19:18:12 [144/179] Compiling _StringProcessing CharacterClass.swift
19:18:12 error: compile command failed due to signal 6 (use -v to see invocation)Begin Error in Function: '$s17_StringProcessing7compile_7optionsAA6RECodeV15_MatchingEngine3ASTO_AA9REOptionsVtKF0C4NodeL_yyAHKF'
19:18:12 Error! Found a leak due to a consuming post-dominance failure!
19:18:12 Value:   %90 = load [take] %89 : $*{ var AST.Atom }      // users: %286, %242, %158, %143, %123, %91
19:18:12 Post Dominating Failure Blocks:
19:18:12  bb15
19:18:12 End Error in Function: '$s17_StringProcessing7compile_7optionsAA6RECodeV15_MatchingEngine3ASTO_AA9REOptionsVtKF0C4NodeL_yyAHKF'
19:18:12 Found ownership error?!
19:18:12 triggering standard assertion failure routine
19:18:12 
```

milseman:
In the end, do we want to do this or would we want our sources to be slurped into the stdlib's Swift module? We'll need to figure out a direction regarding `_Unicode`, as we'll need to be able to call the same internal functionality as the stdlib.

rxwei:
IMO this should be part of the Swift module when it also becomes a package.  Either way, the package will be built with resilience enabled in the compiler, and I think the package build config should be as close to that as possible.

rxwei:
This is failing with DEVELOPMENT-SNAPSHOT-2022-01-09-a (rdar://87427450).  Note if I simply added `-enable-library-evolution` _without_ fixing warnings by adding `@unknonwn default`:, the package compiles successfully.

milseman (approved):
I’d slightly prefer sticking frozen on the internal module’s enums, but otherwise LGTM

rxwei:
> I’d slightly prefer sticking frozen on the internal module’s enums,

I'll do that for the internal ones. `CharacterClass` however is the one that's triggering the SILGen bug, and it's public, so I still need to fix the bug.

rxwei:
Ok, added a bunch of `@frozen` and there's no need for any `@unknown default:` now.

rxwei:
@swift-ci please test linux and merge

rxwei:
@swift-ci please test linux

PR #110, opened 2022-01-13
WIP: DSL AST

milseman:
WIP of a tree representation for the result builder DSL.

- Gets `case groupTransform` off the literal AST
- Gets `makeAST` off of `CharacterClass`, which can just use a consumer or predicate interface instead
- Compiles most of the tree

Still need to hook up compilation, but shouldn't be too hard from here.


Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -123,6 +123,12 @@ struct VariadicsGenerator: ParsableCommand {
       import _MatchingEngine
 
 
+      private func concat(
```
rxwei, line 126: It would be simpler if we change code generation to output `.concatenation(...)` instead of `concat(...)`.

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -13,13 +13,22 @@ import _MatchingEngine
 
 // MARK: - Primitives
 
+private func concat(
```
rxwei, line 16: Looks unused, and not sure if it's any better than `.concatenation(...)` at call sites.

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -59,16 +69,23 @@ extension CharacterClass: RegexProtocol {
 
 // MARK: Repetition
 
+private func quant(
```
rxwei, line 72: IMO using the enum constructor `.quantification(...)` would be simpler and cleaner than a top-level function.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -17,30 +17,71 @@ struct RegexProgram {
 }
 
 class Compiler {
```
rxwei, line 19: I'm okay with merging the current implementation to unblock things, but the reason I made `Compiler` a class that owns an AST is because we could add environment properties to it later, e.g. function tables, cached `hasCaptures`, etc that are specific to the current AST being compiled.  As such it doesn't seem ideal to make `Compiler` stateless to accommodate two types of ASTs.

rxwei (approved):

Sources/_StringProcessing/Compiler.swift
milseman: Do you think you an sketch out what you were thinking a bit more?

rxwei: I haven't thought very carefully about this... Perhaps the `Compiler` here can act as the shared state that's generic over an AST-specific struct that actually performs code emission.

milseman: Long-term, we'll have an optimizable IR (which is also our ABI) and lowering/emission would be a separate phase. I think it makes sense for phases to have their own state, and there might be some whole-program state. Typically I'd expect lowering or instruction selection to have self-contained state.

I think something interesting and under-explored-so-far is our caching strategy for run-time compilation. I know the `Regex` type has a program in it, but I don't think we've interwoven options (which affect compilation) into it yet.

PR #102, opened 2022-01-10
Add tests for Character semantics in regex matching

natecook1000:
Several of these tests are xfail'd, as we haven't implemented this
level of character semantics yet.

PR #112, opened 2022-01-13
Add options support to the compiler

natecook1000:
This adds support for tracking options to the Regex compiler. Most options still don't have any effect, but I've added support for the `s` flag (single-line) and limited support for `u` (Unicode scalar semantics).

Note: This is based on top of #102.


Sources/_StringProcessing/Compiler.swift
```diff
@@ -42,11 +46,6 @@ class Compiler {
   func emit(_ node: AST) throws {
 
     switch node {
-    // Any: .
-    //     consume 1
-    case .atom(let a) where a.kind == .any && matchLevel == .graphemeCluster:
-      builder.buildAdvance(1)
-
```
natecook1000, line 49: The behavior of "any" depends of the current matching options, so we can't just advance here.

Sources/_StringProcessing/CharacterClass.swift
```diff
@@ -338,8 +338,6 @@ extension AST.Atom {
     switch kind {
     case let .escaped(b): return b.characterClass
 
-    case .any: return .any
-
```
natecook1000, line 342: Removing this means that the `.any` case will be handled using a consumer function, rather than the old `CharacterClass` type.

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
```diff
@@ -79,6 +93,74 @@ extension AST {
...
+        
+    // PCRE options
+    public static var caseInsensitive: Self { .init(.caseInsensitive) }
+    public static var allowDuplicateGroupNames: Self { .init(.allowDuplicateGroupNames) }
+    public static var multiline: Self { .init(.multiline) }
+    public static var noAutoCapture: Self { .init(.noAutoCapture) }
+    public static var singleLine: Self { .init(.singleLine) }
+    public static var reluctantByDefault: Self { .init(.reluctantByDefault) }
+    public static var extended: Self { .init(.extended) }
+    public static var extraExtended: Self { .init(.extraExtended) }
```
milseman, line 136: Who is calling these?

Sources/_StringProcessing/Compiler.swift
```diff
@@ -107,6 +109,18 @@ class Compiler {
...
+        let updated = optionSequence.options(merging: currentOptions)
+        
+        if isIsolated {
+          optionStack.replaceTop(updated)
+          try emit(g.child)
+        } else {
+          optionStack.push(updated)
+          try emit(g.child)
+          optionStack.pop()
+        }
```
milseman, line 122: Who not just have both replace top if we're always pushing/popping at the top?

Sources/_StringProcessing/Compiler.swift
natecook1000: In addition to the semantic level, this also needs to take into account whether the regex is in single-line mode — if not, `.` should match line breaks, so there's two levels of checks to make here. In single-line mode we could use the `advance` approach, though.

milseman: Let's go with the call-out to a closure for now, and we can add real engine support in the future.

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
```diff
@@ -79,6 +93,74 @@ extension AST {
       self.minusLoc = minusLoc
       self.removing = removing
     }
+    
+    public func options(merging optionSet: MatchingOptionSet = []) -> MatchingOptionSet {
+      var result = optionSet
+      for opt in adding {
```
hamishknight, line 99: `MatchingOptionSequence` also has a bit that indicates whether it clears out all the prior matching options except the ones it's adding, currently it's stored as `caretLoc != nil` (really it should have a proper API though). Would it be worth adding support for that feature here so we can handle e.g `(?^i)`?

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
```diff
@@ -36,6 +36,11 @@ extension AST {
       // be unset, only flipped between)
       case textSegmentGraphemeMode  // y{g}
       case textSegmentWordMode      // y{w}
+      
+      // Swift semantic matching level
+      case graphemeClusterSemantics // X
+      case unicodeScalarSemantics   // u
+      case byteSemantics            // b
```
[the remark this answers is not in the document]
hamishknight: Yeah these are Swift-specific, they don't conflict with the matching options from other engines tho AFAIK

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
```diff
@@ -53,6 +58,15 @@ extension AST {
         return false
       }
     }
+    
+    public var isSemanticMatchingLevel: Bool {
+      switch kind {
+      case .graphemeClusterSemantics, .unicodeScalarSemantics, .byteSemantics:
```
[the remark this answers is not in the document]
natecook1000: The only existing engine I've found that has this semantic switching is Rust's `regex::bytes`, which lets you switch from the default byte-matching mode into a Unicode mode with `(?u)`.

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -550,6 +550,11 @@ extension Source {
         try src.expect("}")
         return opt
 
+      // Swift semantic level options
+      case "X": return advanceAndReturn(.graphemeClusterSemantics)
+      case "u": return advanceAndReturn(.unicodeScalarSemantics)
+      case "b": return advanceAndReturn(.byteSemantics)
+
```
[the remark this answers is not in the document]
natecook1000: These would be Swift-specific flags. The `X` flag is to match the `\X` metacharacter, which matches a grapheme cluster; given that, it might be better to go with `O` for Unicode scalar semantics to match `\O`. I'll go through this when I write up the flags.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -474,6 +488,47 @@ class Compiler {
   }
 }
 
+/// A stack of `MatchingOptionSet`s.
+fileprivate struct MatchingOptionSetStack {
+  internal var stack: [AST.MatchingOptionSet]
```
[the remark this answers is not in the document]
natecook1000: I moved `MatchingOptionSet` up to the compiler's level, removed the OptionSet conformance, and just gave it the API that the compiler needs. `MatchingOptionSetStack` now enforces the non-empty requirement.

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -77,10 +72,13 @@ extension AST.Atom {
   }
 
   func generateConsumer(
-    _ opts: CharacterClass.MatchLevel
+    _ opts: AST.MatchingOptionSet
   ) throws -> Program<String>.ConsumeFunction? {
     // TODO: Wean ourselves off of this type...
-    if let cc = self.characterClass?.withMatchLevel(opts) {
+    let matchLevel: CharacterClass.MatchLevel = opts.contains(.unicodeScalarSemantics)
```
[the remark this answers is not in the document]
natecook1000: Right, I think we need to get rid of `CharacterClass.MatchLevel` altogether. I'll handle that in a different change.

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
natecook1000: Oh yeah, this is the right place for that! I'll add that.

natecook1000: Done — I also implemented the `U` option, which has an effect outside of a consumer closure, to see the effects in testing better.

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
```diff
@@ -12,7 +12,7 @@
 extension AST {
   /// An option written in source that changes matching semantics.
   public struct MatchingOption: Hashable {
-    public enum Kind {
+    public enum Kind: Int {
```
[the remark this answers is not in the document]
natecook1000: When this is `Int`-backed, we can use the raw value as the bit offset for `MatchingOptionSet.rawValue`.

Sources/_StringProcessing/ConsumerInterface.swift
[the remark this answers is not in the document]
milseman: Using a set-membership check of a _syntactic_ enum feels a bit like a smell. What if there are multiple ways to spell it, or it came in via API?

Should this instead be a `isScalarSemantic: Bool` on the regex options model type, or perhaps this match level enum is vended by that type?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -77,10 +72,13 @@ extension AST.Atom {
...
   func generateConsumer(
-    _ opts: CharacterClass.MatchLevel
+    _ opts: MatchingOptionSet
   ) throws -> Program<String>.ConsumeFunction? {
     // TODO: Wean ourselves off of this type...
-    if let cc = self.characterClass?.withMatchLevel(opts) {
+    let matchLevel: CharacterClass.MatchLevel = opts.contains(.unicodeScalarSemantics)
+      ? .unicodeScalar
+      : .graphemeCluster
+    if let cc = self.characterClass?.withMatchLevel(matchLevel) {
```
milseman, line 81: Yeah, let's get that logic above into the richer semantic type....

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -109,9 +107,20 @@ extension AST.Atom {
         // TODO: alias? casing?
         $0.name == name || $0.nameAlias == name
       }
+      
+    case .any:
+      return { input, bounds in
+        let curIndex = bounds.lowerBound
+        if !opts.contains(.singleLine) && input[curIndex].isNewline {
```
milseman, line 114: Again, `.singleLine` is one particular spelling. It might be that a 1-1 happens to fall out, but it's a bit of a smell. We could have a `var isSingleLine: Bool` or a `var lineMode: LineMode` enum on the options type.

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -545,7 +554,7 @@ extension Unicode.POSIXProperty {
 extension Unicode.ExtendedGeneralCategory {
   // FIXME: Semantic level
   func generateConsumer(
-    _ opts: CharacterClass.MatchLevel
+    _ opts: MatchingOptionSet
```
milseman, line 557: Is there any practical difference between the type that does stack logic and the this type?

Sources/_StringProcessing/MatchingOptionSet.swift
```diff
@@ -0,0 +1,114 @@
...
+      self.rawValue |= 1 << kind.rawValue
+    }
+  }
+
+  fileprivate mutating func remove(_ kind: Self) {
+    self.rawValue &= ~kind.rawValue
+  }
+  
+  fileprivate mutating func insert(_ kind: Self) {
+    self.rawValue |= kind.rawValue
```
milseman, line 36: assertions?

Sources/_StringProcessing/MatchingOptionSet.swift
```diff
@@ -0,0 +1,114 @@
...
+import _MatchingEngine
+
+/// A set of matching options.
+struct MatchingOptionSet: RawRepresentable {
+  var rawValue: UInt32
+}
+
+extension MatchingOptionSet {
+  init(_ kind: AST.MatchingOption.Kind) {
+    self.rawValue = 1 << kind.rawValue
```
milseman, line 21: Hmm.... are we certain that syntactic options correspond 1-1 to semantic concerns?

Sources/_StringProcessing/MatchingOptionSet.swift
```diff
@@ -0,0 +1,114 @@
...
+  init(_ initial: MatchingOptionSet) {
+    self.stack = [initial]
+  }
+
+  private func _invariantCheck() {
+    assert(!stack.isEmpty, "Unbalanced matching options pop")
+  }
+  
+  var top: MatchingOptionSet {
+    _invariantCheck()
```
milseman, line 94: Typically, it's better to call `_invariantCheck()` at the end of every init, usually by having every init relegate to a common raw-value init, i.e. we're laying our initializations to always check our assumptions.

Then, we only have to make sure that mutating methods, which don't boil down to new `.init` calls, check that these invariants are maintained at the end.

I.e. ideally we'd structure code such that we'd never need to call it from a getter, unless there's some way for mutations to occur outside the domain of its interfaces.

Sources/_StringProcessing/MatchingOptionSet.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+    for opt in sequence.removing {
+      result.remove(.init(opt.kind))
+    }
+    return result
+  }
+}
+
+/// A never-empty stack of `MatchingOptionSet`s.
+struct MatchingOptionSetStack {
```
milseman, line 82: From the compilers perspective, why are there two separate types? It could be that we can further abstract/encapsulate both the concept of what the current options say we should emit and the concept of scopes and option switching.

milseman (commented):
This looks a lot better after the refactoring, thanks!

I think we should make a more semantically rich type for the compiler. E.g. a type `MatchingOptions` which can give a semantic-level enum, a line-handling-strategy enum, assist in choosing what character class definition to use, etc. 

It could have a `mutating func beginScope(AST.Options)`, `mutating func endScope()`, and `mutating func override/merge(AST.Options)`. This model might naturally support hierarchical options from API too. (that is, API would be implemented in terms of this, not that this would necessarily _be_ the API).

We probably want to get as far away from thinking in terms of AST options as we can at this level, other than how to lower them into this representation. There are other options than just those in a literal and other ways to spell or specify options (e.g. http://pcre.org/current/doc/html/pcre2unicode.html). Some options have their own state machine governing them, some don't, etc.

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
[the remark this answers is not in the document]
natecook1000: In Oniguruma, `.textSegmentGraphemeMode` and `.textSegmentWordMode` are spelled `y{g}` and `y{w}`, and change the behavior of `\y` and `\Y` between marking grapheme boundaries and marking word boundaries (as defined by https://unicode.org/reports/tr29/). That in turn changes the behavior of `\X`, since it consumes up to the next `\y` boundary.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -452,7 +455,18 @@ class Compiler {
...
+    let kind: AST.Quantification.Kind
+    if currentOptions.contains(.reluctantByDefault)
+        && quant.kind.value != .possessive
+    {
+      kind = quant.kind.value == .eager
+        ? .reluctant
+        : .eager
+    } else {
+      kind = quant.kind.value
+    }
```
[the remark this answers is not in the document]
natecook1000: Right, it's either the default (`/a+/`) or it's flipped and you use the "reluctant" syntax for eager (`/(?U)a+?/`).

natecook1000:
@milseman Thanks for the feedback! I've moved to a single `MatchingOptions` type that provides API to begin and end scope as well as to query the specific, supported behavior that the options represent. I'm still using the `AST.MatchingOption` enum as the source of the option set bit offsets; there's still plenty of room for any additional options we might need to add, or we can drop that connection if it becomes a problem.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -110,6 +106,10 @@ class Compiler {
         try emit(g.child)
         builder.buildEndCapture(cap)
 
+      case .changeMatchingOptions(let optionSequence, _):
+        options.replaceCurrent(optionSequence)
```
milseman, line 110: Is this a replacement, or a merging-in?

Sources/_StringProcessing/Compiler.swift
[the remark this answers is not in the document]
milseman: Would it be more broadly useful for this to be a method on the options? Something like `remap(_ kind: AST.Quantification.Kind) -> AST.Quantification.Kind`?

I'm also thinking that we might want to pull some of these concepts out of the AST, especially if we want to surface them through the DSL. My DSLTree prototypes have these inside them.

Sources/_StringProcessing/Compiler.swift
```diff
@@ -477,6 +513,19 @@ class Compiler {
   }
 }
 
+// Deprecated matchLevel-based initializer
```
milseman, line 516: Who calls this?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -77,10 +72,10 @@ extension AST.Atom {
   }
 
   func generateConsumer(
-    _ opts: CharacterClass.MatchLevel
+    _ opts: MatchingOptions
   ) throws -> Program<String>.ConsumeFunction? {
     // TODO: Wean ourselves off of this type...
-    if let cc = self.characterClass?.withMatchLevel(opts) {
+    if let cc = self.characterClass?.withMatchLevel(opts.matchLevel) {
```
milseman, line 78: (post this PR) definitely be thinking about how we should model character classes better, given that options can change their definitions

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -109,9 +104,15 @@ extension AST.Atom {
         // TODO: alias? casing?
         $0.name == name || $0.nameAlias == name
       }
+      
+    case .any:
+      fatalError(".atom(.any) is handled in emitAny")
```
milseman, line 109: Note to self: make a `struct Unrecoverable: Error` and throw them

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -0,0 +1,151 @@
...
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
```
milseman, line 11: Thanks for cleanly delineating this!

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -0,0 +1,151 @@
...
+        }
+        if opt.isTextSegmentMode {
+          remove(.textSegmentOptions)
+        }
+        
+        insert(.init(opt.kind))
+      }
+      for opt in sequence.removing {
+        remove(.init(opt.kind))
+      }
```
milseman, line 61: @hamishknight can you double check the logic and make sure this does what you expect it to do?

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -0,0 +1,151 @@
...
+  }
+  
+  mutating func endScope() {
+    _ = stack.removeLast()
+    _invariantCheck()
+  }
+
+  mutating func replaceCurrent(_ sequence: AST.MatchingOptionSequence) {
+    stack[stack.count - 1].merge(with: sequence)
+  }
```
milseman, line 90: So maybe call this `merge`?

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -0,0 +1,151 @@
...
+  
+  var semanticLevel: SemanticLevel {
+    stack.last!.contains(.graphemeClusterSemantics)
+      ? .graphemeCluster
+      : .unicodeScalar
+  }
+}
+
+// Deprecated CharacterClass.MatchLevel API
```
milseman, line 114: Who calls these?

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -0,0 +1,151 @@
...
+
+extension MatchingOptions.Representation {
+  fileprivate init(_ kind: AST.MatchingOption.Kind) {
+    self.rawValue = 1 << kind.rawValue
+  }
+  
+  fileprivate init(unchecked kinds: AST.MatchingOption.Kind...) {
+    self.rawValue = 0
+    for kind in kinds {
+      self.rawValue |= 1 << kind.rawValue
```
milseman, line 148: What if these are mutually exclusive? Can we at least assert, if not merge?

Sources/_StringProcessing/Compiler.swift
```diff
@@ -155,6 +155,31 @@ class Compiler {
...
+        input.unicodeScalars.index(after: bounds.lowerBound)
+      }
+    case (.unicodeScalar, false):
+      builder.buildConsume { input, bounds in
+        input[bounds.lowerBound].isNewline
+          ? nil
+          : input.unicodeScalars.index(after: bounds.lowerBound)
+      }
+    }
+  }
```
milseman, line 182: Do we have tests for the combination of scalar-semantic with different newline behaviors?

milseman (commented):
Looking a lot better! I have a couple questions or clarifications then this looks merge-ready

Sources/_StringProcessing/MatchingOptions.swift
hamishknight: Looks good to me, though I actually realized that the `x` and `xx` options are also mutually exclusive, which we'll likely want to handle here (e.g unsetting `x` unsets `xx`). Though I'm happy to take care of that when implementing the parsing for PCRE extended syntax as we'll need some way of tracking the `x` and `xx` options in the parser for that. Perhaps the parser could use this semantic model to do that, or else we maybe could implement a simpler version of it just for the `x` and `xx` options.

natecook1000: `x` and `xx` are necessarily irrelevant by this stage, right? They affect parsing, but don't have any continued impact during compilation. Do we want to allow them in this grouped/intra-literal position, or should they only be permitted as "global" options?

Sources/_StringProcessing/ConsumerInterface.swift
natecook1000: Yeah, we at least need to transition from `CharacterClass.MatchLevel` to something like the new `MatchingOptions` type.

Sources/_StringProcessing/Compiler.swift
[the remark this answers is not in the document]
natecook1000: Perhaps, but I think the "toggle reluctance" flag only makes sense in the terse jumble of regex literal syntax. We should choose a default and just require always spelling alternatives for the DSL.

Sources/_StringProcessing/MatchingOptions.swift
hamishknight: Yeah they're only relevant to the parser, my thinking was that if we wanted the parser to also be able to use this semantic model to track which of the syntax options is enabled in a given scope, then we could handle them here, or if the parser implements its own logic for that, we could disregard them here and just have this be specific to the matching engine.

> Do we want to allow them in this grouped/intra-literal position, or should they only be permitted as "global" options?

As I understand it, they can be set and unset arbitrarily throughout the regex, e.g `(?x) a b c (?-x)d e f`, only the whitespace in ` a b c ` is treated as non-semantic

hamishknight: Thanks!

> can implement whitespace discarding in a follow-up

Shouldn't this be a case of the parser producing `Trivia` nodes, and the matching engine disregarding them? FWIW I have a patch that implements that https://github.com/apple/swift-experimental-string-processing/commit/a83ca4d6671127402f6d162c94acfc7589a5821e. I'm not sure it could be done completely in the matching engine, as PCRE has some funky rules that means e.g `x * +` is a quantification under extended syntax (my patch doesn't yet implement it tho).

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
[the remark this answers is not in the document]
milseman: What are your thoughts there? This seems relevant to whether we should be using an enum in the AST as the same type/representation for our model.

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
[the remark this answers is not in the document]
milseman: Are there options that make sense for the DSL but are not present in a literal? How divergent is our syntactic representation from our semantic model?

Sources/_StringProcessing/CharacterClass.swift
```diff
@@ -338,12 +338,14 @@ extension AST.Atom {
...
     case .property:
       // TODO: Would our model type for character classes include
       // this? Or does grapheme-semantic mode complicate that?
       return nil
+      
+    case .any:
+      // 'any' is handled by Compiler.emitAny(), not `CharacterClass`
```
milseman, line 347: Why?

milseman: Should be a fatal error then? Unfortunately this is trying to make a CharacterClass, and there's an `any` class, so it's weird to just return `nil` from this.

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/CharacterClass.swift
```diff
@@ -344,11 +344,10 @@ extension AST.Atom {
...
     case .any:
-      // 'any' is handled by Compiler.emitAny(), not `CharacterClass`, to
-      // provide lower level instructions than the CharacterClass-generated
-      // consumer closure
-      // TODO: throw an error instead of trapping here
-      fatalError(".any is handled in Compiler.emitAny()")
+      // `.any` is handled in the matching engine by Compiler.emitAny() and in
+      // the legacy compiler by the `.any` instruction, which can provide lower
+      // level instructions than the CharacterClass-generated consumer closure
+      return nil
```
milseman, line 350:
```suggestion
      // `.any` is handled in the matching engine by Compiler.emitAny() and in
      // the legacy compiler by the `.any` instruction, which can provide lower
      // level instructions than the CharacterClass-generated consumer closure
      //
      // FIXME: We shouldn't be returning `nil` here, but instead fixing the call
      // site to check for any before trying to construct a character class.
      return nil
```

milseman:

milseman (approved):
LGTM

PR #115, opened 2022-01-16
Add in static lookup for character classes in result builders

milseman:


milseman:

milseman:
Merging to unblock some other work, but I am interested in naming that protocol or further refining our approach. I do wonder if there's a place for a `RegexComponent` protocol that would help alleviate some of the name lookup or type checking concerns, but that's for later.

PR #118, opened 2022-01-19
Match API on Substring too

milseman:


milseman:
@swift-ci please test

rxwei (approved):

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -112,10 +112,22 @@ public struct Regex<Match: MatchProtocol>: RegexProtocol {
...
+  public func match(in input: Substring) -> RegexMatch<Match>? {
+    _match(
+      input.base, in: input.startIndex..<input.endIndex)
+  }
+
   // FIXME: This is mostly hacky because we go down two different paths based on
   // whether there are captures. This will be cleaned up once we deprecate the
   // legacy virtual machines.
-  public func match(in input: String) -> RegexMatch<Match>? {
+  func _match(
```
rxwei, line 127: I don’t feel we need an underscore here.

milseman: It's often bad form to have internal methods with the same base-name as public methods. If we end up internalizing all of these or making this one public, then I'd agree.

rxwei: > It's often bad form to have internal methods with the same base-name as public methods.

I've not heard about this practice -- is it documented in any guidelines?

milseman: A less-strict version of https://github.com/apple/swift/blob/main/docs/StandardLibraryProgrammersManual.md#the-leading-underscore-rule

milseman:
I don't think the linux bot is going to free up anytime soon...

milseman:
Merging to unblock

Sources/_StringProcessing/RegexDSL/Core.swift
rxwei: Thanks!  I was coming from framework development so wasn't familiar with the stdlib conventions.

milseman: I'm not arguing it's a _good_ rule, but I see some value when things have the same name. Otherwise I'm not pedantic enough to go as far as the stdlib that all internal entities must have an underscore.

PR #119, opened 2022-01-20
Fix tuple construction memory corruption.

rxwei:
This bug was uncovered when working with transformed captures using the DSL. When advancing the address to the next element in a tuple, we should advance by size rather than stride before aligning.

### DIFF
--- a/Sources/_MatchingEngine/Utility/TypeConstruction.swift
+++ b/Sources/_MatchingEngine/Utility/TypeConstruction.swift
@@ -128,13 +128,13 @@ public enum TypeConstruction {
         // Open existential on each element type.
         func initializeElement<T>(_ element: T) {
           currentElementAddressUnaligned =
-          currentElementAddressUnaligned.roundedUp(toAlignmentOf: T.self)
+            currentElementAddressUnaligned.roundedUp(toAlignmentOf: T.self)
           currentElementAddressUnaligned.bindMemory(
             to: T.self, capacity: MemoryLayout<T>.size
           ).initialize(to: element)
           // Advance to the next element (unaligned).
           currentElementAddressUnaligned =
-          currentElementAddressUnaligned.advanced(by: MemoryLayout<T>.stride)
+            currentElementAddressUnaligned.advanced(by: MemoryLayout<T>.size)
         }
         _openExistential(element, do: initializeElement)
       }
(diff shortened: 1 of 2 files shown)


rxwei:
@swift-ci please test Linux

rxwei:
Note: Ideally, we should use `_forEachField(of:options:body:)` to avoid manual address calculation. However that would prevent the project from building with stock Xcode toolchains as `@_spi(Reflection)` cannot be imported there.

Sources/_MatchingEngine/Utility/TypeConstruction.swift
```diff
@@ -128,13 +128,13 @@ public enum TypeConstruction {
         // Open existential on each element type.
         func initializeElement<T>(_ element: T) {
           currentElementAddressUnaligned =
-          currentElementAddressUnaligned.roundedUp(toAlignmentOf: T.self)
+            currentElementAddressUnaligned.roundedUp(toAlignmentOf: T.self)
```
milseman, line 131:
```suggestion
          currentElementAddressUnaligned.roundedUp(toAlignmentOf: T.self)
```

Sources/_MatchingEngine/Utility/TypeConstruction.swift
```diff
@@ -128,13 +128,13 @@ public enum TypeConstruction {
...
           currentElementAddressUnaligned.bindMemory(
             to: T.self, capacity: MemoryLayout<T>.size
           ).initialize(to: element)
           // Advance to the next element (unaligned).
           currentElementAddressUnaligned =
-          currentElementAddressUnaligned.advanced(by: MemoryLayout<T>.stride)
+            currentElementAddressUnaligned.advanced(by: MemoryLayout<T>.size)
```
milseman, line 137:
```suggestion
          currentElementAddressUnaligned.advanced(by: MemoryLayout<T>.size)
```

milseman (approved):
LGTM!

Sources/_MatchingEngine/Utility/TypeConstruction.swift
rxwei: Why is that? This and the previous line are part of the same statement. So this is line wrapping according to Google Swift Style Guide. Is there a different style guide that wants this done differently?

milseman: Oh, I didn't see that. Argh, I think Xcode refuses to indent it

PR #120, opened 2022-01-20
Fix two crashers with quantification on transformed capture.

rxwei:
This patch fixes two issues:
- When an optional quantification matches 1 occurrence, the capture type constructed by the engine unexpectedly drops the `Optional` wrapper type. `Capture.value` is fixed by including the `Optional` wrapper type.
- When an optional or array quantification matches 0 occurrence, the capture type constructed by the engine unexpectedly uses `Substring` as the atom type instead of the transformed type from `CaptureTransform`. This is fixed by propagating capture transform result type information from the DSL API down to `captureNil` and `captureArray` instructions in the bytecode. The engine will use the type information to construct the correct types for `nil` and `[]`.

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/AST.swift
+++ b/Sources/_MatchingEngine/Regex/AST/AST.swift
@@ -94,10 +94,13 @@ extension AST {
 
   /// Whether this node has nested somewhere inside it a capture
   public var hasCapture: Bool {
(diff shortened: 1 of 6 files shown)


rxwei:
@swift-ci please test Linux

milseman (approved):
LGTM

PR #121, opened 2022-01-20
Introduce `tryCapture(_:)`.

rxwei:
`tryCapture(_:)` works like `capture(_:)` but accepts an optional or throwing capture transform instead. When the transform fails (returns `nil` or throws), it will cause matching to fail. This works well with failable conversion initializers such as `Int.init?(_:)`, `Unicode.Scalar.init?(_:)`, etc.

----

Example:

With `tryCapture`, the following example avoids unnecessary levels of optionals and double optionals.

```swift
line.match {
  OneOrMore(.hexDigit).tryCapture(Unicode.Scalar.init(hex:))
  Optionally {
    ".."
    OneOrMore(.hexDigit).tryCapture(Unicode.Scalar.init(hex:))
  }
  OneOrMore(.whitespace)
  ";"
  OneOrMore(.whitespace)
  OneOrMore(.word).tryCapture(Unicode.GraphemeBreakProperty.init)
  Repeat(.any)
} // Regex<(Substring, Unicode.Scalar, Unicode.Scalar?, Unicode.GraphemeBreakProperty)>
```

### DIFF
--- a/Sources/Exercises/Participants/RegexParticipant.swift
+++ b/Sources/Exercises/Participants/RegexParticipant.swift
@@ -43,6 +43,8 @@ struct RegexLiteralParticipant: Participant {
   }
 }
 
+// MARK: - Regex literal
+
 private func extractFromCaptures(
   _ match: Tuple4<Substring, Substring, Substring?, Substring>
 ) -> GraphemeBreakEntry? {
@@ -63,23 +65,6 @@ private func graphemeBreakPropertyData<RP: RegexProtocol>(
   line.match(regex).map(\.match).flatMap(extractFromCaptures)
 }
 
-private func graphemeBreakPropertyData(
-  forLine line: String
-) -> GraphemeBreakEntry? {
-  graphemeBreakPropertyData(forLine: line, using: Regex {
-    OneOrMore(.hexDigit).capture()
-    Optionally {
-      ".."
-      OneOrMore(.hexDigit).capture()
-    }
-    OneOrMore(.whitespace)
-    ";"
-    OneOrMore(.whitespace)
-    OneOrMore(.word).capture()
-    Repeat(.any)
-  })
-}
-
 private func graphemeBreakPropertyDataLiteral(
   forLine line: String
 ) -> GraphemeBreakEntry? {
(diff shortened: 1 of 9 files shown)


rxwei:
Note: This PR depends on #120 and #119. Please review 36c1a03 for changes specific to `tryCapture(_:)`.

milseman:
Should we have a throw abort? That's the pattern we're planning to follow elsewhere and Perl/PCRE's more general-form callouts have that capability.

Sources/Exercises/Participants/RegexParticipant.swift
```diff
@@ -63,28 +65,34 @@ private func graphemeBreakPropertyData<RP: RegexProtocol>(
...
-private func graphemeBreakPropertyDataLiteral(
-  forLine line: String
-) -> GraphemeBreakEntry? {
-  return graphemeBreakPropertyData(
-    forLine: line,
-    using: r(#"([0-9A-F]+)(?:\.\.([0-9A-F]+))?\s+;\s+(\w+).*"#,
-             matching: Tuple4<Substring, Substring, Substring?, Substring>.self))
+  }.map {
+    let (_, lower, upper, property) = $0.match.tuple
+    print(property)
```
milseman, line 95:
```suggestion
```

Sources/_MatchingEngine/Regex/AST/AST.swift
```diff
@@ -207,14 +210,61 @@ extension AST {
...
+  }
+
+  public init(
+    resultType: Any.Type,
+    _ closure: @escaping (Substring) throws -> Any
+  ) {
+    self.init(resultType: resultType, closure: .throwing(closure))
+  }
+
+  public func callAsFunction(_ input: Substring) -> Any? {
```
milseman, line 247: Should this be `throws` then?

Tests/RegexTests/RegexDSLTests.swift
```diff
@@ -233,3 +316,13 @@ class RegexDSLTests: XCTestCase {
     }
   }
 }
+
+extension Unicode.Scalar {
+  // Convert a hexadecimal string to a scalar
+  public init?<S: StringProtocol>(hex: S) {
```
milseman, line 322:
```suggestion
 init?<S: StringProtocol>(hex: S) {
```

milseman (approved):
Overall LGTM!

Sources/_MatchingEngine/Regex/AST/AST.swift
rxwei: No, this returns an optional for the engine to determine whether to fail matching. When the closure throws, this returns nil.

rxwei: > Should we have a throw abort?
Just saw this comment. In that case I think we can have if.

milseman: SGTM

milseman:
If this passes your local testing, I'm ok merging this.

@shahmishal any idea why Linux CI can't get us a machine? We run in around 1min total time, so we're not going to clog up any pipelines.

rxwei:
I haven't triggered tests here, as #119 needed to be merged first.

rxwei:
@swift-ci please test Linux

PR #122, opened 2022-01-20
Initial matcher/validator support

timvermeulen:
Adds `MatchingCollectionConsumer` and `MatchingCollectionSearcher` as a matching/validating analogues of `CollectionConsumer` and `CollectionSearcher`:

```swift
public protocol MatchingCollectionConsumer: CollectionConsumer {
  associatedtype Value

  func matchingConsuming(
    _ consumed: Consumed,
    in range: Range<Consumed.Index>
  ) -> (Value, Consumed.Index)?
}

public protocol MatchingCollectionSearcher: CollectionSearcher {
  associatedtype Value

  func matchingSearch(
    _ searched: Searched,
    _ state: inout State
  ) -> (Value, Range<Searched.Index>)?
}
```

...as well as matching equivalents of the other consumer/searcher protocols.

Also adds a few algorithms that work with matches:

* `firstMatch`/`lastMatch`
* `matches(of:)`
* `replacing(_:with:)` using a closure:

```swift
let regex = Regex(OneOrMore(.digit).capture { Int($0)! })
let str = "foo 160 bar 99 baz"
print(str.replacing(regex, with: { match, _ in String(match.1, radix: 8) }))
// foo 240 bar 143 baz
```

### DIFF
--- a/Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
+++ b/Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
@@ -82,7 +82,7 @@ extension RegexConsumer: StatelessCollectionSearcher {
 }
 
 // TODO: Bake in search-back to engine too
-extension RegexConsumer: StatelessBackwardCollectionSearcher {
+extension RegexConsumer: BackwardStatelessCollectionSearcher {
   public typealias BackwardSearched = Consumed
   
   public func searchBack(
--- a/Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
+++ b/Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
@@ -10,23 +10,25 @@
 //===----------------------------------------------------------------------===//
 
 public struct RegexConsumer<
-  Consumed: BidirectionalCollection, Capture: MatchProtocol
+  R: RegexProtocol, Consumed: BidirectionalCollection
 > where Consumed.SubSequence == Substring {
   // TODO: Should `Regex` itself implement these protocols?
-  let regex: Regex<Capture>
+  let regex: R
 
-  public init(_ regex: Regex<Capture>) {
+  public init(_ regex: R) {
     self.regex = regex
   }
-  
+}
+
+extension RegexConsumer {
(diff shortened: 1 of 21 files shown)


Tests/RegexTests/AlgorithmsTests.swift
```diff
@@ -114,6 +114,20 @@ class RegexConsumerTests: XCTestCase {
...
   }
+  
+  func testMatches() {
+    let regex = Regex(OneOrMore(.digit).capture { 2 * Int($0)! })
+    let str = "foo 160 bar 99 baz"
+    XCTAssertEqual(str.matches(of: regex).map(\.0.1), [320, 198])
+  }
+  
+  func testMatchReplace() {
+    let regex = Regex(OneOrMore(.digit).capture { Int($0)! })
```
rxwei, line 125: It would be great if we get an example of find-and-replace with a regex with multiple captures, both with the regex literal and with the result builder syntax. That will help inform our decisions around Swift tuples vs nominal tuples.

timvermeulen:
@swift-ci Please test Linux

Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
```diff
@@ -9,56 +9,54 @@
 //
 //===----------------------------------------------------------------------===//
 
-import _MatchingEngine
+public struct RegexConsumer<
```
milseman, line 12: This looks fine to me for now.

Outside scope of this PR, we might make RegexProtocol itself a consumer/searcher/matcher. But, I think it's best to establish parent-child relationships after we have the fundamental model in place. The important thing is getting this code and conformances in.

BTW, why does this need to be public?

milseman (approved):
LGTM

Sources/_StringProcessing/Algorithms/Consumers/RegexConsumer.swift
timvermeulen: Some consumer/searcher algorithms return a type generic over the searcher, so the regex overloads return a type generic over `RegexConsumer`. So same story as `PredicateConsumer` at the moment.

PR #123, opened 2022-01-21
Parse global matching options and Oniguruma recursion levels

hamishknight:
- Parse the PCRE global matching options e.g `(*CR)`, `(*UTF)`, and `(*LIMIT_DEPTH=d)`. Such options may only appear at the very start of the regex. For now, this forms a new AST node, but in the future we may want to subsume it into a top-level AST type.

- Parse an additional `+n` or `-n` suffix on references in conditionals and the `\k` backreference, which is valid Oniguruma that allows a recursion level to be specified for the reference.

- Also correct parsing for `\k<...>` such that it allows a numeric reference, which is permitted by Oniguruma and .NET.

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/AST.swift
+++ b/Sources/_MatchingEngine/Regex/AST/AST.swift
@@ -38,6 +38,8 @@ public indirect enum AST:
 
   case customCharacterClass(CustomCharacterClass)
 
+  case globalMatchingOptions(GlobalMatchingOptions)
+
   case empty(Empty)
 
   // FIXME: Move off the regex literal AST
@@ -55,16 +57,17 @@ extension AST {
   // over `self` _everywhere_ we want to do anything.
   var _associatedValue: _ASTNode {
     switch self {
-    case let .alternation(v):          return v
-    case let .concatenation(v):        return v
-    case let .group(v):                return v
-    case let .conditional(v):          return v
-    case let .quantification(v):       return v
-    case let .quote(v):                return v
-    case let .trivia(v):               return v
-    case let .atom(v):                 return v
-    case let .customCharacterClass(v): return v
-    case let .empty(v):                return v
+    case let .alternation(v):           return v
+    case let .concatenation(v):         return v
+    case let .group(v):                 return v
+    case let .conditional(v):           return v
+    case let .quantification(v):        return v
+    case let .quote(v):                 return v
+    case let .trivia(v):                return v
+    case let .atom(v):                  return v
+    case let .customCharacterClass(v):  return v
+    case let .empty(v):                 return v
+    case let .globalMatchingOptions(v): return v
 
     case let .groupTransform(g, _):
       return g // FIXME: get this out of here
@@ -113,7 +116,7 @@ extension AST {
     case .group, .conditional, .customCharacterClass:
       return true
     case .alternation, .concatenation, .quantification, .quote, .trivia,
-        .empty, .groupTransform:
+        .empty, .groupTransform, .globalMatchingOptions:
       return false
     }
   }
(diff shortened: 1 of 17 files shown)


hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/AST/AST.swift
```diff
@@ -207,19 +210,43 @@ extension AST {
...
     }
 
     /// Whether this is a reference that recurses the whole pattern, rather than
     /// a group.
     public var recursesWholePattern: Bool { kind == .recurseWholePattern }
   }
+
+  /// An AST node containing global matching options along with a child that
+  /// uses those options.
+  /// TODO: Should this be subsumed into a top-level AST type?
```
milseman, line 236: I think it should. The AST is recursive, but this cannot be a member. So I think it might be time for `struct AST` and `enum AST.Node`.

Sources/_MatchingEngine/Regex/AST/Atom.swift
```diff
@@ -76,6 +76,32 @@ extension AST {
...
+    case .startOfLine:                  return nil
+    case .endOfLine:                    return nil
+    }
+  }
+
+  func `as`<T>(_ t: T.Type = T.self) -> T? {
+    _associatedValue as? T
+  }
+}
+
```
milseman, line 104: I wonder if it's useful/interesting to formalize this as some kind of `enum-of-structs` protocol.

Sources/_MatchingEngine/Regex/Parse/Parse.swift
```diff
@@ -116,10 +116,27 @@ extension Parser {
 extension Parser {
   /// Parse a regular expression
   ///
-  ///     Regex        -> '' | Alternation
-  ///     Alternation  -> Concatenation ('|' Concatenation)*
+  ///     Regex        -> GlobalMatchingOption* RecursiveRegex
```
milseman, line 119: I think I slightly prefer this, but let me know what you think:

```
/// RegexLiteral -> GlobalOption* Regex
/// Regex -> ...
```

Sources/_MatchingEngine/Regex/AST/AST.swift
milseman: Can you try prototyping that direction? Feel free to split the global options out of this PR if it helps land stuff.

hamishknight: Yeah I can work on implementing that. I have committed a bunch of stuff on top of this though, so would have to work through a rebase to split off the global options commit. Would you mind if I landed it in its current state and moved it onto `AST` in a follow-up?

Sources/_MatchingEngine/Regex/AST/Atom.swift
hamishknight: Yeah we could do, though it would mean we'd no longer be able to keep `_associatedValue` private, but maybe that doesn't matter as long as it's underscored

Sources/_MatchingEngine/Regex/Parse/Parse.swift
hamishknight: Yeah that crossed my mind, but I felt that `RegexLiteral` sounded like it would cover the grammar including the delimiters. With the implementation of `AST.Node` perhaps instead `RecursiveRegex` could be `RegexNode`? I don't feel strongly about it though, and am fine with `RegexLiteral` & `Regex`.

milseman: `Regex` and `RegexComponent` or `RegexNode` sound good to me.

Sources/_MatchingEngine/Regex/AST/AST.swift
hamishknight: @milseman To be clear, I mean that I have local commits based on top of this that I'd need to rebase. I'm happy to do that, but it'd be more convenient for me to land this as-is.

PR #124, opened 2022-01-23
PEG: remove errant XCTest dependency

compnerd:
Remove errant `import XCTest`.  The core libraries should not depend on the testing framework.


compnerd:
@swift-ci please test

rxwei (approved):

milseman:
I'm happy to remove this, but the basic premise...

> The core libraries should not depend on the testing framework.

None of this code, or any code outside `_MatchingEngine` and `_StringProcessing`, should be part of a core library.

rxwei:
No, we are not building them.

PR #129, opened 2022-01-25
Parse Oniguruma callout and absent function syntax

hamishknight:
Parse both variants of Oniguruma callout syntax `(*name[tag]{args, ...})` and `(?{...}[])`, as well as Oniguruma absent functions `(?~|)`. In addition, diagnose ending `)`s that don't close any groups.

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/AST.swift
+++ b/Sources/_MatchingEngine/Regex/AST/AST.swift
@@ -38,6 +38,8 @@ public indirect enum AST:
 
   case customCharacterClass(CustomCharacterClass)
 
+  case absentFunction(AbsentFunction)
+
   case empty(Empty)
 
   // FIXME: Move off the regex literal AST
@@ -55,16 +57,17 @@ extension AST {
   // over `self` _everywhere_ we want to do anything.
   var _associatedValue: _ASTNode {
     switch self {
-    case let .alternation(v):          return v
-    case let .concatenation(v):        return v
-    case let .group(v):                return v
-    case let .conditional(v):          return v
-    case let .quantification(v):       return v
-    case let .quote(v):                return v
-    case let .trivia(v):               return v
-    case let .atom(v):                 return v
-    case let .customCharacterClass(v): return v
-    case let .empty(v):                return v
+    case let .alternation(v):           return v
+    case let .concatenation(v):         return v
+    case let .group(v):                 return v
+    case let .conditional(v):           return v
+    case let .quantification(v):        return v
+    case let .quote(v):                 return v
+    case let .trivia(v):                return v
+    case let .atom(v):                  return v
+    case let .customCharacterClass(v):  return v
+    case let .empty(v):                 return v
+    case let .absentFunction(v):        return v
 
     case let .groupTransform(g, _):
       return g // FIXME: get this out of here
@@ -110,7 +113,7 @@ extension AST {
     switch self {
     case .atom(let a):
       return a.isQuantifiable
-    case .group, .conditional, .customCharacterClass:
+    case .group, .conditional, .customCharacterClass, .absentFunction:
       return true
     case .alternation, .concatenation, .quantification, .quote, .trivia,
         .empty, .groupTransform:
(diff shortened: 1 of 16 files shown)


hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/AST/AST.swift
```diff
@@ -185,6 +188,43 @@ extension AST {
     }
   }
 
+  public struct AbsentFunction: Hashable, _ASTNode {
```
milseman, line 191: I know Oniguruma uses this "absent function" name, but is there a better or more descriptive name? What does this do?

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -942,6 +967,19 @@ extension Source {
...
 
+  /// Try to consume the start of an absent function.
+  ///
+  ///     AbsentFunctionStart -> '(?~' '|'?
+  ///
+  mutating func lexAbsentFunctionStart(
+  ) -> Located<AST.AbsentFunction.Start>? {
+    recordLoc { src in
+      if src.tryEat(sequence: "(?~|") { return .withPipe }
+      if src.tryEat(sequence: "(?~") { return .withoutPipe }
```
milseman, line 978: Is there a more descriptive name than "with(out) pipe"?

Sources/_MatchingEngine/Regex/Parse/Parse.swift
```diff
@@ -237,19 +254,66 @@ extension Parser {
...
     // An implicit scoped group has already consumed its closing paren.
     if !kind.value.hasImplicitScope {
       try source.expect(")")
     }
     return .init(kind, child, loc(start))
   }
 
+  /// Consume the body of an absent function.
+  ///
+  ///     AbsentFunction -> '(?~' RecursiveRegex ')'
```
milseman, line 267:
```suggestion
  ///     AbsentFunction -> '(?~' RegexNode ')'
```

milseman (approved):
Overall LGTM

Sources/_MatchingEngine/Regex/AST/AST.swift
hamishknight: Hmm, its primary role it to match against an inverse of a pattern (though the `.expression` variant also specifies what it can match). So maybe something like `MatchInverse` or `Exclusion`?

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
hamishknight: I mean the syntax is only used for the repeater syntax, so we could call it `repeater`? Though `withPipe` would then seem out of place, maybe that should be `normal` or something?

Sources/_MatchingEngine/Regex/AST/AST.swift
milseman: Ah, I see. This is Oniguruma's approach to the regex-inversion problem (i.e. inverting each component of a concatenation is surprising and not the same as inverting the regex). I'm fine keeping this name with a comment for now, as a term-of-art or something to be renamed later. 

@rxwei any thoughts on how to model pattern inversions in the DSL?

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
milseman: I think Oniguruma calls them `repeater` and `stopper`, which sounds good enough to me for now.

hamishknight: Note this is just for the opening syntax, `(?~|` is not just for absent stoppers, but also for absent expressions and clearers.

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/LexicalAnalysis.swift
hamishknight: I've left the naming as-is for now, but am happy to change in a follow-up

PR #114, opened 2022-01-14
Reduce `buildBlock` overloads with pairwise `buildBlock(combining:into:)`

rxwei:
This patch replaces `O(2^arity)` overloads of `buildBlock` with `O(arity^2)` overloads of `buildBlock(combining:into:)` with the language feature implemented in apple/swift#40799. This improves library compilation time, and fixes an issue where nested captures did not have a flattened type. Before this patch, fixing this would need `O(arity!)` overloads. This also allows us to switch back to native tuples for `Match`.

Also, rename `renderAsPattern` to `renderAsBuilderDSL` to work around a CI stale file issue.

This depends on swift-DEVELOPMENT-SNAPSHOT-2022-02-03 or later.

PR #130, opened 2022-01-26
Refactor AST into a struct containing AST.Node enum

hamishknight:
Introduce `AST` as a new root type, with `AST.Node` becoming the type used to represent a particular node in the AST. This will allow us to store top-level state on the `AST` type rather than for each `AST.Node`.

### DIFF
--- a/Sources/VariadicsGenerator/VariadicsGenerator.swift
+++ b/Sources/VariadicsGenerator/VariadicsGenerator.swift
@@ -255,7 +255,7 @@ struct VariadicsGenerator: ParsableCommand {
     outputForEach(
       0..<arity, separator: ", ", lineTerminator: ""
     ) { i in
-      "x\(i).\(patternProtocolRequirementName).ast"
+      "x\(i).\(patternProtocolRequirementName).ast.root"
     }
     output("))\n")
     output("  }\n}\n\n")
--- a/Sources/_MatchingEngine/Regex/Parse/Parse.swift
+++ b/Sources/_MatchingEngine/Regex/Parse/Parse.swift
@@ -127,7 +127,7 @@ extension Parser {
       }
       fatalError("Unhandled termination condition")
     }
-    return ast
+    return .init(ast)
   }
 
   /// Parse a regular expression node. This should be used instead of `parse()`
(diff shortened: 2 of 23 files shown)


hamishknight:
@swift-ci please test Linux

Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -118,6 +118,17 @@ struct VariadicsGenerator: ParsableCommand {
...
+      //
+      // This source file is part of the Swift.org open source project
+      //
+      // Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+      // Licensed under Apache License v2.0 with Runtime Library Exception
+      //
+      // See https://swift.org/LICENSE.txt for license information
+      //
+      //===----------------------------------------------------------------------===//
+
```
hamishknight, line 131: @rxwei Does it make sense to include the copyright notice here? As otherwise it gets removed when re-running the generator.

rxwei: Yes it does. I'm already adding it here: https://github.com/apple/swift-experimental-string-processing/pull/114, but the PR is blocked until there's a new toolchain.

rxwei: It might be better to change this so that I won't have to resolve conflicts for my chain of 3 PRs 😂

Sources/_MatchingEngine/Regex/Parse/Parse.swift
```diff
@@ -127,7 +127,7 @@ extension Parser {
       }
       fatalError("Unhandled termination condition")
     }
-    return ast
+    return .init(ast)
```
milseman, line 130: Where do global options come in?

milseman (approved):
LGTM

Sources/_MatchingEngine/Regex/Parse/Parse.swift
hamishknight: We'll parse them here and store them on `AST`

Sources/VariadicsGenerator/VariadicsGenerator.swift
hamishknight: Yeah I was just thinking that, though I think there will still be a conflict as I'm still changing Concatenation.swift, but at least that can be resolved by re-running the generator

hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

milseman:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Parse/Parse.swift
milseman: Oh, right, that will be based on top of this. SGTM

PR #133, opened 2022-01-27
Parse global matching options

hamishknight:
Parse the PCRE global matching options e.g `(*CR)`, `(*UTF)`, and `(*LIMIT_DEPTH=d)`. Such options may only appear at the very start of the regex, and as such are stored on the top-level AST type.

### DIFF
--- a/Sources/_MatchingEngine/Regex/Printing/PrintAsCanonical.swift
+++ b/Sources/_MatchingEngine/Regex/Printing/PrintAsCanonical.swift
@@ -32,12 +32,8 @@ extension AST.Node {
     showDelimiters delimiters: Bool = false,
     terminateLine: Bool = false
   ) -> String {
-    var printer = PrettyPrinter()
-    printer.printAsCanonical(
-       self,
-       delimiters: delimiters,
-       terminateLine: terminateLine)
-    return printer.finish()
+    AST(self, globalOptions: nil).renderAsCanonical(
+      showDelimiters: delimiters, terminateLine: terminateLine)
   }
 }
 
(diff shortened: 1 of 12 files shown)


hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Printing/PrintAsCanonical.swift
```diff
@@ -34,7 +34,7 @@ extension AST.Node {
   ) -> String {
     var printer = PrettyPrinter()
     printer.printAsCanonical(
-       self,
+       .init(self, globalOptions: nil),
        delimiters: delimiters,
        terminateLine: terminateLine)
     return printer.finish()
```
milseman, line 40: My intuition is that we then want to hoist this to `AST` instead of on nodes. The render which takes various options makes sense on the AST so it picks up global options, etc, and the recursive logic is on node.

Tests/RegexTests/ParseTests.swift
```diff
@@ -42,20 +42,33 @@ func parseTest(
...
+  syntax: SyntaxOptions = .traditional,
+  captures expectedCaptures: CaptureStructure = .empty,
+  file: StaticString = #file,
+  line: UInt = #line
 ) {
   let ast = try! parse(input, syntax)
-  guard ast.root == expectedAST
-          || ast.root._dump() == expectedAST._dump() // EQ workaround
+  guard ast == expectedAST
+          || ast._dump() == expectedAST._dump() // EQ workaround
```
milseman, line 61: Argh, this is still here! Any chance of adding a bool parameter for something like "ignore trivia", which we can approximate with the dump check? IIRC, this was mostly for trivia like non-semantic whitespace.

milseman (approved):
Overall LGTM. I think it's worth thinking (briefly and in a follow-up PR!) about what the pretty-print APIs should be, but this seems fine as it is.

Sources/_MatchingEngine/Regex/Printing/PrintAsCanonical.swift
hamishknight: There is an overload for `AST` too, this overload is mainly for convenience e.g a few call sites doing `throw unsupported(node.renderAsCanonical())`. Though we could reduce the duplication here by having it call into the other overload, let me change that.

Tests/RegexTests/ParseTests.swift
hamishknight: Yeah this is awful, IIRC it's done to ignore both trivia and source location info. We could add a flag, though that wouldn't be sufficient to handle the character class trivia stripping which happens in `_dumpBase` in https://github.com/apple/swift-experimental-string-processing/pull/136. Really we ought to strip trivia from the entire tree prior to dumping (or if we strip enough, just compare with `==`), or otherwise have a parent type (like `PrettyPrinter`) that manages the dumping logic as it walks the tree and can store an `ignoreTrivia` flag. I can look into that in a follow-up.

hamishknight:
@swift-ci please test Linux

PR #134, opened 2022-01-27
DSLTree

milseman:
DSLTree is an internal representation shared between literals and result builders. 

- Lowering/transforming AST -> DSLTree
  - TODO: More atoms, built-in character classes, etc
- Pretty-printing DSLTree
  - At rough parity of the original PrintAsPattern
- Switch compiler over to DSLTree
- Switch result builders over to DSLTree
- Legacy VM support for everything except the new extensions
- CaptureStructure support


rxwei (approved):

Sources/_MatchingEngine/Regex/DSLTree.swift
```diff
@@ -0,0 +1,332 @@
...
+            result += str
+            astChildren.formIndex(after: &idx)
+          }
+          return result.isEmpty ? nil : result
+        }
+
+        // No need to nest single children concatenations
+        if astChildren.count == 1 {
+          return astChildren.first!.dslTreeNode
+        }
```
hamishknight, line 200: When do single-child concatenations occur? I kind of wonder whether `AST.Concatenation` should enforce that `children.count >= 2` (and maybe have a utility that produces concat-or-self), or is it better for compatibility with the DSL to allow them?

hamishknight (approved):

Sources/_MatchingEngine/Regex/DSLTree.swift
rxwei: I’m in favor of enforcing a plural count.

milseman: How would we enforce that in the DSL? Would we error out if given a single member? I worry there could be impact with sensible refactoring like quoting a literal run of characters vs listing each separately.

rxwei: The builder DSL always supports having a singular element in a block, e.g. `Regex { "a" }`. It's just that `buildBlock(_:)` (unary) never creates a concatenation in this scenario, but returns the argument instead.  As such, I'm not sure when else a singular concatenation would be created.

Sources/_MatchingEngine/Regex/DSLTree.swift
```diff
@@ -0,0 +1,332 @@
...
+    case trivia(String)
+
+    // TODO: Probably some atoms, built-ins, etc.
+
+    case empty
+
+    case stringLiteral(String)
+
+    /// An embedded literal
+    case regexLiteral(AST.Node)
```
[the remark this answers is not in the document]
hamishknight: Hmm yeah, I guess they would have to form their own scope in which they're only active in that portion of the regex

milseman:

milseman:
@hamishknight @rxwei I had to move some stuff around, can you take a look at this PR again to see if it's mergeable?

rxwei (approved):

PR #127, opened 2022-01-25
Switch back to Swift tuples for typed captures

rxwei:
This depends on #114 and #126. Please review 3db17b5.

PR #135, opened 2022-01-27
Heterogenous alternation

rxwei:
Introduce `oneOf` combinator, which takes a builder block and joins all elements into a regex alterantion. Fix the legacy VM to support alternation with captures.

-----

Example:

    oneOf {
      "a".capture()
      "b".capture()
      "c"
    } => `.Match = (Substring, Substring?, Substring?)`

### DIFF
--- a/Sources/_StringProcessing/RegexDSL/Core.swift
+++ b/Sources/_StringProcessing/RegexDSL/Core.swift
@@ -130,12 +130,6 @@ extension RegexProtocol {
     in inputRange: Range<String.Index>,
     mode: MatchMode = .wholeString
   ) -> RegexMatch<Match>? {
-    // Casts a Swift tuple to the custom `Tuple<n>`, assuming their memory
-    // layout is compatible.
-    func bitCastToMatch<T>(_ x: T) -> Match {
-      assert(MemoryLayout<T>.size == MemoryLayout<Match>.size)
-      return unsafeBitCast(x, to: Match.self)
-    }
     // TODO: Remove this branch when the matching engine supports captures.
     if regex.hasCapture {
       let vm = HareVM(program: regex.program.legacyLoweredProgram)
@@ -151,7 +145,7 @@ extension RegexProtocol {
         let typeErasedMatch = captures.matchValue(
           withWholeMatch: input[range]
         )
-        convertedMatch = _openExistential(typeErasedMatch, do: bitCastToMatch)
+        convertedMatch = typeErasedMatch as! Match
       }
       return RegexMatch(range: range, match: convertedMatch)
     }
--- a/Sources/VariadicsGenerator/VariadicsGenerator.swift
+++ b/Sources/VariadicsGenerator/VariadicsGenerator.swift
@@ -160,6 +160,20 @@ struct VariadicsGenerator: ParsableCommand {
       print(to: &standardError)
     }
 
+    print("Generating alternation overloads...", to: &standardError)
+    for (leftArity, rightArity) in Permutations(totalArity: maxArity) {
+      print(
+        "  Left arity: \(leftArity)  Right arity: \(rightArity)",
+        to: &standardError)
+      emitAlternation(leftArity: leftArity, rightArity: rightArity)
+    }
+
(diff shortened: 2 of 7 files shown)


Package.swift
```diff
@@ -42,14 +42,18 @@ let package = Package(
             name: "_StringProcessing",
             dependencies: ["_MatchingEngine"],
             swiftSettings: [
-                .unsafeFlags(["-enable-library-evolution"])
+                .unsafeFlags(["-enable-library-evolution"]),
+                .unsafeFlags(["-Xfrontend", "-enable-experimental-pairwise-build-block"])
```
milseman, line 46: I'm not sure why Github shows these diffs given that they're merged. Do you think you can rebase?

milseman (approved):
LGTM AFAICT, though I would like to see the reduced diff after a rebase to make sure.

Package.swift
rxwei: Yeah I'm in the process of rebasing. There's a lot of conflicts here.

Sources/_StringProcessing/RegexDSL/Core.swift
```diff
@@ -131,7 +125,7 @@ extension RegexProtocol {
         let typeErasedMatch = captures.matchValue(
           withWholeMatch: input[range]
         )
-        convertedMatch = _openExistential(typeErasedMatch, do: bitCastToMatch)
+        convertedMatch = typeErasedMatch as! Match
```
milseman, line 128: (out of curiosity) Is this from this change, or is this from the proper use of tuples?

Sources/_MatchingEngine/Regex/AST/AST.swift
```diff
@@ -153,7 +153,7 @@ extension AST {
     public init(_ mems: [AST.Node], pipes: [SourceLocation]) {
       // An alternation must have at least two branches (though the branches
       // may be empty AST nodes), and n - 1 pipes.
-      precondition(mems.count >= 2)
+      // precondition(mems.count >= 2)
```
milseman, line 156: What's the story here?

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
```diff
@@ -30,7 +30,6 @@ extension AST.Node {
     // Note: This implementation could be more optimized.
     switch self {
     case .alternation(let alternation):
-      assert(alternation.children.count > 1)
```
milseman, line 33: Does this trip something up?

Package.swift
milseman: Hmm, would it be a rebase or just a cherry-pick of 41ffdc7af3?

Sources/_MatchingEngine/Regex/AST/AST.swift
rxwei: This had to be uncommented before we had `DSLTree`, but now I can revert this part and make this be DSL-specific.

Sources/_StringProcessing/RegexDSL/Core.swift
rxwei: >  is this from the proper use of tuples?

Since we are no longer using nominal tuples, this makes it a safer than a bitcast.  It should have been done in #127 but I just came across this yesterday.

Sources/_MatchingEngine/Regex/Parse/CaptureStructure.swift
rxwei: Reverted.

Sources/_MatchingEngine/Regex/AST/AST.swift
rxwei: Done.

Package.swift
rxwei: Phew... done.

rxwei:
@swift-ci please test Linux

rxwei:
@swift-ci please test Linux

rxwei:
@swift-ci please test Linux

milseman (approved):
LGTM

natecook1000 (approved):
LGTM!

PR #136, opened 2022-01-28
Parse extended syntax (?x)

hamishknight:
This is based on top of #133, only the last 3 commits are relevant.

---

Parse extended syntax which can be enabled with the `(?x)` and `(?xx)` options. This enables non-semantic whitespace both inside and outside custom character classes, and allows end-of-line `# abc` comment syntax.

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/CustomCharClass.swift
+++ b/Sources/_MatchingEngine/Regex/AST/CustomCharClass.swift
@@ -42,6 +42,9 @@ extension AST {
       /// the contents should be interpreted literally.
       case quote(Quote)
 
+      /// Trivia such as non-semantic whitespace.
+      case trivia(Trivia)
+
       /// A binary operator applied to sets of members `abc&&def`
       case setOperation([Member], Located<SetOp>, [Member])
     }
@@ -81,11 +84,27 @@ extension CustomCC.Member {
     case .range(let r): return r
     case .atom(let a): return a
     case .quote(let q): return q
+    case .trivia(let t): return t
     case .setOperation(let lhs, let op, let rhs): return (lhs, op, rhs)
     }
   }
 
   func `as`<T>(_ t: T.Type = T.self) -> T? {
     _associatedValue as? T
+  }
+
+  public var isTrivia: Bool {
+    if case .trivia = self { return true }
+    return false
+  }
+}
+
+extension AST.CustomCharacterClass {
+  /// Strip trivia from the character class members. This does not recurse into
+  /// nested custom character classes.
+  public var strippingTriviaShallow: Self {
+    var copy = self
+    copy.members = copy.members.filter { !$0.isTrivia }
+    return copy
   }
 }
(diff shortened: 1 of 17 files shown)


hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

hamishknight:
@swift-ci please test Linux

Sources/_MatchingEngine/Regex/Printing/DumpAST.swift
```diff
@@ -282,7 +290,15 @@ extension AST.Quantification: _ASTPrintable {
 
 extension AST.CustomCharacterClass: _ASTNode {
   public var _dumpBase: String {
-    "customCharacterClass(\(members))"
+    // Exclude trivia for now, as we don't want it to appear when performing
+    // comparisons of dumped output in tests.
+    // TODO: We should eventually have some way of filtering out trivia for
+    // tests, so that it can appear in regular dumps.
```
milseman, line 296: Even more so, we may want a tree transform to strip trivia, to convert trivia into semantic whitespace, to convert semantic whitespace into trivia, etc. Not super critical to add these transforms, but I think it gives us some flexibility and demonstrates that we can add conversion features

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -123,7 +123,7 @@ extension AST.Atom {
 extension AST.CustomCharacterClass.Member {
   func generateConsumer(
     _ opts: MatchingOptions
-  ) throws -> Program<String>.ConsumeFunction {
+  ) throws -> Program<String>.ConsumeFunction? {
```
milseman, line 126: I'd rather strip trivia first and error (i.e. throw unreachable) if we see it. 

How about a `var strippingTrivia: Self` on `CustomCharacterClass`?

milseman (approved):
LGTM, but would like to at least get `CustomCharClass.strippingTrivia` to avoid the optional consumer

hamishknight:
@swift-ci please test Linux

Sources/_StringProcessing/ConsumerInterface.swift
hamishknight: How does this look @milseman?

Sources/_MatchingEngine/Regex/AST/CustomCharClass.swift
```diff
@@ -81,11 +84,25 @@ extension CustomCC.Member {
...
+    if case .trivia = self { return true }
+    return false
+  }
+}
+
+extension AST.CustomCharacterClass {
+  public var strippingTrivia: Self {
+    var copy = self
+    copy.members = copy.members.filter { !$0.isTrivia }
+    return copy
```
milseman, line 106: Wouldn't this need to recursively strip?

hamishknight: Not for the places where it's currently used, I've renamed it to `strippingTriviaShallow` for now to at least make it clear, and will look at adding a recursive version when looking at adding more general AST trivia stripping utilities

PR #137, opened 2022-01-31
Diagnose quantifiers without operands

hamishknight:
Avoid treating such cases as literal.


hamishknight:
@swift-ci please test Linux

PR #140, opened 2022-02-01
Regex Syntax Pitch

hamishknight:


Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+We aim to implement a syntactic superset of:
+
+- [PCRE 2][pcre2-syntax], an "industry standard" of sorts, and a rough superset of Perl, Python, etc.
+- [Oniguruma][oniguruma-syntax], an internationalization-oriented engine with some modern features
+- [ICU][icu-syntax], used by NSRegularExpression, a Unicode-focused engine.
+- [.NET][.net-syntax]'s regular expressions, which support delimiter-balancing and some interesting minor details on conditional patterns.
+- **TODO: List Java here? It doesn't really add any more syntax than the above other than `\p{javaLowerCase}`**
+
+We also intend to achieve at least Level 1 (**TODO: do we want to promise Level 2?**) [UTS#18][uts18] conformance, which specifies regular expression matching semantics without mandating any particular syntax. However we can infer syntactic feature sets from its guidance.
+
```
milseman, line 22:
```suggestion
**TODO(Michael): Rework and expand prose**
```

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+## Regex syntax supported
+
+### General syntax
+
+The following syntax are supported by all the above engines.
```
milseman, line 27: I think this note is fine, though we may want to start shifting how we frame or talk about this into what Swift will support. It's super helpful to note and show what other engines support/do, especially since the main value add of the pitched syntax is as a superset.

Something more like we're formally proposing to add the following to Swift, with other engine notes along the way.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Alternation
+
+```
+Regex       -> '' | Alternation
+Alternation -> Concatenation ('|' Concatenation)*
+```
+
+This is the operator with the lowest precedence in a regular expression, and checks if any of its branches match the input.
```
milseman, line 36: I'm really glad to see a grammar. Prose could be more of the form of briefly describing what it is/does. Something in-between the terseness of pcre2syntax and the fuller specification of pcre2pattern.

Maybe something like "`|` denotes what is formally called an alternation, or a choice between alternatives. Any number of alternatives may appear, including empty alternatives. This operator has the lowest precedence of all operators in a regex literal."

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+#### Quantification
+
+```
+Quantification -> QuantOperand Quantifier?
+Quantifier     -> ('*' | '+' | '?' | '{' Range '}') QuantKind?
+QuantKind      -> '?' | '+'
+```
+
+Specifies that the operand may be matched against a certain number of times.
+
```
milseman, line 56:
```suggestion
**TODO: Briefly mention each and what it means, noting that options can swap eager/reluctant. Might be a good time to introduce the eager/reluctant/possessive terminology**
```

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+            
+NamedGroup    -> 'P<' GroupNameBody '>'
+               | '<' GroupNameBody '>'
+               | "'" GroupNameBody "'"
+
+GroupNameBody -> Identifier
+```
+
+Groups define a new scope within which a recursive regular expression pattern may occur. Groups have different semantics depending on how they are introduced, some may capture the nested match, some may match against the input without advancing, some may change the matching options set in the new scope, etc.
+
```
milseman, line 72:
```suggestion
**TODO: Something like "note that there are other things that may syntactically appear similarly to groups, but are their own constructs. See .... in-line options, backreferences, ... **
```

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+#### Anchors
+
+```
+Anchor -> '^' | '$' | '\b'
+```
+
+Anchors match against a certain position in the input rather than on a particular character of the input.
```
milseman, line 79:
```suggestion
Anchors match against a certain position in the input rather than on a particular character of the input.

**TODO: We support a lot more anchors. Might be worth listing them with a very brief description of what they mean.**
```

 Need to balance this with brevity, of course. We can use disclosure triangles for anything that gets too in-depth, like options and nitty-gritty engine differences.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Callout syntax
+
+#### Absent functions
+
+### ICU-specific syntax
+
+
+
+### .NET-specific syntax
```
milseman, line 120: I think we want to organize around what we're adding to Swift rather than organize around other engines. The notes about other engines are useful for evaluating it and are the primary motivation, but for the description of the syntax and solution I think we should frame it around Swift foremost.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Balancing groups
+
+```
+GroupNameBody -> Identifier | Identifier? '-' Identifier
+```
+
+.NET supports the ability for a group to reference a prior group, causing the prior group to be deleted, and any intermediate matched input to become the capture of the current group.
```
milseman, line 128: This could be something like...

"Introduced by .NET, balancing groups ...."

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Character class subtraction with `-`
+
+
+
+## Syntactic differences between engines
+
+### Conflicting differences
```
milseman, line 136: Maybe just an intro paragraph and then everything listed is a potential conflict with our demonstrated resolution strategy. I.e. everything gets elevated one level of markdown-nesting .

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Character class set operations
+
+In a custom character class, some engines allow for binary set operations that take two character class inputs, and produce a new character class output. However which set operations are supported and the spellings used differ by engine.
+
+| PCRE | ICU | UTS#18 | Oniguruma | .NET | Java |
+|------|-----|--------|-----------|------|------|
+| ❌ | Intersection `&&`, Subtraction `--` | Intersection & Subtraction | Intersection `&&` | Subtraction via `-` | Intersection  `&&` |
```
milseman, line 144: Isn't UTS\#18 "Intersection, subtraction, and symmetric difference"?

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+
+[UTS#18][uts18] requires intersection and subtraction, and uses the operation spellings `&&` and `--` in its examples, though it doesn't mandate a particular spelling. In particular, conforming implementations could spell the subtraction `[[x]--[y]]` as `[[x]&&[^y]]`. UTS#18 also suggests a symmetric difference operator `~~`, and uses an explicit `||` operator in examples, though doesn't require either operations.
+
+These differences are conflicting, as engines that don't support a particular operator treat them as literal, e.g `[x&&y]` in PCRE is the character class of `["x", "&", "y"]` rather than an intersection.
+
+We intend to support the operators `&&`, `--`, `-`, and `~~`. This means that any regex literal containing these sequences in a custom character class while being written for an engine not supporting that operation will have a different semantic meaning in our engine. However this ought not to be a common occurrence, as specifying a character multiple times in a custom character class is redundant. However, we intend on providing a strict compatibility mode that may be used to emulate behavior of a particular engine (**TODO: all engines, or just PCRE?**).
+
```
milseman, line 151: I feel like the double-characters are no-brainers for us and should be their own paragraph. I think the single character `-` is a little more arguable, so I'd put it in its own paragraph and be up front with that fact.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,245 @@
...
+| PCRE | ICU | UTS#18 | Oniguruma | .NET | Java |
+|------|-----|--------|-----------|------|------|
+| ❌ | ✅ | 💡 | ✅ | **TODO** | ✅ |
+
+UTS#18 doesn't require this, though it does suggest it as a way to clarify precedence for chains of character class set operations e.g `[\w--\d&&\s]`, which the user could write as `[[\w--\d]&&\s]`.
+
+PCRE does not support this feature, and as such treats `]` as the closing character of the custom character class. Therefore `[[a]b[c]]` is interpreted as the character class `["[", "a"]`, followed by literal `b`, and then the character class `["c"]`, followed by literal `]`.
+
+We aim to support nested custom character classes, with a strict PCRE mode for emulating the PCRE behavior if desired.
```
milseman, line 164: Strictness being very much future work and not required for this to meet Swift's aims, AFAICT. We have a pragmatic choice between tradeoffs, and we have future extensibility to address these issues later. I'll work on intro prose and maybe future work prose to help out

Documentation/Evolution/RegexSyntax.md
hamishknight: The example they give makes it look like symmetric difference is a requirement, but the text says otherwise:

> To meet this requirement, an implementation shall supply mechanisms for union, intersection and set-difference of sets of characters within regular expression character class expressions.

So we don't have to support symmetric difference if we don't want to.

milseman: I feel like we might as well, so I'd say we have support. I'd avoid the use of a `&` if it's not syntax, as that could be confusing.

milseman:
This exhaustive treatment is so, so helpful. Thanks for continuing to push on it and I think it's coming along well.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,698 @@
...
+
+- [PCRE 2][pcre2-syntax], an "industry standard" of sorts, and a rough superset of Perl, Python, etc.
+- [Oniguruma][oniguruma-syntax], an internationalization-oriented engine with some modern features
+- [ICU][icu-syntax], used by NSRegularExpression, a Unicode-focused engine.
+- [.NET][.net-syntax]'s regular expressions, which support delimiter-balancing and some interesting minor details on conditional patterns.
+- **TODO: List Java here? It doesn't really add any more syntax than the above other than `\p{javaLowerCase}`**
+
+We also intend to achieve at least Level 1 (**TODO: do we want to promise Level 2?**) [UTS#18][uts18] conformance, which specifies regular expression matching semantics without mandating any particular syntax. However we can infer syntactic feature sets from its guidance.
+
+**TODO(Michael): Rework and expand prose**
```
milseman, line 23:
```suggestion
<!--
Hello, we want to issue an update to [Regular Expression Literals](https://forums.swift.org/t/pitch-regular-expression-literals/52820) and prepare for a formal proposal. The great delimiter delibration continues to unfold, so in the meantime, we have a significant amount of surface area to present for review/feedback: the syntax _inside_ a regex literal.
-->

# Regex Literal Interior Syntax

- Authors: Hamish Knight, Michael Ilseman

## Introduction

Regex literals declare a string processing algorithm using syntax familiar across a variety of languages and tools throughout programming history. Formalizing regex literals in Swift requires choosing a delimiter strategy (e.g. `#/.../#` or `re'...'`), detailing the syntax accepted in between the delimiters ("interior syntax"), and specifying actual types and any relevant protocols for the literal itself.

This proposal-component focuses on the interior syntax, which is large enough for its own targeted discussion ahead of the full proposal. Regex literal interior syntax will be part of Swift's source-compatibility story (and to some extent binary compatibility), so we present a detailed and comprehensive design.

## Motivation

Swift aims to be a pragmatic programming language, balancing (TODO: prose). Rather than pursue a novel interior syntax, (TODO: prose).

Regex interior syntax is part of a larger [proposal](https://forums.swift.org/t/pitch-regular-expression-literals/52820), which in turn is part of a larger [string processing effort](https://forums.swift.org/t/declarative-string-processing-overview/52459).

## Proposed Solution

We propose accepting a syntactic "superset" of the following existing regular expression engines:

- [PCRE 2][pcre2-syntax], an "industry standard" and a rough superset of Perl, Python, etc.
- [Oniguruma][oniguruma-syntax], a modern engine with additional features.
- [ICU][icu-syntax], used by NSRegularExpression, a Unicode-focused engine.
- [.NET][.net-syntax], which adds delimiter-balancing and some interesting minor details around conditional patterns.

To our knowledge, all other popular regex engines support a subset of the above syntaxes.

We also support [UTS#18][uts18]'s full set of character class operators (to our knowledge no other engine does). Beyond that, UTS#18 deals with semantics rather than syntax, and what syntax it uses is covered by the above list. We also parse `\p{javaLowerCase}`, meaning we support a superset of Java 8 as well.

Note that there are minor syntactic incompatibilities and ambiguities involved in this approach. Each is addressed in the relevant sections below

```

milseman (commented):
Intro, motivation, and proposed solution.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,871 @@
...
+
+### Nested custom character classes
+
+This allows e.g `[[a]b[c]]`, which is interpreted the same as `[abc]`. It also allows for more complex set operations with custom character classes as the operands.
+
+| PCRE | ICU | UTS#18 | Oniguruma | .NET | Java |
+|------|-----|--------|-----------|------|------|
+| ❌ | ✅ | 💡 | ✅ | ❌ | ✅ |
+
+UTS#18 doesn't require this, though it does suggest it as a way to clarify precedence for chains of character class set operations e.g `[\w--\d&&\s]`, which the user could write as `[[\w--\d]&&\s]`.
```
milseman, line 653: Extra newline for some markdown renders which otherwise treat the next paragraph is another row in the table.

```suggestion


UTS#18 doesn't require this, though it does suggest it as a way to clarify precedence for chains of character class set operations e.g `[\w--\d&&\s]`, which the user could write as `[[\w--\d]&&\s]`.
```

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,871 @@
...
+- `(?~absent)`: Absent repeater, which matches against any input not matched by `absent`. Equivalent to `(?~|absent|\O*)`.
+- `(?~|absent)`: Absent stopper, which limits any subsequent matching to not include `absent`.
+- `(?~|)`: Absent clearer, which undoes the effects of the absent stopper.
+
+## Syntactic differences between engines
+
+The above regular expression grammar covers a superset of the syntax accepted by PCRE, ICU, Oniguruma, .NET, and Java. However there are cases where the same syntax is parsed differently by these engines. This section provides a summary of these differences, and specifies the interpretation that will be made by the Swift regex parser.
+
+These differences inherently mean that our default parser behavior cannot be fully compatible with these other engines. However, this would not preclude the potential future implementation of different compatibility modes for different engines in which we support their parsing behavior of certain syntax.
+
```
milseman, line 626: Huh, I thought this was marked for me to fill out, so I did. But now I don't see the notes. Here's what I wrote in case it's helpful, not that you have to take it.

> The proposed "syntactic superset" introduces some minor ambiguities, as each engine supports a slightly different set of features. When a particular engine's parser sees a feature it doesn't support, it typically has a fall-back behavior, such as treating the unknown feature as literal contents.
>
>Explicit compatibility modes, i.e. precisely mimicing emergent behavior from a specific engine's parser, is deferred as future work from this proposal. Conversion from this "syntactic superset" to a particular engine's syntax (e.g. as an AST "pretty printer") is deferred as future work from this proposal.
>
>Below is an exhaustive treatment of every syntactic ambiguity we have encountered.

hamishknight: I kind of prefer this version, updated to take it

milseman:
(I want to share my progress quickly and won't be able to formally commit to this PR yet, so I apologize for using the comment stream for this)

Reworking the start of Detailed Design a bit, trying to crunch it down a little bit. With this we can delete the Trivia and Comment sections too.

## Detailed Design

We propose the following syntax for use inside Swift regex literals.

*TODO:* Disclosure triangle explaining the grammar conventions?

### Top-level regular expression

```
Regex     -> GlobalMatchingOptionSequence? RegexNode
RegexNode -> '' | Alternation
Alternation -> Concatenation ('|' Concatenation)*
Concatenation   -> (!'|' !')' ConcatComponent)*
```

A regex literal may be prefixed with a sequence of global matching options(*TODO*: intra-doc link). A literal's contents can be empty or a sequence of alternatives separated by `|`.

Alternatives are a series of expressions concatenated together. The concatentation ends with either a `|` denoting the end of the alternative or a `)` denoting the end of a recursively parsed group.

Alternation has a lower precedence than concatenation or other operations, so e.g `abc|def` matches against `abc` or `def`..

### Concatenated subexpressions

```
ConcatComponent -> Trivia | Quote | Quantification

Trivia  -> Comment | NonSemanticWhitespace
Comment -> '(?#' (!')')* ')' | EndOfLineComment

(extended syntax only) EndOfLineComment      -> '#' .*$
(extended syntax only) NonSemanticWhitespace -> \s+

Quote -> '\Q' (!'\E' .)* '\E'

```

Each component of a concatenation may be "trivia" (comments and non-semantic whitespace, if applicable), a quoted run of literal content, or a potentially-quantified subexpression.

In-line comments, similarly to C, are lexical and are not recursively nested like normal groups are. A closing `)` cannot be escaped. Quotes are similarly lexical, non-nested, and the `\` before a `\E` cannot be escaped.

For example, `\Q^[xy]+$\E`, is treated as the literal characters `^[xy]+$` rather than an anchored quantified character class. `\Q\\E` is a literal `\`. 

*TODO*: Do we want to warn for a `\E` appearing without a preceding `\Q`? Do we want to warn for a `)` without anything opening it?

### Quantified subexpressions

(... rest of document)

milseman:
I pushed a commit that trimmed prose a little bit and reorganized some of the detailed design.

milseman:
I finished the intro, motivation, and proposed solution sections. I added 3 alternatives considered. 

@hamishknight how close are we to having this ready? I made a `TODO: Hamish` not in the Swift canonical syntax part, basically it's not formally proposed by useful to discuss a preferred spelling, so I don't think we need anything as heavily structured as the normal proposal parts.

milseman (commented):
>  We also parse `\p{javaLowerCase}`, meaning we support a superset of Java 8 as well.

I guess we'd say "we also parse Java's lowercase, uppercase, and whitespace properties, meaning we support a superset of Java 8 as well".

milseman (commented):
Ugh, had this in-progress review that got jumbled up. Sorry about that.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -298,13 +314,13 @@ Identifier -> [\w--\d] \w*
 
 Groups define a new scope that contains a recursively nested regex. Groups have different semantics depending on how they are introduced.
 
-Note there are additional constructs that may syntactically appear similar to groups, but are distinct. See the *group-like atoms* section.
+Note there are additional constructs that may syntactically appear similar to groups, such as backreferences and conditionals, but are distinct.
```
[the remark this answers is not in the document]
hamishknight: Oops, definitely didn't mean to put conditionals here, not sure how that happened

Documentation/Evolution/RegexSyntax.md
```diff
@@ -257,12 +271,14 @@ Other Unicode properties however must specify both a key and value.
...
-- The special properties `any`, `assigned`, `ascii`.
+- The UTS#18 special properties `any`, `assigned`, `ascii`.
 - The POSIX compatibility properties `alnum`, `blank`, `graph`, `print`, `word`, `xdigit`. The remaining POSIX properties are already covered by boolean Unicode property spellings.
+- The special PCRE2 properties `Xan`, `Xps`, `Xsp`, `Xuc`, `Xwd`.
+- The special Java properties `javaLowerCase`, `javaUpperCase`, `javaWhitespace`, `javaMirrored`.
 
 Note that the internal `PropertyContents` syntax is shared by both the `\p{...}` and POSIX-style `[:...:]` syntax, allowing e.g `[:script=Latin:]` as well as `\p{alnum}`.
 
-#### `\K`
+### `\K`
```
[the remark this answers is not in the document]
hamishknight: Oops, yes this should be reverted

Documentation/Evolution/RegexSyntax.md
```diff
@@ -427,7 +443,7 @@ We support all the matching options accepted by PCRE, ICU, and Oniguruma. In add
 - `n`: Disables the capturing behavior of `(...)` groups. Named capture groups must be used instead.
 - `s`: Changes `.` to match any character, including newlines.
 - `U`: Changes quantifiers to be reluctant by default, with the `?` specifier changing to mean greedy.
-- `x`, `xx`: Enables extended syntax mode, which allows non-semantic whitespace and end-of-line comments. See the *trivia* section for more info.
+- `x`, `xx`: Enables extended syntax mode, which allows non-semantic whitespace and end-of-line comments. See [Extended Syntax Modes](#extended-syntax-modes) for more info.
```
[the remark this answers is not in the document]
hamishknight: Updated the extended syntax section to also include a mention of Perl, is that sufficient? Or do we want to go into more detail?

Documentation/Evolution/RegexSyntax.md
```diff
@@ -619,12 +619,12 @@ OnigurumaCalloutArgList -> OnigurumaCalloutArg (',' OnigurumaCalloutArgList)*
...
+OnigurumaCalloutContents   -> <String> | '{' OnigurumaCalloutContents '}'
 OnigurumaCalloutDirection  -> 'X' | '<' | '>'
 ```
 
-A callout is a feature that allows a user-supplied function to be called when matching reaches that point in the pattern. We supported parsing both the PCRE and Oniguruma callout syntax. The PCRE syntax accepts a string or numeric argument that is passed to the function. The Oniguruma syntax is more involved, and may accept a tag, argument list, or even an arbitrary program in the 'callout of contents' syntax.
+A callout is a feature that allows a user-supplied function to be called when matching reaches that point in the pattern. We supported parsing both the PCRE and Oniguruma callout syntax. The PCRE syntax accepts a string or numeric argument that is passed to the function. The Oniguruma syntax is more involved, and may accept an identifier with an optional tag and argument list. It may also accept an arbitrary program in the 'callout of contents' syntax. This is an expanded version of Perl's interpolation syntax, and allows an arbitrary nesting of delimiters in addition to an optional tag and direction.
```
milseman, line 627: Worth an additional paragraph perhaps:

> While we propose parsing these for the purposes of issuing helpful diagnostics, we are deferring full support for interpolation for the future.

We may want to remove the "Oniguruma" from the title of some of the productions, as this may end up being the Swift interpolation syntax. We just need to figure out what types of things go inside there (analogous to string interpolations) and if that's protocol driven. Clearly future work, though it would bring literals more up to parity with result builders' ability to interweave library calls.

hamishknight: How does this look?

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,848 @@
...
+#### Unicode scalars
+
+```
+UnicodeScalar -> '\u{' HexDigit{1...} '}'
+               | '\u'  HexDigit{4}
+               | '\x{' HexDigit{1...} '}'
+               | '\x'  HexDigit{0...2}
+               | '\U'  HexDigit{8}
+               | '\o{' OctalDigit{1...} '}'
+               | '\0' OctalDigit{0...3}
```
natecook1000, line 220: In earlier discussions, we had talked about only supporting the `'\u{' HexDigit{1...} '}'` style for unicode scalar literals, to match strings, and offering fixits for all the others. Are we planning to support all these varied styles instead?

natecook1000 (commented):
Awesome! 💯

My only general comment is that the last paragraph in the proposal should come very early on, so that this syntactical/parsing proposal isn't misinterpreted as promising the feature set of Swift regular expressions.

Documentation/Evolution/RegexSyntax.md
hamishknight: I believe we want to support at least parsing them, but we will likely want to warn and tell users to use `\u{...}` instead

milseman: "Support" should be maximally permissive (balanced with being clear and  unambiguous) as that's the main value of going with existing syntax instead of a novel syntax. But, we can also negotiate what is warned on and what has fixits.

I think the "Swift canonical" scalar syntax is clearly superior and should be encouraged, but there's a choice of mechanism here. We certainly should accept the other forms and compile them to the same thing in the end, it's a question of whether and what kind of diagnostics we emit.

I am a little concerned with tons of warnings, especially since Swift doesn't support fine-grained warning control (err, it does internally because I added it way back when when I added global warning suppression and warnings-as-errors, but I don't think it's been exposed). A bad developer experience is warnings that are unavoidable or for whom addressing is worse than living with them.

For a function taking a run-time string containing regex syntax, we'd support all the ways to spell it rather than throw a cannot-compile error. A different API would be whether it's in Swift canonical form or to put it into Swift canonical form. If we warn too heavily, that could even encourage people to use raw string literals instead of regex literals, which would be awkward.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,845 @@
...
+Backreference -> '\g{' NamedOrNumberRef '}'
+               | '\g' NumberRef
+               | '\k<' NamedOrNumberRef '>'
+               | "\k'" NamedOrNumberRef "'"
+               | '\k{' NamedRef '}'
+               | '\' [1-9] [0-9]+
+               | '(?P=' NamedRef ')'
+```
+
+A backreference evaluates to the value last captured by the referenced capturing group. If the referenced capture has not been evaluated yet, the match fails.
```
rxwei, line 495: Should we say something like "if a referenced capture does not exist anywhere in the regular expression, the regular expression is not valid"?

hamishknight: How's this?

Documentation/Evolution/RegexSyntax.md
natecook1000: FWIW, I disagree about accepting the other forms. This list is a superset; AFAIK there's no language that supports the full laundry list. That means that whenever you're moving regexes from language to language there's some necessary massaging of things like this. If we can make that as easy as clicking a fix-it or pressing ⌃⌥⌘F, that's more support than you typically get.

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,845 @@
...
+#### Anchors
+
+```
+Anchor -> '^' | '$' | '\A' | '\b' | '\B' | '\G' | '\y' | '\Y' | '\z' | '\Z'
+```
+
+Anchors match against a certain position in the input rather than on a particular character of the input.
+
+- `^`: Matches at the start of a line.
+- `$`: Matches at the end of a line.
```
natecook1000, line 159:
```suggestion
- `^`: Matches at the start of the input string, or the start of a line when in multi-line.
- `$`: Matches at the end of a line.
```

hamishknight:
@swift-ci please test Linux

Documentation/Evolution/RegexSyntax.md
```diff
@@ -0,0 +1,845 @@
...
+We propose the following syntax for regex.
+
+<details><summary>Grammar Notation</summary>
+
+For the grammar sections, we use a modified PEG-like notation, in which the grammar also describes an unambiguous top-down parsing algorithm.
+
+- `<Element> -> <Definition>` gives the definition of `Element`
+- The `|` operator specifies a choice of alternatives
+- `'x'` is the literal character `x`, otherwise it's a reference to x
+      + A literal `'` is spelled `"'"`
```
benrimmington, line 52:
```suggestion
    - A literal `'` is spelled `"'"`
```

PR #142, opened 2022-02-02
Cleanups and refactorings

milseman:
- Split off matching code/logic/types from `DSL/Core.swift`
- Make a `Util/Errors.swift` for our `Unsupported` and `Unreachable` errors, and use them more broadly
- Split off AST conversion logic from `DSLTree.swift`

PR #143, opened 2022-02-03
Move `MatchingEngine` out of `_MatchingEngine` 😐

milseman:
Currently based on top of https://github.com/apple/swift-experimental-string-processing/pull/142


milseman:
@rxwei Do you see any impact of this on the compiler integration side?

rxwei:
Is it just moving from _ME to _SP?  It should be fine as far as compiler integration is concerned.

Sources/_StringProcessing/Engine/StringProcessor.swift
```diff
@@ -0,0 +1,31 @@
...
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public // For prototypes
```
rxwei, line 12: Whatever's public in `_StringProcessing` is public to the user. Why do we want to expose `Program` or `MEProgram`?

milseman: It's because we're restricted to only 2 modules here. The engine should be its own module, internal to everything else, so that it can have API that's usable by prototypes and `_StringProcessing`. But since we have to slurp everything up into `_StringProcessing`, then this needs to service the prototypes too.

We can drop the prototypes support at some point, at which point this would be dropped. We have loads of public entities that we know we aren't going to ship as public already, so there will be an audit / purge. I can denote it with a FIXME comment or something to make it easier to find in the future.

rxwei: Would something like `@_spi(PrototypeSupport)` work?

milseman:

milseman:

PR #145, opened 2022-02-03
Move the _Unicode module into _StringProcessing

Azoy:
This removes the _Unicode target and instead moves those files into the _StringProcessing target as a sub folder.


Utils/gen-unicode-data/.gitignore
```diff
@@ -0,0 +1,8 @@
+.DS_Store
+/.build
+/Packages
+/*.xcodeproj
+xcuserdata/
+DerivedData/
+.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+Package.resolved
```
milseman, line 8: What's up with this? Why doesn't it just inherit from the package parent folder?

Utils/gen-unicode-data/Package.swift
```diff
@@ -0,0 +1,26 @@
...
+    .executableTarget(
+      name: "GenNormalization",
+      dependencies: ["GenUtils"]
+    ),
+    .executableTarget(
+      name: "GenScalarProps",
+      dependencies: ["GenUtils"]
+    )
+  ]
+)
```
milseman, line 26: Why is this a nested package?

milseman (commented):
Overall +1, some minor questions.

@rxwei , do you see any problems with us adding these data files as far as integrating with the compiler is concerned?

rxwei:
> @rxwei , do you see any problems with us adding these data files as far as integrating with the compiler is concerned?

Are the generated files Swift files?  Are these generated manually or by the build system?

Utils/gen-unicode-data/Package.swift
Azoy: It can be part of the string processing package if you feel it should be, but it's a pretty stand alone package that is very severable.

Azoy:
@milseman are you referring to the _Unicode source files being added to _StringProcessing, or the future c files for Unicode data?

Azoy:
@swift-ci please test

Azoy:
@swift-ci please test Linux

PR #146, opened 2022-02-03
Add Grapheme Breaking and Normalization

Azoy:
This adds the Unicode data needed and the stdlib implementations of grapheme breaking and normalization for use in _StringProcessing.


Package.swift
```diff
@@ -35,9 +35,12 @@ let package = Package(
         .testTarget(
             name: "MatchingEngineTests",
             dependencies: ["_MatchingEngine"]),
+        .target(
+            name: "_CUnicode",
+            dependencies: []),
         .target(
             name: "_StringProcessing",
-            dependencies: ["_MatchingEngine"],
+            dependencies: ["_MatchingEngine", "_CUnicode"],
```
rxwei, line 43: Hmm I don't think we can add arbitrary dependencies for `_MatchingEngine` or `_StringProcessing` until we 1) set up cmake on the compiler side to produce the _CUnicode module and 2) coordinate with @edymtt for internal builds.

Sources/_StringProcessing/Unicode/Decoding.swift
```diff
@@ -33,13 +33,13 @@
 
 enum UnsafeAssumingValidUTF8 {
   @inlinable @inline(__always)
-  public func decode(_ x: UInt8) -> Unicode.Scalar {
+  public static func decode(_ x: UInt8) -> Unicode.Scalar {
```
milseman, line 36: Part of moving these into the `_StringProcessing` module should be dropping all `public` keywords, since `_StringProcessing` will be supplying API as well.

Package.swift
milseman: Is there any way to avoid `_CUnicode`?

rxwei: Maybe you can build the C files as part of _StringProcessing via CMake? But I’m not sure how you can do that using SwiftPM.

stephentyrone: I don't see any reason why the two build systems _have_ to work the same way. If we need to include these directly in the CMake build for now, let's do that.

Azoy:
@swift-ci please test Linux

Issue #116, opened 2022-01-18
AST post-processing


hamishknight:
This issue tracks logic that needs to be implemented after the parser has produced an AST.

To be implemented:
- Group reference validity checking (as group references may refer to groups that come after them)
  - This will require deciding how to handle the ambiguity with the syntax  `(?(xxx))`. PCRE always treats this as a named reference. .NET only treats it as a named reference if there is a group defined with that name, otherwise it treats it as an arbitrary regex condition. It's possible we may want to require users explicitly spell named references `(?('xxx'))` to avoid this ambiguity.
- Errors for AST nodes that are unsupported by the matching engine
- Warnings for syntax that has a better spelling

milseman:
Is there global data we might also want to store?

hamishknight:
@milseman Did you have something specific in mind? Not sure if this is what you mean, but we will need to decide where to store the global matching options that PCRE allows to be parsed at the start of the regex, e.g `(*CR)` & `(*UTF)`. I have a patch that parses them, and for now stores them as an AST node that we'll produce if they're there. But it's possible we may want to have the parser produce an "AST root" type that wraps an AST along with the global options. I don't know if you've had any thoughts on that?

milseman:
It may make sense to have `struct AST` and `enum AST.Node`. So we have a struct that has an enum that stores structs...

hamishknight:
@milseman Makes sense, presumably recursive AST nodes would then store `AST.Node` instead of `AST`?

hamishknight:
This is mostly handled by https://github.com/apple/swift-experimental-string-processing/pull/379, I've split off https://github.com/apple/swift-experimental-string-processing/issues/380 for the parser warnings.

milseman:
@hamishknight should this be closed then or is it tracking something?

hamishknight:
I was going to close it with the merging of https://github.com/apple/swift-experimental-string-processing/pull/379, but am also happy to close now

milseman:
Oh right, forgot Github had that feature. Neat.

PR #148, opened 2022-02-08
Add DSL support for backreferences.

rxwei:
Allow `capture` and `tryCapture` to assign the captured value to a `Reference`, which can be used as a regex later in the scope.

`Reference` initialization creates a unique identifier. The compiler converts the identifier to an absolute backreference offset.

-----

Example:

```swift
let regex = Regex {
  let a = Reference()
  let b = Reference()
  capture("abc", as: a)
  capture("def", as: b)
  a
  capture(b)
}
```

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -31,6 +31,9 @@ extension Compiler.ByteCodeGen {
     case let .backreference(ref):
       try emitBackreference(ref)
 
+    case let .symbolicReference(id):
+      builder.buildUnresolvedReference(id: id)
+
     case let .unconverted(astAtom):
       if let consumer = try astAtom.generateConsumer(options) {
         builder.buildConsume(by: consumer)
@@ -293,8 +296,14 @@ extension Compiler.ByteCodeGen {
   }
 
   mutating func emitGroup(
-    _ kind: AST.Group.Kind, _ child: DSLTree.Node
+    _ kind: AST.Group.Kind,
+    _ child: DSLTree.Node,
+    _ referenceID: Reference.ID?
   ) throws -> CaptureRegister? {
+    guard kind.isCapturing || referenceID == nil else {
+      throw Unreachable("Reference ID shouldn't exist for non-capturing groups")
+    }
+
     options.beginScope()
     defer { options.endScope() }
 
@@ -303,7 +312,7 @@ extension Compiler.ByteCodeGen {
     //
     // FIXME: Unify with .groupTransform
     if kind.isCapturing, case let .matcher(_, m) = child {
-      let cap = builder.makeCapture()
+      let cap = builder.makeCapture(id: referenceID)
       emitMatcher(m, into: cap)
       return cap
     }
@@ -319,7 +328,7 @@ extension Compiler.ByteCodeGen {
       throw Unreachable("TODO: reason")
 
     case .capture, .namedCapture:
-      let cap = builder.makeCapture()
+      let cap = builder.makeCapture(id: referenceID)
       builder.buildBeginCapture(cap)
       try emitNode(child)
       builder.buildEndCapture(cap)
@@ -552,8 +561,8 @@ extension Compiler.ByteCodeGen {
         try emitConcatenationComponent(child)
       }
 
-    case let .group(kind, child):
-      _ = try emitGroup(kind, child)
+    case let .group(kind, child, referenceID):
+      _ = try emitGroup(kind, child, referenceID)
 
     case .conditional:
       throw Unsupported("Conditionals")
@@ -592,8 +601,8 @@ extension Compiler.ByteCodeGen {
     case let .convertedRegexLiteral(n, _):
       try emitNode(n)
 
-    case let .groupTransform(kind, child, t):
-      guard let cap = try emitGroup(kind, child) else {
+    case let .groupTransform(kind, child, t, referenceID):
+      guard let cap = try emitGroup(kind, child, referenceID) else {
         assertionFailure("""
           What does it mean to not have a capture to transform?
           """)
--- a/Sources/_StringProcessing/Engine/MEBuilder.swift
+++ b/Sources/_StringProcessing/Engine/MEBuilder.swift
@@ -42,6 +42,14 @@ extension MEProgram where Input.Element: Hashable {
     // as we compile?
     var captureStructure: CaptureStructure = .empty
 
+    // Symbolic reference resolution
+    var unresolvedReferences: [Reference.ID: [InstructionAddress]] = [:]
+    var captureOffsets: [Reference.ID: Int] = [:]
+    var captureCount: Int {
+      // We currently deduce the capture count from the capture register number.
+      nextCaptureRegister.rawValue
+    }
+
     public init() {}
   }
 }
@@ -65,6 +73,10 @@ extension MEProgram.Builder {
 
   public init<S: Sequence>(staticElements: S) where S.Element == Input.Element {
     staticElements.forEach { elements.store($0) }
+  }
+
+  var lastInstructionAddress: InstructionAddress {
+    .init(instructions.endIndex - 1)
   }
 
   public mutating func buildNop(_ r: StringRegister? = nil) {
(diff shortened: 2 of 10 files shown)


tevelee:
Idea for an alternative syntax:
```swift
let regex = Regex {
  let a = Reference<String>()
  let b = Reference<String>()
  capture("abc", into: a)
  capture("def", into: b)
  a
  capture(b)
}
```

rxwei:
@tevelee Good point.  We are going with using `let` statements.

rxwei:
@swift-ci please test Linux

rxwei:
@swift-ci please test Linux

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -35,6 +37,12 @@ extension Compiler.ByteCodeGen {
     case let .backreference(ref):
       try emitBackreference(ref)
 
+    case let .symbolicReference(id):
+      guard let offset = referencedCaptureOffsets[id] else {
+        throw RegexCompilationError.uncapturedReference
```
natecook1000, line 42: I don't think this should be an error — a valid regex can include a back reference before its been matched. 

Consider `(?:\w\1|:(\w):)+` — the first time through the outer group, `\1` doesn't have a value, so the first branch of the alternation can't match. Once the second branch matches something, however, `\1` _does_ have a value, and participates in matching in the way you would expect. So that pattern matches `:a:bacada`, but not `bacada` on its own. (https://regex101.com/r/mMxqcc/1)

rxwei: Updated.

rxwei: Update to support post-hoc captures. However, I think we should still emit an error when a reference is not captured anywhere in the regex, as it seems consistent with the error here https://regex101.com/r/vvXsgy/1.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -266,9 +273,41 @@ extension Compiler.ByteCodeGen {
...
+      }
+      for use in uses {
+        builder.instructions[use.rawValue] =
+          Instruction(.backreference, .init(capture: .init(offset)))
+      }
+    }
+  }
+
+  mutating func emitUnresolvedReference(id: Reference.ID) {
```
rxwei, line 288: Not sure if this is future-proof, as this approach assumes we never insert to the middle of the instruction list.

rxwei:
@swift-ci please test Linux@swift-ci please test Linux

Sources/_StringProcessing/ByteCodeGen.swift
milseman: @hamishknight how does the parser handle this? Maybe it's the same kind of strategy.

rxwei:
@swift-ci please test Linux

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -31,6 +35,9 @@ extension Compiler.ByteCodeGen {
     case let .backreference(ref):
       try emitBackreference(ref)
 
+    case let .symbolicReference(id):
+      emitUnresolvedReference(id: id)
```
milseman, line 39: I wonder if the builder should be tracking this and resolve it, like it does for labels, captures, etc.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -292,9 +299,41 @@ extension Compiler.ByteCodeGen {
...
+
   mutating func emitGroup(
-    _ kind: AST.Group.Kind, _ child: DSLTree.Node
+    _ kind: AST.Group.Kind,
+    _ child: DSLTree.Node,
+    _ referenceID: Reference.ID?
   ) throws -> CaptureRegister? {
+    guard kind.isCapturing || referenceID == nil else {
+      throw Unreachable("Reference ID shouldn't exist for non-capturing groups")
+    }
```
milseman, line 335: I'm also very strongly considering splitting off the capturing concept into its own DSLTree node. But that's a future PR.

milseman (approved):
LGTM, though I would prefer if things like reference resolution was handled by ME's builder.

Sources/_StringProcessing/ByteCodeGen.swift
rxwei: Good call. Done.

Sources/_StringProcessing/ByteCodeGen.swift
rxwei: Yes, I agree. The current construction is a bit weird. Perhaps something like `case capture(DSLTree.Node, CaptureTransform?, ReferenceID?)`

rxwei:
@swift-ci please test Linux

rxwei:
PTAL 58c6428. This adds support for subscripting a match result on a `Reference` to retrieve a capture. This is akin to named captures.

```swift
let a = Reference()
let result = regex.match {
  tryCapture(as: a) {
    oneOrMore(.digit)
  } transform: { Int($0) }
}
result[a] // => Any
result[a, as: Int.self] // => Int
```

rxwei:
@swift-ci please test Linux

Sources/_StringProcessing/ByteCodeGen.swift
hamishknight: The parser doesn't currently diagnose this at all, it'll be a part of the post-processing pass over the AST which we have yet to implement (https://github.com/apple/swift-experimental-string-processing/issues/116). I think we'll want the same diagnostic behavior as proposed here, i.e allow forward references, but error on references to captures that don't exist anywhere in the pattern.

milseman:
> To simplify the call site, I chose to precondition instead of returning optionals.

This is in direct contradiction to stdlib style for non-indexing subscripts. Is there any chance you can split the more controversial part into its own PR?

rxwei:
@swift-ci please test Linux

rxwei:
> This is in direct contradiction to stdlib style for non-indexing subscripts. Is there any chance you can split the more controversial part into its own PR?

This is no longer the case as I've switched to typed references. PTAL.

Sources/_StringProcessing/Capture.swift
```diff
@@ -31,6 +31,19 @@ extension Capture {
...
+extension Capture {
+  subscript(_ index: Int) -> Capture {
+    if case .tuple(let elements) = self {
+      assert(
+        (elements.startIndex..<elements.endIndex).contains(index),
+        "Capture index out of bounds"
+      )
+      return elements[index]
+    }
+    return self
```
milseman: Should this be a fatal error or internal invariant check that `index` is `0`?

Sources/_StringProcessing/Engine/MEBuilder.swift
```diff
@@ -274,7 +274,7 @@ extension MEProgram.Builder {
       .init(.backreference, .init(capture: cap)))
   }
 
-  public mutating func buildUnresolvedReference(id: Reference.ID) {
+  public mutating func buildUnresolvedReference(id: ReferenceID) {
```
milseman: What's the difference between a `ReferenceID` and a `Reference.ID`?

rxwei:
@swift-ci please test Linux

rxwei:
Nonetheless, I don't want to delay longer to merge backreferences, so I will be merging my first commit to add a non-generic `Reference` without subscript.

Tests/RegexTests/RegexDSLTests.swift
```diff
@@ -556,6 +556,55 @@ class RegexDSLTests: XCTestCase {
...
     }
   }
+
+  func testBackreference() throws {
+    try _testDSLCaptures(
+      ("abc#41#42abc#42#42", ("abc#41#42abc#42#42", "abc", 42, "#42", nil)),
+      matchType: (Substring, Substring, Int?, Substring?, Substring?).self, ==)
+    {
+      let a = Reference()
+      let b = Reference()
```
milseman, line 566: What happened to the type?

milseman (commented):
My reviews keep getting clobbered by the force pushes and I don't see the separate commit anymore. 

I saw types and now I don't. What's the plan here?

rxwei:
https://github.com/apple/swift-experimental-string-processing/pull/148#issuecomment-1048288748
I reverted my second commit as the first one (basic backreference functionality) was approved. I will open a new PR to add string types and subscript, to avoid further delays on the basic functionality.

PR #149, opened 2022-02-09
Replace DSL component nominal types with `Regex`.

rxwei:
Eliminates nominal types, e.g. `Concatenation_0_1`, returned by DSL builder methods and replace them with `Regex`. This will make sure we can ship a clean ABI even without variadic generics.


rxwei:
@swift-ci please test Linux

natecook1000 (approved):
Looks great! 🎉

PR #150, opened 2022-02-09
Add support for quantification kinds to the DSL

natecook1000:
This adds support for specifying eager / reluctant / possessive quantification from within the DSL, like the literal quantification operators with an appended `?`.

For example, with the default eager capturing, the following simple regex matches the entire string and capture the `"2"`:

```swift
let regex = Regex {
    oneOrMore(.word, .reluctant)
    CharacterClass.digit.capture()
}
"aaa1bbb2".match(regex).match // ("aaa1", "1")
```


Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -360,21 +352,23 @@ struct VariadicsGenerator: ParsableCommand {
     let matchType = arity == 0 ? baseMatchTypeName : "(\(baseMatchTypeName), \(quantifiedCaptures))"
     output("""
       \(arity == 0 ? "@_disfavoredOverload" : "")
-      public func \(kind.rawValue)<\(genericParameters(withConstraints: true))>(
-        _ component: Component
+      public func \(kind.rawValue)<\(genericParams)>(
+        _ component: Component,
+        _ kind: QuantificationKind = .eager
```
rxwei, line 357: Should `kind` come before `component`? This would make it consistent with the builder API. For the builder API we are kinda out of options since we'd have to support a trailing closure.

---

So it's

```swift
oneOrMore("x", .eager)
oneOrMore(.eager) { "x" }
```

vs.

```swift
oneOrMore(.eager, "x")
oneOrMore(.eager) { "x" }
```

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -62,6 +62,44 @@ extension CharacterClass: RegexProtocol {
 
 // Note: Quantifiers are currently gyb'd.
 
+/// Specifies how much to attempt to match when using a quantifier.
+public struct QuantificationKind {
```
rxwei, line 66: - "Kind" feels a bit too general. Kind could also refer to one-or-more/zero-or-more/zero-or-one. Is there a more specific word for this? Maybe `QuantificationBehavior`?
- For the DSL, should quantifier kinds read like an adverb for fluency?
  ```swift
  oneOrMore("x", .eagerly)
  oneOrMore("x", .reluctantly)
  ```

rxwei (approved):
Nice!

Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -360,21 +352,23 @@ struct VariadicsGenerator: ParsableCommand {
     let matchType = arity == 0 ? baseMatchTypeName : "(\(baseMatchTypeName), \(quantifiedCaptures))"
     output("""
       \(arity == 0 ? "@_disfavoredOverload" : "")
-      public func \(kind.rawValue)<\(genericParameters(withConstraints: true))>(
-        _ component: Component
+      public func \(kind.rawValue)<\(genericParams)>(
+        _ component: Component,
+        _ kind: QuantificationBehavior = .eagerly
```
rxwei, line 357: Nit
```suggestion
        _ behavior: QuantificationBehavior = .eagerly
```

Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -360,21 +352,23 @@ struct VariadicsGenerator: ParsableCommand {
...
       ) -> \(regexTypeName)<\(matchType)> \(whereClause) {
-        .init(node: .quantification(.\(kind.astQuantifierAmount), .eager, component.regex.root))
+        .init(node: .quantification(.\(kind.astQuantifierAmount), kind.astKind, component.regex.root))
       }
 
       \(arity == 0 ? "@_disfavoredOverload" : "")
-      public func \(kind.rawValue)<\(genericParameters(withConstraints: true))>(
+      public func \(kind.rawValue)<\(genericParams)>(
+        _ kind: QuantificationBehavior = .eagerly,
```
rxwei, line 364:
```suggestion
        _ behavior: QuantificationBehavior = .eagerly,
```

Sources/VariadicsGenerator/VariadicsGenerator.swift
natecook1000: Ah, missed these!

Sources/_StringProcessing/RegexDSL/DSL.swift
natecook1000: Agreed on both these counts — good call!

Sources/VariadicsGenerator/VariadicsGenerator.swift
natecook1000: I can't decide about this. On the one hand, it's nice to have them in a consistent order. On the other, I think it reads better to have the thing you're matching come first, and only shift it to the end when it's in the trailing closure position. IIUC, this is how SwiftUI handles it pretty much across the board — for example, when creating a `Button` with just a label, the string is the first parameter, otherwise, the builder closure comes at the end.

Let's punt on this for now to see how many different calls like this we have.

natecook1000:
@swift-ci Please test

rxwei:
Sorry for the merge conflicts!

Sources/_StringProcessing/RegexDSL/DSL.swift
milseman: Would these instead be properties returning `Self` if we're going that route?

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -62,6 +62,44 @@ extension CharacterClass: RegexProtocol {
 
 // Note: Quantifiers are currently gyb'd.
 
+/// Specifies how much to attempt to match when using a quantifier.
+public struct QuantificationBehavior {
```
milseman, line 66: Thoughts on this being a "thing" in API vs a set of operations on quantifiable patterns? E.g. there could be a sub-protocol for quantifiables with a `var eagerly: Self` requirement.

(Not arguing one way or another)

Tests/RegexTests/RegexDSLTests.swift
```diff
@@ -118,22 +143,47 @@ class RegexDSLTests: XCTestCase {
...
+      }
+    }
+
+    try _testDSLCaptures(
+      ("abc1def2", ("abc1def2", ["1", "2"])),
+      captureType: (Substring, [Substring]).self, ==)
+    {
+      oneOrMore {
+        oneOrMore(.word, .reluctantly)
+        CharacterClass.digit.capture()
```
milseman, line 165: Can we drop `CharacterClass`?

At some point I had to add a RegexWithComponent protocol and some overloads to make the static vars discoverable, but @rxwei was able to remove that and keep the lookup. I don't recall the issues there.

milseman (approved):
LGTM

Tests/RegexTests/RegexDSLTests.swift
natecook1000: Is the compiler able to distinguish between `.digit` being a method of `oneOrMore(...)`'s return type and it being a static member call on its own? I haven't been able to get that kind of lookup to work for static protocol methods.

PR #152, opened 2022-02-10
Conform `Substring` to `RegexProtocol`

rxwei:
Currently `String` conforms but `Substring` doesn't. I don't see a reason `Substring` shouldn't be usable in a DSL builder block, so adding this conformance.


rxwei:
@swift-ci please test Linux

milseman (approved):
Should this instead be constructing a `DSLTree`? Also, I'd probably go with a `.quotedLiteral` instead of a concatenation of atoms inside the DSL. (Inside the AST it would affect printing, but for the DSL I think it's the right call).

PR #154, opened 2022-02-10
Add assertions to the DSL

natecook1000:
This adds an `Anchor` type that handles the different kinds of assertions supported by regular expression literals. Some spelling are different — instead of separate `\w` and `\W` assertions, this design provides just `Anchor.wordBoundary` as well as an `inverted` property that is available on all assertions. (This property should perhaps be named `negated` instead.) That is, `\W` in a regex literal maps to `Anchor.wordBoundary.inverted`.

A `lookahead(isNegative:)` function provides the positive/negative lookahead functionality from regex literals.


Sources/_StringProcessing/RegexDSL/Assertion.swift
```diff
@@ -0,0 +1,134 @@
...
+    case startOfSubject
+    case endOfSubjectBeforeNewline
+    case endOfSubject
+    case firstMatchingPositionInSubject
+    case textSegmentBoundary
+    case startOfLine
+    case endOfLine
+    case wordBoundary
+    case lookahead(DSLTree.Node)
+  }
```
milseman, line 25: Is this meant to be a listing of built-in assertions, or are each of these the kinds of assertions someone could write?

Sources/_StringProcessing/RegexDSL/Assertion.swift
```diff
@@ -0,0 +1,134 @@
...
+      }
+    } else {
+      switch kind {
+      case .startOfSubject: fatalError("Not yet supported")
+      case .endOfSubjectBeforeNewline: fatalError("Not yet supported")
+      case .endOfSubject: fatalError("Not yet supported")
+      case .firstMatchingPositionInSubject: fatalError("Not yet supported")
+      case .textSegmentBoundary: return .notTextSegment
+      case .startOfLine: fatalError("Not yet supported")
+      case .endOfLine: fatalError("Not yet supported")
```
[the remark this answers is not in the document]
milseman: Why are we going to an AST here?

Tests/RegexTests/RegexDSLTests.swift
```diff
@@ -180,6 +180,31 @@ class RegexDSLTests: XCTestCase {
...
   }
+  
+  func testAssertions() throws {
+    try _testDSLCaptures(
+      ("aaaaab", "aaaaab"),
+      ("caaaaab", nil),
+      ("aaaaabc", nil),
+      captureType: Substring.self, ==)
+    {
+      Assertion.startOfLine
```
milseman, line 191: Many of the built-in ones are more commonly called "anchors", which might be worth considering too.

Sources/_StringProcessing/RegexDSL/Assertion.swift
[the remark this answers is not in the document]
natecook1000: `DSLTree` tracks assertions using `AST.Atom.AssertionKind` right now.

Sources/_StringProcessing/RegexDSL/Assertion.swift
```diff
@@ -0,0 +1,121 @@
...
+
+  public static var wordBoundary: Anchor {
+    Anchor(kind: .wordBoundary)
+  }
+  
+  public var inverted: Anchor {
+    var result = self
+    result.isInverted.toggle()
+    return result
+  }
```
milseman, line 106: Would we want an `isInverted` then? Also, is it the case that all anchors can be inverted?

milseman (approved):
Mostly LGTM.

natecook1000:

Sources/_StringProcessing/RegexDSL/Assertion.swift
```diff
@@ -0,0 +1,121 @@
...
+}
+
+public func lookahead<R: RegexProtocol>(
+  isNegative: Bool = false,
```
rxwei, line 110: While generally Boolean properties should read like an assertion about the receiver, I'm not sure it's quite as useful for function parameters. In this case, `isNegative` is not forming a phrase with the base name to produce Boolean result, but a dictation by the caller to modify the behavior of the callee. As such, I wonder if we should call it `negative` instead.

natecook1000: You're right, this is like the second parameter in `split(separator: "-", omittingEmptySubsequences: false)`. 👍🏻

rxwei (approved):

natecook1000:

natecook1000:

PR #161, opened 2022-02-14
Add `repeatMatch` functions to DSL

natecook1000:
These allow you to spell literal expressions like `a{3,}` or `a{2,5}` when using the DSL. I've gone with the `repeatMatch` name to align a bit with `repeatElement` in the stdlib, and because `repeating(...)` sounds too much to my ear like you have to repeat the matched text, not repeat the match, if that makes sense.

Usage:

```swift
let regex = Regex {
    repeatMatch(2...) {
        repeatMatch(CharacterClass.word, count: 3)
        CharacterClass.digit
    }
}
"abc1".matches(regex)        // false
"abc1def2".matches(regex)    // true
"abc1defg2".matches(regex)   // false
```

### DIFF
--- a/Sources/_StringProcessing/RegexDSL/DSL.swift
+++ b/Sources/_StringProcessing/RegexDSL/DSL.swift
@@ -145,6 +145,34 @@ postfix operator .?
 postfix operator .*
 postfix operator .+
 
+/// Generates a DSLTree node for a repeated range of the given DSLTree node.
+/// Individual public API functions are in the generated Variadics.swift file.
+internal func _repeatingNode(
+  _ range: Range<Int>,
+  _ behavior: QuantificationBehavior,
+  _ node: DSLTree.Node
+) -> DSLTree.Node {
+  // TODO: Throw these as errors
+  assert(range.lowerBound >= 0, "Cannot specify a negative lower bound")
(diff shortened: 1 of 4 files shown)


milseman (approved):
LGTM, I'll let @rxwei comment on generator script.

My vote is for
```
repeat {
  ...
}

repeat(3...) {
  ...
}
```

I think `repeat` as a verb is fine alongside `capture`, but I'm not a picky grammar namer.

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -145,6 +145,34 @@ postfix operator .?
 postfix operator .*
 postfix operator .+
 
+/// Generates a DSLTree node for a repeated range of the given DSLTree node.
+/// Individual public API functions are in the generated Variadics.swift file.
+internal func _repeatingNode(
```
rxwei, line 150: Nit: How about a static method `DSLTree.Node.repeating(_:_:_:)`

rxwei (approved):
LGTM. I'd vote for `repeat` as the name.

natecook1000:
@swift-ci Please test

PR #162, opened 2022-02-15
Drop arrays and unzip tuples for quantified captures.

rxwei:
To bridge the typing gap between regex literals and regex builder, this patch introduces the following changes:

1. Repeating quantifiers no longer produce an array capture. Instead, produce an optional capture when the lower bound is `0`, otherwise an atom capture. After matching, this capture corresponds to the last occurrence in the match. A history-saving quantifier variant will be introduced at a later time.
2. Applying a quantifier on a tuple-capturing regex will map quantification onto every tuple element, resulting in a tuple of quantified captures (unzipped).

-----

```swift
let literal = /a((b)(b)+)*(c)+(d)?/
//           0 ^~~~~~~~~~~~~~~~
//            1 ^~~~~~
//             2 ^~~
//                3 ^~~
//                   4 ^~~
//                       5 ^~~
// => Regex<(Substring, Substring?, Substring?, Substring?, Substring, Substring?)>

// Equivalent regex using regex builder:
let dsl = Regex { // 0
  "a"
  zeroOrMore {
    capture { // 1
      capture { // 2
        "b"
      }
      oneOrMore {
        capture { // 3
          "b"
        }
      }
    }
  }
  oneOrMore {
    capture { // 4
      "c"
    }
  }
  optionally {
    capture { // 5
      "d"
    }
  }
}

// => Regex<(Substring, Substring?, Substring?, Substring?, Substring, Substring?)>
```

### DIFF
--- a/Sources/VariadicsGenerator/VariadicsGenerator.swift
+++ b/Sources/VariadicsGenerator/VariadicsGenerator.swift
@@ -294,8 +294,8 @@ struct VariadicsGenerator: ParsableCommand {
     var operatorName: String {
       switch self {
       case .zeroOrOne: return ".?"
-      case .zeroOrMore: return ".+"
-      case .oneOrMore: return ".*"
+      case .zeroOrMore: return ".*"
+      case .oneOrMore: return ".+"
       }
     }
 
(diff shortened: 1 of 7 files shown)


rxwei:
@swift-ci please test Linux

milseman (commented):
Mostly LGTM, but I'm still unsure about `*` producing optional.

Also, while some of the capture tests have equivalent versions for `*` and `+`, I eventually stopped making copies for `+` since it was the same. Now that it's different, we may want to introduce it back in. 

On the other hand, with the dropping of Array, so many things should start working so we can still overhaul those tests.

rxwei:
@swift-ci please test Linux

rxwei:
> but I'm still unsure about `*` producing optional.

Discussed in person and converged on producing optional as PCRE produces null on empty match.  Going to merge this to unblock future work.  We can add more quantifier tests in a different PR.

PR #163, opened 2022-02-16
Custom regex components support

milseman:
Basic support for custom matcher regex components.

This is only lightly tested, and I think we'll be cleaning up a lot of the structuralization code anyways soon. But this should unblock a lot.


milseman:

Tests/RegexTests/CustomTests.swift
```diff
@@ -0,0 +1,113 @@
...
+}
+
+extension Nibbler {
+  // Default implementation, just feed the character in
+  func matchingConsuming(
+    _ consumed: Consumed,
+    in range: Range<Consumed.Index>
+  ) -> (upperBound: Consumed.Index, match: Match)? {
+    guard !range.isEmpty else {
+      // FIXME: Do we have a full story here?
```
natecook1000, line 18: `range` should be allowed to be empty, and (when we add it) `currentIndex` should be allowed to be any index in `range.lowerBound...range.upperBound`. Otherwise you wouldn't be able to implement something like `$` using this interface.

Tests/RegexTests/CustomTests.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    XCTAssertEqual(res2.match, "123")
+
+    let regex3 = Regex {
+      capture {
+        oneOrMore {
+          Numbler()
+        }
+      }
+    }
```
natecook1000, line 87: This behavior seems just right — if you capture a custom type (as below), you can get that specific type's match, but capturing an enclosing element means you just get the substring.

natecook1000 (approved):
LGTM

Tests/RegexTests/CustomTests.swift
itingliu: I think the interface we talked yesterday was something like 
```swift
func matchingConsuming(
    _ consumed: Consumed,
    startingAt index: Consumed.Index, /* currentIndex in @natecook1000's comment */
    in range: Range<Consumed.Index>
  ) -> (upperBound: Consumed.Index, match: Match)?
```

and so `_MatcherInterface` will have to be updated to take the `startingAt` index too, right?

Sources/_StringProcessing/RegexDSL/DSLConsumers.swift
```diff
@@ -0,0 +1,32 @@
...
+  }
+}
+
+extension CollectionConsumer where Consumed == String {
+  public var regex: Regex<Void> {
+    Regex(node: .consumer {
+      self.consuming($0, in: $1)
+    })
+  }
+}
```
timvermeulen, line 26: I don't think we'll need these once we add the `CustomRegexComponent` protocol (but I can remove these when that happens).

Tests/RegexTests/CustomTests.swift
timvermeulen: Yeah, `_MatcherInterface` will include an extra index parameter. I'll follow up with a PR that makes this change and introduces the `CustomRegexComponent` protocol

timvermeulen (approved):

rxwei (approved):
🎉

Sources/_StringProcessing/RegexDSL/DSLConsumers.swift
rxwei: Before we have that protocol, do we also want overloads that take `Substring`?

PR #168, opened 2022-02-17
Adds support for case insensitivity

natecook1000:
We don't want to do this all through consumer functions, but this allows for case insensitive matching throughout.

### DIFF
--- a/Sources/_StringProcessing/ConsumerInterface.swift
+++ b/Sources/_StringProcessing/ConsumerInterface.swift
@@ -56,28 +56,28 @@ extension DSLTree.Atom {
   func generateConsumer(
     _ opts: MatchingOptions
   ) throws -> MEProgram<String>.ConsumeFunction? {
-    let isCaseSensitive = opts.isCaseSensitive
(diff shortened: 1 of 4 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -51,24 +51,34 @@ extension DSLTree.Node {
...
+            : nil
         }
-        return input.index(after: low)
       }
     case let .scalar(s):
-      return consumeScalar { $0 == s }
+      return consumeScalar {
+        isCaseInsensitive
+          ? $0.properties.lowercaseMapping == s.properties.lowercaseMapping
+          : $0 == s
```
milseman, line 80: I still feel like there's quite a bit of duplicated code throughout. Would a helper function, similar to how `consumeScalar` is a helper function for making predicate-based scalar consumers, alleviate this some? Maybe something that is parameterized over case sensitivity?

PR #169, opened 2022-02-17
Update \DDD parsing

hamishknight:
Previously we followed PCRE's parsing of this syntax such that it may either be an octal sequence or backreference depending on a list of heuristics. However this model is complicated and not particularly intuitive, especially as there are other engines that disambiguate using subtly different rules.

Instead, always parse `\DDD` as a backreference, unless it begins with `0`, in which case it is an octal sequence. This matches ICU and Java's behavior. Once we start validating group references, we can then start emitting an error on invalid backreferences using this syntax, and suggest prefixing with 0 if an octal sequence is desired.


hamishknight:
@swift-ci please test Linux

milseman (approved):
LGTM

hamishknight:
@swift-ci please test

PR #171, opened 2022-02-18
Option APIs for RegexProtocol

natecook1000:
This adds an API to any conforming type that lets the user control whether a match uses case sensitive comparison from within a DSL declaration. Applying modifiers like this has the opposite precedence as a regex literal, where the nearest option setter has precedence.

For example, these two regex declarations are equivalent:

```swift
let regexDSL = Regex {
    oneOrMore {
        "a"
    }
    .caseSensitive(true)
    .caseSensitive(false)
}

let regexLiteral = #/(?-i)(?i)a+/#
```

Note: This requires #168.
Note 2: More options TK once we're okay with this approach.

### DIFF
--- a/Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
+++ b/Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
@@ -110,6 +110,16 @@ extension AST {
   }
 }
 
+extension AST.MatchingOptionSequence {
+  public init(adding: [AST.MatchingOption]) {
+    self.init(caretLoc: nil, adding: adding, minusLoc: nil, removing: [])
+  }
+
+  public init(removing: [AST.MatchingOption]) {
+    self.init(caretLoc: nil, adding: [], minusLoc: nil, removing: removing)
+  }
+}
+
 extension AST.MatchingOption: _ASTPrintable {
   public var _dumpBase: String { "\(kind)" }
 }
--- a/Sources/_StringProcessing/MatchingOptions.swift
+++ b/Sources/_StringProcessing/MatchingOptions.swift
@@ -189,6 +189,19 @@ extension MatchingOptions {
       contains(.init(kind))
     }
     
+    mutating func add(_ opt: Option) {
+      // If opt is in one of the mutually exclusive groups, clear out the
+      // group before inserting.
+      if Self.semanticMatchingLevels.contains(opt.representation) {
+        remove(.semanticMatchingLevels)
+      }
+      if Self.textSegmentOptions.contains(opt.representation) {
+        remove(.textSegmentOptions)
+      }
+
+      insert(opt.representation)
+    }
+    
     /// Applies the changes described by `sequence` to this set of options.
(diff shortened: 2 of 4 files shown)


kylemacomber:
In SwiftUI, the tighter binding modifier wins. I think it'd be good to follow the same convention so when people more between result builder DSLs they don't have to turn their brain inside-out.

Tighter binding is also consistent with regex literals and variable binding in general.

rxwei (approved):

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -51,24 +52,34 @@ extension DSLTree.Node {
...
-        guard input[low] == c else {
-          return nil
+        if isCaseSensitive || !c.isCased {
+          return input[low] == c
+            ? input.index(after: low)
+            : nil
+        } else {
+          return input[low].lowercased() == c.lowercased()
+            ? input.index(after: low)
+            : nil
```
milseman, line 74: This code/logic is repeated a lot. Any chance to refactor it a little?

Tests/RegexTests/MatchTests.swift
```diff
@@ -1117,6 +1121,43 @@ extension RegexTests {
...
+      ("cafe", false),
+      ("CAFE", false),
+      ("caFE", true),
+      ("EFac", true))
+    matchTest(
+      #"(?i)[caFE]{4}"#,
+      ("cafe", true),
+      ("CaFe", true),
+      ("EfAc", true))
+  }
```
milseman, line 1159: Thanks for the tests! What about multi-scalar expansions, like the German SS?

natecook1000: I believe we need case folding to do all this properly, don't we? I'll follow up with some tests showing the holes.

Sources/_StringProcessing/RegexDSL/DSLTree.swift
```diff
@@ -198,7 +202,22 @@ extension DSLTree.Atom {
 
 extension DSLTree {
   struct Options {
-    // TBD
+    enum Kind {
```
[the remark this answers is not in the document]
natecook1000: I flattened this out — with the SwiftUI-style options, we don't need to coalesce, so the existing `MatchingOptionsSequence` works fine.

natecook1000:

natecook1000:

Sources/_MatchingEngine/Regex/AST/MatchingOptions.swift
```diff
@@ -110,6 +110,16 @@ extension AST {
   }
 }
 
+extension AST.MatchingOptionSequence {
```
milseman, line 113: FWIW, when we pay off some technical debt, DSLTree will have fewer and fewer AST constructs in it.

In general, we have 3 things, the AST's representation, the API's representation, and the model type.

Literal/AST _lowers-to_ model
API _lowers-to_ model

It's possible in some situations for all 3 to be the same type, or for the model to also be made public in API.

PR #172, opened 2022-02-18
Add test for backreference before group

natecook1000:
A literal can specify a backreference that references a group before the group is opened. This most frequently shows up with a repeated alternation - the first part of an alternation can include a reference to a group in a different alternation, and simply doesn't match anything until the group has matched and captured a portion of the input.


natecook1000:

PR #177, opened 2022-02-22
Simplify Capture and DynamicCaptures storage

milseman:
Change to an array of structs instead of recursive indirect enums.

### DIFF
--- a/Sources/_StringProcessing/Capture.swift
+++ b/Sources/_StringProcessing/Capture.swift
@@ -11,112 +11,57 @@
 
 import _MatchingEngine
 
-// TODO: what here should be in the compile-time module?
+/// A structured capture
+struct StructuredCapture {
+  /// The `.optional` height of the result
+  var numOptionals = 0
 
-enum Capture {
-  case atom(Any)
-  indirect case tuple([Capture])
-  indirect case some(Capture)
-  case none(childType: AnyType)
-  indirect case array([Capture], childType: AnyType)
-}
+  var storedCapture: StoredCapture?
 
-extension Capture {
-  static func none(childType: Any.Type) -> Capture {
-    .none(childType: AnyType(childType))
-  }
-
-  static func array(_ children: [Capture], childType: Any.Type) -> Capture {
-    .array(children, childType: AnyType(childType))
+  var numSomes: Int {
+    storedCapture == nil ? numOptionals - 1 : numOptionals
   }
 }
 
-extension Capture {
-  static func tupleOrAtom(_ elements: [Capture]) -> Self {
-    elements.count == 1 ? elements[0] : .tuple(elements)
-  }
+/// A storage form for a successful capture
+struct StoredCapture {
+  // TODO: drop optional when engine tracks all ranges
+  var range: Range<String.Index>?
 
-  static var void: Capture {
-    .tuple([])
-  }
+  // If strongly typed, value is set
+  var value: Any? = nil
+}
 
-  var value: Any {
-    switch self {
-    case .atom(let atom):
-      return atom
-    case .tuple(let elements):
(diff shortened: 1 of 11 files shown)


milseman:
@swift-ci please test

Sources/_StringProcessing/RegexDSL/DynamicCaptures.swift
```diff
@@ -17,28 +17,36 @@ extension Regex where Match == (Substring, DynamicCaptures) {
...
 
-public enum DynamicCaptures: Equatable {
-  case substring(Substring)
-  indirect case tuple([DynamicCaptures])
-  indirect case optional(DynamicCaptures?)
-  indirect case array([DynamicCaptures])
-
-  public static var empty: Self {
-    .tuple([])
+public struct DynamicCapture: Hashable {
```
rxwei, line 20: - This is a user-facing type but it does not have any public API. How would a user access it?
- Since it now represents a single capture rather than a tree, should we name it `AnyCapture` instead?
- We can have properties `public var base: Any` and `public var range: Range<String.Index>`.

milseman: 1. Is there any reason for DynamicCaptures to be public other than for the initializer taking one?
2. SGTM
3. Err, are back to exposing this to users?

milseman:
@rxwei updated. I think some of the remaining questions are what API are we exposing in the dynamic case. I'm trying to work towards a world where we have the freedom to have a different storage representation than literally storing `Match`, but I'm not sure how this applies to `DynamicCaptures` and/or whether `AnyCapture` could include values.

Sources/_StringProcessing/RegexDSL/DynamicCaptures.swift
rxwei: > Is there any reason for DynamicCaptures to be public other than for the initializer taking one?

The user would need to access the captured range and substring.

>  It would be `AnySubstringCapture`

Sure.

milseman: > The user would need to access the captured range and substring. How else would the user do it?

We want API to get the number of captures, and for a given capture to get the slice/range/etc. I was thinking this would be more like static captures, where the more common approach API is on `RegexMatch` rather than `Match`. That is, `enum DynamicCaptures {}` would suffice as just a marker.

Or, we can have a dynamic capture type which holds the API. I'm not too particular on one approach over the other, just wondering what the rationale was or the tradeoffs.

milseman:
@rxwei I pushed a simple mock API for dynamic captures. I feel like at this point, designing the API is a separate discussion/PR than designing the storage representation, but I'm providing some basic API for future prototyping purposes.

milseman:
@swift-ci please test

rxwei (approved):

Sources/_StringProcessing/Capture.swift
```diff
@@ -32,21 +32,42 @@ struct StoredCapture {
...
+    someCount = optionalCount
+  } else {
+    // Ok since we Any-box every step up the ladder
+    underlying = Optional<Any>(nil) as Any
+    someCount = optionalCount - 1
+  }
+
+  for _ in 0..<someCount {
+    underlying = Optional(underlying) as Any
+  }
```
rxwei, line 58: The recursive type erasure is unfortunate and will make the final conversion (i.e. `as! Match`) slower.  This could've opened the existentials so that only the outer result is type-erased.
```swift
  for _ in 0..<someCount {
    func wrap<T>(_ x: T) {
      underlying = Optional(x) as Any
    }
    _openExistential(underlying, do: wrap)
  }
```

I've applied this in #182.

PR #180, opened 2022-02-22
Implement a basic benchmark suite

Azoy:


Sources/Benchmark/main.swift
```diff
@@ -0,0 +1,330 @@
...
+    let match = try nativeMatch(#"a(?#. comment)b"#, input: "123abcxyz")
+    
+    blackHole(match)
+  }
+}
+
+@inline(never)
+func run_simple_comment_nsregex(_ N: Int) throws {
+  for _ in 0 ..< N {
+    let match = try nsregexMatch(#"a(?#. comment)b"#, input: "123abcxyz")
```
milseman, line 311: Do we want a function that scales up the input length (so we have longer inputs and fewer compilations)? Similarly, do we want to have all-ranges style benchmarks?

Sources/Benchmark/regexredux.swift
```diff
@@ -0,0 +1,77 @@
...
+                                           options: .regularExpression)
+  
+  let codeLength = sequence.utf8.count
+  
+  var resultLength = 0
+  
+  let group = DispatchGroup()
+  
+  group.enter()
+  DispatchQueue.global().async {
```
milseman, line 33: If it's simpler, we can prototype or benchmark a single-threaded approach primarily.

PR #181, opened 2022-02-22
Store integer registers in SavePoint

natecook1000:
Without saving the int registers, quantifications can't accurately keep track of their loop count when backtracking, so some patterns were incorrectly matching.

For example, `/(a|a){3}a/` incorrectly matched against `"aaaxyz"`, because after failing to match the final `a` against the `"x"`, the engine backtracked to the second alternation without resetting the number of times the quantification had been satisfied.

### DIFF
--- a/Sources/_StringProcessing/Engine/Backtracking.swift
+++ b/Sources/_StringProcessing/Engine/Backtracking.swift
@@ -29,13 +29,18 @@ extension Processor {
     // perhaps current start)
     var captureEnds: [_StoredCapture]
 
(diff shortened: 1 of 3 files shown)


natecook1000:

milseman (approved):
LGTM.

Another area where this bug could surface is a nested reluctant quantification, where after a restore the old int register values would still be around.

Long term, I think we want to avoid having such heavy-weight save points, but that would require switching the engine off of generalized control flow.

PR #182, opened 2022-02-22
Generic `Reference` type and match result subscript

rxwei:
Change `Reference` to generic type over `Capture`. Add a `subscript<T>(_: Reference<T>) -> T` to `RegexMatch` to allow the user to access a capture via a reference.

Alternatives considered:
- `subscript(_: Reference) -> Any` with non-generic `Reference`. While creating a `Reference` would remain easy, it would lose the benefits of strong types.
- `Reference<Capture>.init() where Capture == Substring`. This allows writing `Reference()` to conveniently create a `Reference<Substring>`, but when it’s used on any transformed capture, it causes the “expression is too complex” error. I think this is a harmful outcome as it's hard for developers to figure that the fix is actually specifying a type in the `Reference` initializer.

-----

Examples:

```swift
let a = Reference(Int.self)
let result = regex.match {
  tryCapture(as: a) {
    oneOrMore(.digit)
  } transform: { Int($0) }
}
result[a] // => Int
```


rxwei:
@swift-ci please test Linux@swift-ci please test

natecook1000:
Now that we've flattened the structured capture groups, can the `capture(_:as:)` method capture anything other than a `Substring`? I'm not sure the generic parameter is required in the non-transforming overload. Without that, we should be able to revert to the optional/`nil` default approach, which cuts the number of overloads in half.

rxwei:
Yes, the new capture type produced by `capture` and `tryCapture` is the whole match type of the component.

Sources/_StringProcessing/RegexDSL/Match.swift
```diff
@@ -11,8 +11,24 @@
 
 @dynamicMemberLookup
 public struct RegexMatch<Match> {
+  let input: String
```
milseman, line 14: IIUC, this will (eventually) allow our storage form to be purely range based.

Tests/RegexTests/RegexDSLTests.swift
```diff
@@ -559,11 +559,11 @@ class RegexDSLTests: XCTestCase {
...
   func testBackreference() throws {
     try _testDSLCaptures(
-      ("abc#41#42abc#42#42", ("abc#41#42abc#42#42", "abc", 42, "#42", nil)),
+      ("abc#41#42abcabcabc", ("abc#41#42abcabcabc", "abc", 42, "abc", nil)),
       matchType: (Substring, Substring, Int?, Substring?, Substring?).self, ==)
     {
-      let a = Reference()
-      let b = Reference()
+      let a = Reference(Substring.self)
+      let b = Reference(Int.self)
```
milseman, line 566: Do we have tests where the types don't match up? What happens?

milseman (approved):
LGTM

Tests/RegexTests/RegexDSLTests.swift
rxwei: You won't be able to capture something onto a `Reference` of a non-matching type. That's a type error.

milseman: That's very compelling and alleviates the need for another way to name captures.

BTW, outside of scope of this PR, are we still trying to figure out what happens when a regex literal has names? Do we drop them completely, synthesize these references, or also have a stringy lookup?

rxwei: I think we drop them completely with the `._Unlabeld` extension I'm working on.  I'm aiming to create a PR for that later today.

Sources/_StringProcessing/RegexDSL/Match.swift
rxwei: This is now purely range based.

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

PR #183, opened 2022-02-23
Delete `groupTransform` from AST.

rxwei:
This is unused.


rxwei:
@swift-ci please test Linux

PR #186, opened 2022-02-25
Work around compiler bug rdar://89449323

rxwei:
This was uncovered when testing dev/7 against apple/swift#41430.


rxwei:
@swift-ci please test

PR #187, opened 2022-02-25
Regex Literal Pitch v2

hamishknight:


Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,165 @@
...
+## Detailed Design
+
+A regular expression literal will be introduced using `re'...'` delimiters, within which the compiler will parse a regular expression (the details of which are outlined in **TODO: Internal syntax link**):
+
+```
+// Matches "<identifier> = <hexadecimal value>", extracting the identifier and hex number
+let regex = re'([[:alpha:]]\w*) = ([0-9A-F]+)'
+```
+
+The use of a two letter prefix allows for easy future extensibility of such literals, by allowing different prefixes to indicate different types of literal. **TODO: examples**
```
milseman, line 22: (super early musings). I think it's interesting to contrast single-letter, double-quotes, and full spelling (e.g. `regex'...'`).

In my mind, we're proposing the following world:

1. Regexes are special enough to be "blessed" in some fashion
2. Regexes are not uniquely special though, other foreign algorithm literals are similarly arguable
3. We want to keep the possibility of supporting library-provided foreign algorithm literals in the (distant?) future

Single-letter namespaces run out fast, and it's not the clearest in the world if `r'...'` means raw or regex, while `re'...'` is very clear. `regex'...'` might be even more clear, but it feels (at least to me) overdone.

I think future extension to an extended Swift regex syntax using `rx'...'` is a nice-to-have and can serve as a clear example of extensibility. But, it's not super compelling on its own.

I don't have a lot of experience using other embedded languages like graphql or sql, so I'm not sure how those would play out. I don't know if there's something similar (i.e. a compact common notation) in the numerics world (@stephentyrone). 

Double quotes seem good on the surface, but these are very much not strings. Backslashes mean something completely different, etc. These are _foreign program literals_ not _data literals_. Raw strings are a way to write run-time regexes for use with NSRegularExpression. But I haven't really thought this argument out a ton either.

milseman: I wonder about `sh'...'` for shelling out, etc.

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,165 @@
...
+
+```
+re'''
+  abc
+  def
+  '''
+```
+
+This would follow the precedent set by [SE-0168] for multi-line string literals, and obey the same rules, in particular with the stripping of any leading whitespace prior to the position of the closing delimiter.
+
```
milseman, line 54: What about line-terminating comments that use `#`, as is typical in multi-line mode? That is, the use of multi-line delimiter is an "option" relevant to the parsing of the foreign program literal.

Would this only be a problem if the line-ending comment was immediately followed by a backslash?

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,165 @@
...
+We could choose to use double quotes instead of single quotes. This would be similar in appearance to string literals, however it could be argued that regex literals are distinct from string literals in that they introduce their own specific language to parse. As such, regex literals are more like "program literals" than "data literals", and the use of single quote instead of double quote could express this difference.
+
+### Single letter `r'...'`
+
+We could choose to shorten the literal prefix to just `r`. However this could potentially be confused to mean "raw", especially as Python uses this syntax for raw strings.
+
+### Forward slashes `/.../`
+
+Forward slashes are a regex term of art, and are used as the delimiters for regex literals in Perl, JavaScript and Ruby. However, they would be an awkward fit in Swift's language grammar, and would not provide a path for extensibility.
```
milseman, line 67: And a huge note is that even in Perl and Ruby, a common practice is to use alternate delimiters to avoid the readability or escaping issues with `/`.

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,165 @@
...
+
+Finally, there would be a minor ambiguity with infix operators used with regex literals. When used without whitespace, e.g `x+/y/`, the expression will be treated as using an infix operator `+/`. Whitespace is therefore required `x + /y/` for regex literal interpretation.
+
+#### Language changes required
+
+In addition to ambiguities listed above, there are also some parsing ambiguities that would require the following language changes:
+
+- Deprecation of prefix operators containing the `/` character.
+- Changing parsing rules such that `/` starting a new line forms a regex literal, which may break source for operator chains over multiple lines.
+
```
milseman, line 97: A huge point is that in a result builder, such as the regex DSL for whom intermixing regex literals is a high-impact feature, slash would continue an expression instead of define a new one. Would we treat `re'...'` differently, since it's clearly not a binary operator?

Documentation/Evolution/DelimiterSyntax.md
milseman: Oh, that's an interesting tidbit. I think the argument for `rx'...'` next release might be stronger in that case.

For the line-ending comments with `#`, I was wondering how that could interact with raw literals. I think it should be fairly straightforward to explain though.

hamishknight:
> I vote for insisting on raw syntax here. Every little heuristic like this is a strike against elegance. It's something that every editor will have to deal with. Raw literals are an obvious extension, and are probably easier to implement.

Yeah, I think I agree. I guess it ultimately depends on what the raw syntax rules regarding backslash end up being, if they end up treating backslashes as literal, we may not want to recommend raw syntax, as that would require changing any escape sequences written. We would likely want the user to use the alternative spellings e.g `(?<...>)` instead.

However even if this is not something we support, I still feel there is some value in the compiler at least implementing the heuristic, as it only impacts invalid code, and allows us to effectively emit a diagnostic with a fix-it to change the regex. Though I don't have a good sense of how common the `(?'...')` syntax is in the wild.

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,388 @@
+# Regex Literals
+
+- Authors: [Hamish Knight](https://github.com/hamishknight), [Michael Ilseman](https://github.com/milseman), [David Ewing](https://github.com/DaveEwing)
+
+## Introduction
+
+This proposal helps complete the story told in *[Regex Type and Overview][regex-type]* and [elsewhere][pitch-status]. We propose the introduction of regex literals to Swift source code, providing compile-time checks and typed-capture inference.
```
milseman, line 7: > This proposal helps complete the story

Well, all these proposals do that 😁.

```suggestion
We propose the introduction of regex literals to Swift source code, providing compile-time checks and typed-capture inference. Regex literals help complete the story told in *[Regex Type and Overview][regex-type]*.
```

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Detailed design
+
+### Upgrade path
+
+Due to the source breaking changes needed for the `/.../` syntax, it will be introduced in Swift 6 mode. However, projects will be able to adopt it earlier by using the compiler flag `-enable-regex-literals`. Note this does not affect the extended syntax `#/.../#`, which will be usable immediately.
+
+### Named typed captures
+
+Regex literals have their capture types statically determined by the capture groups present. This follows the same inference behavior as [the DSL][regex-dsl], and is explored in more detail in *[Strongly Typed Captures][strongly-typed-captures]*. One aspect of this that is currently unique to the literal is the ability to infer labeled tuple elements for named capture groups. For example:
```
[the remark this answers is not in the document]
hamishknight: We do link the DSL here (`[the DSL][regex-dsl]`), should it be used as the main reference for the typed capture behavior? I was hesitant to talk more about typed captures here, as there's quite a bit to cover, and I believe most of it is shared behavior with the DSL. But maybe an overview would be reasonable?

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,388 @@
...
+```swift
+let regex = #/usr/lib/modules/([^/]+)/vmlinuz/#
+// regex: Regex<(Substring, Substring)>
+```
+
+The number of pounds may be further increased to allow the use of e.g `/#` within the literal. This is similar in style to the raw string literal syntax introduced by [SE-0200], however it has a couple of key differences. Backslashes do not become literal characters. Additionally, a multi-line mode, where whitespace and line-ending comments are ignored, is entered when the opening delimiter is followed by a newline.
+
+```swift
+let regex = #/
+  /usr/lib/modules/ # Prefix
```
hamishknight, line 101: Note this is a slightly different regex to the single-line version as it includes a `/` at the start, we probably ought to be consistent, what do you think? I mainly avoided it in the single-line version as GitHub syntax colors it as a comment 😬

Documentation/Evolution/DelimiterSyntax.md
```diff
@@ -0,0 +1,388 @@
...
+  
+  return (match.identifier, hex)
+}
+```
+
+This allows the captures to be referenced as `match.identifier` and `match.hex`, in addition to numerically (like unnamed capture groups) as `match.1` and `match.2`. This label inference behavior is not available in the DSL, however users are able to [bind captures to named variables instead][dsl-captures].
+
+### Extended delimiters `#/.../#`, `##/.../##`
+
+Backslashes may be used to write forward slashes within the regex literal, e.g `/foo\/bar/`. However, this can be quite syntactically noisy and confusing. To avoid this, a regex literal may be surrounded by an arbitrary number of balanced octothorpes. This changes the delimiter of the literal, and therefore allows the use of forward slashes without escaping. For example:
```
hamishknight, line 90: We call `#` a pound symbol elsewhere in the proposal, should we be consistent?

Documentation/Evolution/DelimiterSyntax.md
milseman: Any extra slashes are unintentional and the natural outcome of using slash as both part of a delimiter and an interior character.

It's interesting that `#//usr/local/#` and `#/user/local//#` would sensibly be lexed as comments by source tools. That's a downside of `/` we should mention too. I suppose the workaround is to escape the slash if it's at the very beginning or end, which is another little wrinkle that `#/.../#` doesn't alleviate.

Documentation/Evolution/DelimiterSyntax.md
milseman: FWIW Unicode calls it "number sign", which is less confusable with £ (whose official name is "POUND SIGN"). How does "number sign" consistently sound? Even in the US, we don't dial numbers that often anymore, so I expect "pound sign" to have limited life span, especially in the era of hashtags.

hamishknight: Updated to use "number sign", though a couple of occurrences were phrased as "number of ...", which sounded a bit awkward, so I changed those to use the `#` character directly.

airspeedswift:
This is looking good, the only things I think might want tweaking:
- I would spend less time on comments. It's a challenge, but not a major one
- I think it's worth acknowledging in the alternatives considered that in some communities such as Perl, the `/` syntax has become less popular.

hamishknight:
@swift-ci please test

PR #188, opened 2022-02-25
Pitch: String processing algorithms

itingliu:


Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+    /// - Returns: `true` if the initial elements of the sequence matches the
+    /// beginning of `regex`; otherwise, `false`.
+    public func starts<R: RegexProtocol>(with regex: R) -> Bool
+}
+```
+
+#### Trim prefix
+
+```swift
+extension Collection {
```
itingliu, line 149: I find myself squinting to see the differences between these functions. I wonder if it's easier to review if we group them by extension other than algorithms, like

```swift
extension Collection where Element: Equatable {
    public func contains<S: Sequence>(_ other: S) -> Bool
    public func trimmingPrefix<Prefix: Collection>(_ prefix: Prefix) -> SubSequence
}

extension BidirectionalCollection where SubSequence == Substring {
    public func contains<R: RegexProtocol>(_ regex: R) -> Bool
    public func starts<R: RegexProtocol>(with regex: R) -> Bool
}
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+let dateParser = Date.ParseStrategy(format: "\(day: .twoDigits) \(month: .abbreviated) \(year: .padded(4))"
+if let dateMatch = header.firstMatch(of: regex)?.0 {
+    let date = try? Date(dateMatch, strategy: dateParser)
+}
+```
+
+This works, but wouldn't it be much more approachable if you can directly use the date parser within the string match function?
```
milseman, line 66:
```suggestion
While this approach happens to work for this example, it is fragile and more onerous than just directly using Foundation's date parser directly within the regex.
```

I would recommend against rhetorical questions, especially "wouldn't it be better" style ones.

milseman: Also, you can consider streamlining some of this as part of an extended motivation section.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+CREDIT    04/06/2020    Paypal transfer    $4.99
+CREDIT    04/03/2020    Payroll            $69.73
+DEBIT     04/02/2020    ACH transfer       ($38.25)
+DEBIT     03/24/2020    IRX tax payment    ($52,249.98)
+"""
+```
+
+We have already seen that parsing a date string can be tricky since it could contain localized month name (`"Feb"` as seen from above). Parsing a currency string such as `$3,020.85` with regex is not trivial either -- it can contain grouping separators, a decimal separator, and a currency symbol, all of which can be localized. 
+
+Foundation has various parsers for localized strings like these. Delegating this task to dedicated parsers alleviates the need to handle it yourself. In the second part of the pitch, we introduce the `CustomRegexComponent` protocol that conveniently lets types from outside the standard library participate in regex builders and `RegexProtocol` algorithms.
```
milseman, line 90:
```suggestion
Parsing a date string can be tricky, as it could contain localized information (`"Feb"` as seen from above). Parsing a currency string such as `$3,020.85` with regex is also tricky -- it can contain grouping separators, a decimal separator, and a currency symbol, all of which can be localized. This is why Foundation provides industrial-strength parsers for localized strings like these.

 We propose a `CustomRegexComponent` protocol which allows types from outside the standard library participate in regex builders and `RegexProtocol` algorithms.
```

My suggestions are minor localized editorial changes. But, we're repeat the same information throughout, so there's more tidying up. Since this is the proposed solution section, it's a little weird to be talking about the "second part of this pitch", because this is where the second part goes.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+
+If `Foundation.FloatingPointFormatStyle<Double>.Currency` conformed to `CustomRegexComponent`, you would be able to retrieve the currency from the bank statement above as a list of `Double` values this way:
```
milseman, line 92:
```suggestion
`CustomRegexComponent` allows types, such as `Foundation.FloatingPointFormatStyle<Double>.Currency`, to be used directly within a regex.
```

I understand that you're not proposing the conformances for Foundation here, but I don't think the "if ... you would be able to..." adds clarity.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+    ///   there isn't a match.
+    func match(
+        _ input: String,
+        startingAt index: String.Index,
+        in bounds: Range<String.Index>
+    ) -> (upperBound: String.Index, match: Match)?
+}
+```
+
+You can conform your custom parser to `CustomRegexComponent`. Conformance is simple: implement the `match` function to return the upper bound of the matched substring, and the type represented by the matched range. It inherits from `RegexProtocol`, so you will be able to use it with all of the string algorithms that take a `RegexProtocol` type.
```
milseman, line 351:
```suggestion
Custom parsers can conform by implementing the `match` function to return the upper bound of the matched substring, and the value constructed by the match. Conformers naturally inherit from `RegexProtocol`, so they can be used with all of the string algorithms generic over `RegexProtocol`.
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+
+Here, we use Foundation framework's `FloatingPointFormatStyle<Double>.Currency` as an example. `FloatingPointFormatStyle<Double>.Currency` would conform to `CustomRegexComponent` by implementing the `match` function, with `Match` being a `Double`. It could also add a static function `.localizedCurrency(code:)` as a member of `RegexProtocol`, so you can refer to it as `.localizedCurrency(code:)` in the `Regex` result builder.
```
milseman, line 353: It really seems like this should be merged into the "proposed solution" sub-section, where the example's fresh in the reader's mind.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,402 @@
...
+        in bounds: Range<String.Index>
+    ) -> (upperBound: String.Index, match: Double)?
+}
+
+extension RegexProtocol where Self == FloatingPointFormatStyle<Double>.Currency {
+    public static func localizedCurrency(code: Locale.Currency) -> Self
+}
+```
+
+Users could specify a pattern to match a localized currency amount such as `"$3,020.85"` simply with the following, and use it in any of the string matching algorithms introduced above.
```
milseman, line 369:
```suggestion
Matching and extracting a localized currency amount, such as `"$3,020.85"`, can be done directly within a regex:
```

milseman (commented):
Just some wordsmithing and editorial adjustments. This is looking good so far, though there's opportunity to have less redundancy.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,515 @@
...
+    ///   sequence matching `regex` to replace. Default is `Int.max`.
+    public mutating func replace<R: RegexProtocol, Replacement: Collection>(
+        _ regex: R,
+        with replacement: Replacement,
+        maxReplacements: Int = .max
+    ) where Replacement.Element == Element
+  
+    public func replacing<R: RegexProtocol, Replacement: Collection>(
+        _ regex: R,
+        with replacement: (RegexMatch<R.Match>) throws -> Replacement,
```
itingliu, line 399: @timvermeulen @milseman I need help for documentation here. I'm not sure what this closure does.

milseman: The closure receives the full match info, including captures, and returns a replacement string.

milseman:
@itingliu let me know when's a good time to do another pass over the prose. I don't want to conflict or interrupt any reorganizing you're doing.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,545 @@
...
+
+We propose:
+
+1. New regex-powered algorithms over strings, bringing the standard library up to parity with scripting languages
+2. Generic `Collection` equivalents of these algorithms in terms of subsequences
+3. `protocol CustomRegexComponent`, allowing libraries to vend types that can be intermixed as components of regexes
+
+This proposal is part of a larger [regex-powered string processing initiative](https://forums.swift.org/t/declarative-string-processing-overview/52459). Throughout the document, we will reference the still-in-progress [`RegexProtocol`, `Regex`](https://github.com/apple/swift-experimental-string-processing/blob/main/Documentation/Evolution/StronglyTypedCaptures.md), and [result builder DSL](https://forums.swift.org/t/pitch-regular-expression-literals/52820), but these are in flux and not formally part of this proposal. Further discussion of regex specifics is out of scope of this proposal and better discussed in another thread (see [Pitch and Proposal Status](https://github.com/apple/swift-experimental-string-processing/issues/107) for links to relevant threads).
```
rxwei, line 13: The "result builder DSL" is linking to the wrong pitch though, the link points to the regex literal pitch. We don't have a DSL pitch yet — I think you can just use plain text for now, or link to https://github.com/rxwei/swift-experimental-string-processing/blob/dsl-api-walkthrough/Documentation/DSLAPIWalkthough.md (which I plan to turn into a pitch very soon).

itingliu: Thanks for catching that. I removed the link all together.

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,544 @@
...
+
+This proposal is part of a larger [regex-powered string processing initiative](https://forums.swift.org/t/declarative-string-processing-overview/52459). Throughout the document, we will reference the still-in-progress [`RegexProtocol`, `Regex`](https://github.com/apple/swift-experimental-string-processing/blob/main/Documentation/Evolution/StronglyTypedCaptures.md), and result builder DSL, but these are in flux and not formally part of this proposal. Further discussion of regex specifics is out of scope of this proposal and better discussed in another thread (see [Pitch and Proposal Status](https://github.com/apple/swift-experimental-string-processing/issues/107) for links to relevant threads).
+
+## Motivation
+
+Below is how [Python string processing API](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) compare to their counterparts in Swift. Many essential string processing APIs are evidently missing in the standard library. While most of those can be substituted by chaining multiple functions or using Foundation, they are basic enough to warrant a place in the standard library.
```
milseman, line 17:
```suggestion
Below is a comparison of how Swift's APIs stack up with Python's. Some of the gaps can be accomplished through a series of API calls, but every gap adds a burden to developers doing frequent or complex string processing.
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,544 @@
...
+| `splitlines(keepends)` | `split(separator:maxSplits:...)` |  |
+| `startswith(prefix, start, end)` | `starts(with:)` |  |  |
+| `strip([chars])` |  | Alternative: `trimmingCharacters(in:)` in Foundation  |
+| `translate(table)` |  |  |
+| `zfill(width)` |  | Alternative: `padding(toLength:)` in Foundation |
+
+
+Note: The comparison table omits functions to query if all characters in the string are of a specified category, such as `isalnum()` and `isalpha()`. These are achievable in Swift by passing in the corresponding character set to `allSatisfy(_:)`, so they're omitted in this table for simplicity. 
+
+### Processing domain-specific strings
```
milseman, line 52:
```suggestion
### Complex string processing
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,544 @@
...
+
+Another common task regarding string processing is handling those with domain specific information.
```
milseman, line 54:
```suggestion
Even with these API additions, more complex string processing quickly becomes unwieldy. Up-coming support for authoring regexes in Swift help alleviate this somewhat, but string processing in the modern world involves dealing with localization, standards-conforming validation, and other concerns for which a dedicated parser is required.
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,544 @@
...
+    }
+}
+
+let dateParser = Date.ParseStrategy(format: "\(day: .twoDigits) \(month: .abbreviated) \(year: .padded(4))"
+if let dateMatch = header.firstMatch(of: regex)?.0 {
+    let date = try? Date(dateMatch, strategy: dateParser)
+}
+```
+
+While this approach happens to work for this example, it is fragile when it comes to localized strings.
```
milseman, line 75:
```suggestion
This requires writing a simplistic pre-parser before invoking the real parser. The pre-parser will suffer from being out-of-sync and less featureful than what the real parser can do.
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,544 @@
...
+```swift
+let statement = """
+CREDIT    04/06/2020    Paypal transfer    $4.99
+CREDIT    04/03/2020    Payroll            $69.73
+DEBIT     04/02/2020    ACH transfer       ($38.25)
+DEBIT     03/24/2020    IRX tax payment    ($52,249.98)
+"""
+```
+
+Parsing a currency string such as `$3,020.85` with regex is also tricky, as it can contain localized and currency symbols. This is why Foundation provides industrial-strength parsers for localized strings like these.
```
milseman, line 88:
```suggestion
Parsing a currency string such as `$3,020.85` with regex is also tricky, as it can contain localized and currency symbols in addition to accounting conventions. This is why Foundation provides industrial-strength parsers for localized strings.
```

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -0,0 +1,544 @@
...
+    capture(.localizedCurreny(code: "USD"))
+}
+```
+
+
+## Alternatives considered
+
+### Extend `Sequence` instead of `Collection`
+
+All of the proposed algorithms are specific to the `Collection` protocol, without support for plain `Sequence`s. Types conforming to the `Sequence` protocol are not required to support multi-pass iteration, which makes a `Sequence` conformance insufficient for most of these algorithms. In light of this, the decision was made to have the underlying shared algorithm implementations work exclusively with `Collection`s.
```
milseman, line 524:
```suggestion
Most of the proposed algorithms are necessarily on `Collection` due to the use of indices or mutation.  `Sequence` does not support multi-pass iteration, so even `trimPrefix` would problematic on `Sequence` because it needs to look 1 `Element` ahead to know when to stop trimming.
```

Documentation/Evolution/StringProcessingAlgorithms.md
milseman: Just word smithing. You can put the hyperlink back in too.

milseman (approved):
Overall LGTM. I think some of the "backwards" future direction is salvageable and should be mentioned, but it doesn't need to be presented as a big deal.

After that, ship it!

itingliu:
> Overall LGTM. I think some of the "backwards" future direction is salvageable and should be mentioned, but it doesn't need to be presented as a big deal.
> 
> After that, ship it!

I reworked it a little bit and added it back. Thanks!

milseman:
@swift-ci please test

PR #189, opened 2022-02-25
Internalize some decls and some testing infrastructure

milseman:


milseman:
@swift-ci please test

Sources/_StringProcessing/Executor.swift
```diff
@@ -21,8 +20,8 @@ public struct Executor {
   }
 
   // FIXME: Public for prototype
```
rxwei, line 22:
```suggestion
```

rxwei (approved):
We should also be able to change `struct ReferenceID` to internal.

milseman:
Good point, I'll update with several more I found

milseman:
@swift-ci please test

PR #193, opened 2022-03-01
Implement String.WordView

Azoy:
This implements Unicode's defined word boundaries for a string found here: https://unicode.org/reports/tr29/#Word_Boundaries.

Here's a quick snippet:
```swift
for word in "Hello world!".words {
  print(word)
}
// Hello
// 
// world
// !
```

This is marked as WIP because I'm still investigating whether or not these word boundary semantics make sense right now.


natecook1000 (commented):
This is great, @Azoy! I think for matching purposes, the API we need is something like `isOnWordBoundary(_ i: String.Index) -> Bool`. It looks like that would be O(1) in the common case, with the worst case O(n) if called on a sequence of regional indicators. That suggests to me that we may want to include the regional indicator count, or an on/off toggle value, as part of the matching engine's state.

Out of curiosity, does the algorithm say which side of the boundary the "word" is on? Or can you have a word boundary in between two scalars that are both parts of words? (LMK if this question makes no sense.)

milseman (commented):
Progress is looking good. I think what would be really helpful are examples and tests showing the advantages of this approach over the level-1 naive thing of `\w` vs `\W`. If there's  value to this approach, even without localization info or large dictionaries, that's compelling.

Azoy:
@swift-ci please test Linux

PR #194, opened 2022-03-02
Improved delimiter lexing

hamishknight:
Refactor the delimiter lexing logic, and change the behavior such that:

- We now diagnose unprintable ASCII characters to match behavior with the C++ lexer.
- We now allow the C++ lexer to better recover from a missing closing delimiter.
- We now support `rx'...'` for experimental syntax.

Additionally, implement a heuristic that allows skipping over single quotes in `re'...'` literals if the preceding characters are `(?`, `(?(`, `\k`, `\g` or `(?C`.  These would not be valid literal endings anyway, and this allows us to support their single-quoted syntax.

This should be able to integrate without any C++ side changes, though I have some extra test cases I want to commit there whenever this gets integrated in.


hamishknight:
@swift-ci please test

Sources/_MatchingEngine/Regex/Parse/DelimiterLexing.swift
```diff
@@ -0,0 +1,330 @@
...
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+// TODO: mock up multi-line soon
+
+enum Delimiter: Hashable, CaseIterable {
+  case traditional
+  case experimental
+  case reSingleQuote
```
milseman, line 17: How about `rxSingleQuote` for experimental?

Sources/_MatchingEngine/Regex/Parse/DelimiterLexing.swift
```diff
@@ -0,0 +1,330 @@
...
+    switch kind {
+    case .endOfString: return "unterminated regex literal"
+    case .invalidUTF8: return "invalid UTF-8 found in source file"
+    case .unknownDelimiter: return "unknown regex literal delimiter"
+    case .unprintableASCII: return "unprintable ASCII character found in source file"
+    }
+  }
+}
+
+fileprivate struct DelimiterLexer {
```
milseman, line 68: In the end, we might want something that can feed `Source` (or even replace `Source`)

milseman (approved):

hamishknight:
@swift-ci please test

Sources/_MatchingEngine/Regex/Parse/DelimiterLexing.swift
hamishknight: Yeah would be nice to unify on an implementation that scans and produces unicode scalars for both

PR #195, opened 2022-03-02
Change script property default to use Script Extension

hamishknight:
Change the default script property behavior for an unqualified value e.g `\p{Greek}` from `\p{Script=Greek}` to `\p{Script_Extension=Greek}`. This is arguably the more intuitive behavior, and matches what Perl does.


hamishknight:
@swift-ci please test

milseman (approved):
Even if Perl didn't do it, I'd argue we should. The fact that Perl does it makes this unarguable, IMO.

Azoy (approved):
Thank you!

PR #199, opened 2022-03-04
Tweak a few generic signatures to work around a bug in the Requirement Machine

slavapestov:
Here is the reduced test case:

    func foo<C : Collection, R : Sequence>(_: C, _: R) where C.SubSequence == Substring, C.Element == R.Element {}

The Requirement Machine produces a valid but different minimization than the GenericSignatureBuilder for this signature, which flags the assert since right now we run both and compare the results:

    error: compile command failed due to signal 6 (use -v to see invocation)
    RequirementMachine generic signature minimization is broken:
    RequirementMachine says:      <C, R where C : Collection, R : Sequence, C.[Sequence]Element == R.[Sequence]Element, C.[Collection]SubSequence == Substring>
    GenericSignatureBuilder says: <C, R where C : Collection, R : Sequence, C.[Collection]SubSequence == Substring, R.[Sequence]Element == Character>

Radar tracking the fix: rdar://problem/89791117

### DIFF
--- a/Sources/_StringProcessing/Algorithms/Algorithms/Replace.swift
+++ b/Sources/_StringProcessing/Algorithms/Algorithms/Replace.swift
@@ -154,7 +154,7 @@ extension RangeReplaceableCollection where SubSequence == Substring {
     with replacement: Replacement,
     subrange: Range<Index>,
     maxReplacements: Int = .max
-  ) -> Self where Replacement.Element == Element {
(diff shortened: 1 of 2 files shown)


slavapestov:
@swift-ci Please test

stephentyrone:
@swift-ci test macOS

milseman (approved):
LGTM, please revert if/when fixed.

PR #200, opened 2022-03-04
Rename `RegexProtocol` to `RegexComponent` and `RegexBuilder` to `RegexComponentBuilder`.

rxwei:
We've decided to use `RegexComponent` for the protocol, and use `RegexComponentBuilder` for the builder type since we want the user to import a module named `RegexBuilder` to get the DSL.

### DIFF
--- a/Sources/Exercises/Participants/RegexParticipant.swift
+++ b/Sources/Exercises/Participants/RegexParticipant.swift
@@ -58,7 +58,7 @@ private func extractFromCaptures(
 }
 
 @inline(__always) // get rid of generic please
-private func graphemeBreakPropertyData<RP: RegexProtocol>(
+private func graphemeBreakPropertyData<RP: RegexComponent>(
   forLine line: String,
   using regex: RP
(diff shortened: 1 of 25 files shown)


milseman:
> With this change, should String.match(_:) take a Regex instead?

One option is an ABI that takes `Regex` and an inlinable API that's generic and just passes off the static var. (I've also been wondering where the ABI falls for operating between generics and existential). WDYT?

rxwei:
Sure.  My main concern was not about ABI, but whether it makes sense to have matching API on a `RegexComponent`.  As we sometimes want to use a custom regex component as a regex (to obtain its transformed `Match`), it seems like calling it "regex component" may not be entirely appropriate.

milseman:
@natecook1000 how are option scopes and global options coming along? That might help clarify the relationship between a `Regex`, `RegexProtocol`, `RegexComponent`, etc.

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
We converged on this solution in a meeting so I'll go ahead and merge. Any feedback is always welcome.

PR #201, opened 2022-03-08
Add regex builder DSL pitch and update captures pitch

rxwei:
- Add a new pitch for the regex builder DSL.
- Update the strongly typed regex captures pitch to the latest design.


rxwei:
@swift-ci please test

PR #202, opened 2022-03-08
WIP: Subpattern call support

rxwei:


milseman:
This seems like a good initial approach. I don't presently have all the rules regarding how backtracking and recursion-level references works in mind to comment further.

stephentyrone:
Should this be closed out?

PR #126, opened 2022-01-25
DSL quantifier types -> quantifier functions

rxwei:
Change quantifiers in the DSL from structure types to top-level functions.  This allows for a lot more flexibility with overloading a quantifier based on the input `Match` type.  The immediate benefit of this is getting rid of void and nested void types (see example below), and as a result eliminate the need for void-filtering within concatenation.  A more important benefit (IMO) is being able to get rid of nominal tuples and switch back to Swift tuples, as Swift tuples enable strongly typed named captures and eliminates the complexity that comes with nominal tuples.

This PR is based on #114.  Please review commit b14a85b.

-----

Before:
```swift
let r0 = OneOrMore(.digit) // => `.Match == Tuple2<Substring, [()]>`
let r1 = Optionally(.digit) // => `.Match == Tuple2<Substring, ()?>`
let r2 = OneOrMore(Repeat(Optionally(.digit))) // => `.Match == Tuple2<Substring, [[()?]]>`
"123".match(r2) // => `RegexMatch<Tuple2<Substring, [[()?]]>>?`
```
After:
```swift
let r0 = oneOrMore(.digit) // => `.Match == Substring`
let r1 = optionally(.digit) // => `.Match == Substring`
let r2 = oneOrMore(many(optionally(.digit))) // => `.Match == Substring`
"123".match(r2) // => `RegexMatch<Substring>`
```

-----

Before:
```swift
/(?<number>\d+)/ // => `Regex<Tuple2<Substring, Substring>>`
```

After:
```swift
/(?<number>\d+)/ // => `Regex<(Substring, number: Substring)>`
```

PR #203, opened 2022-03-10
DSL free functions -> types

rxwei:
Replace free functions such as `oneOrMore` with types such as `OneOrMore`. This hides overloads of free functions as initializers within those types. It also makes the DSL consistent with SwiftUI.

- `oneOrMore` -> `OneOrMore`
- `zeroOrMore` -> `ZeroOrMore`
- `optionally` -> `Optionally`
- `repeating` -> `Repeat`
- `choiceOf` -> `ChoiceOf`
- `capture` -> `Capture`
- `tryCapture` -> `TryCapture`

Note: The reason we didn't realize this was possible (e.g. in #126) was because we were narrowly focused on including the subpattern type in the quantifier/combinator's generic parameter, i.e. `OneOrMore<Component: RegexComponent>`, which made it impossible to deduce each type's `typealias Match` from `Component`. Now we have an unconstrained generic parameter (e.g. `OneOrMore<Match>`) which gives us the full flexibility to hide `Match` deduction rules in initializers' type signatures.

Result: All non-operator top-level functions have been eliminated.

### DIFF
--- a/Sources/_StringProcessing/RegexDSL/DSL.swift
+++ b/Sources/_StringProcessing/RegexDSL/DSL.swift
@@ -11,6 +11,18 @@
 
 import _MatchingEngine
 
+// A convenience protocol for builtin regex components that are initialized with
+// a `DSLTree` node.
+internal protocol _BuiltinRegexComponent: RegexComponent {
+  init(_ regex: Regex<Match>)
+}
+
+extension _BuiltinRegexComponent {
+  init(node: DSLTree.Node) {
+    self.init(Regex(node: node))
+  }
+}
+
 // MARK: - Primitives
 
 extension String: RegexComponent {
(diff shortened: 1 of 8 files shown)


rxwei:
@swift-ci please test

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -11,6 +11,19 @@
 
 import _MatchingEngine
 
+// A convenience protocol for builtin regex components that are initialized with
+// a `DSLTree` node.
+internal protocol _BuiltinRegexComponent: RegexComponent {
+  var regex: Regex<Match> { get }
```
milseman, line 17: BTW, do we have anything in-tree that uses this as an instance instead of a static var?

Sources/_StringProcessing/RegexDSL/DSL.swift
```diff
@@ -11,6 +11,19 @@
...
+  init(_ regex: Regex<Match>)
```
milseman, line 18: Do we want this defaulted with `self.regex = regex`? Or even just have this in an extension?

milseman, line 18: Do we want this defaulted with `self.regex = regex`?

Sources/_StringProcessing/RegexDSL/DSL.swift
milseman: Also, I don't think we need to redeclare this, do we?

milseman: Also, I don't think we need to redeclare this, do we?

milseman (approved):
Just when I was starting to appreciate the functions...

LGTM!

Sources/_StringProcessing/RegexDSL/DSL.swift
rxwei: Protocol initializers can only do delegate initialization, so I don’t think we can.  There’s no way to express a formal guarantee that a protocol property will be a stored property.

Sources/_StringProcessing/RegexDSL/DSL.swift
rxwei: I think custom regex types can use instance members to allow config options. We don’t have any examples in tree yet

rxwei: Good point, will remove redeclaration.

rxwei:
> Just when I was starting to appreciate the functions...

I had the same inertia myself…

rxwei:
@swift-ci please test

PR #205, opened 2022-03-11
Rename `RegexMatch` to `MatchResult`.

rxwei:
This is something we should've done earlier, as we've been referring to this type as `MatchResult` the whole time.  The only reason I named it `RegexMatch` was because the (now deleted) VM had a lower-level type named `MatchResult`.


rxwei:
@swift-ci please test

milseman (approved):
LGTM, also, what about renaming `Match` to `Captures`?

PR #206, opened 2022-03-12
Separate captures in DSLTree

milseman:
- Renamed `alternation` to `orderedChoice`
- Made a `capture` node and changed group to `nonCapturingGroup`
- Changed `groupTransform` into just a `transform` function, which currently is only nested inside a capture

Open to suggestions. Another option with transforms, if they are only ever accessible directly inside a capture, is a `captureTransform` node. I weakly prefer consistency with `matcher`.


rxwei (approved):

milseman:
@swift-ci please test

PR #207, opened 2022-03-13
Define primary associated type for `RegexComponent`.

rxwei:
This defines `RegexComponent<Match>` using the syntax pitched in https://forums.swift.org/t/pitch-2-light-weight-same-type-requirement-syntax/55081.


rxwei:
@swift-ci please test

natecook1000:
This is superseded by #454.

PR #210, opened 2022-03-17
Fix protocol definition in DSL pitch.

rxwei:


rxwei:
@swift-ci please test linux

PR #212, opened 2022-03-17
Add `Regex.Match.subscript(_: Reference)` to DSL pitch.

rxwei:
Given that the `Regex` type pitch is going to include the `Regex.Match` type, we add the extension that contains `subscript(_: Reference)` to the DSL pitch. Also fix the `Reference` code example.


rxwei:
@swift-ci please test Linux

PR #215, opened 2022-03-18
Clarify integration process.

rxwei:
- Require creating an branch from a commit on main instead of merging from main, as main is moving constantly.
- Require creating a dummy PR in apple/swift before we have full CI in this repo.


rxwei:
@swift-ci please test Linux

PR #218, opened 2022-03-18
Revert "Tweak a few generic signatures to work around a bug in the Requirement Machine"

slavapestov:
This reverts commit 695cf99be2e60c9a33304fefc559d35764d7c800.

PR #220, opened 2022-03-21
Merge from swift/main to pick up #218

rxwei:
This picks up #218.


rxwei:
@swift-ci please test

PR #222, opened 2022-03-21
Replace `DynamicCaptures` with `AnyRegexOutput`.

rxwei:
`AnyRegexOutput` is a collection type that represents a type-erased regex output with two use cases:
- When a `Regex` is initialized from a string, e.g. `Regex("(a)|(b)|c")`, it has type `Regex<AnyRegexOutput>`. One can iterate over the output to access output elements (the whole match followed by any captures).
- When working with an existing code base that uses dynamic `Regex` creation from strings, one can type-erase the match result of a strongly typed regex, e.g. `Regex<(Substring, Substring, Substring)>.Match`, and use it as a drop in replacement for the dynamic regex match.

This has been pitched as part of the [regex type](https://forums.swift.org/t/pitch-regex-type-and-overview/56029).

---

Example:

```swift
let regex = try! Regex(#"([0-9A-F]+)(?:\.\.([0-9A-F]+))?\s+;\s+(\w+).*"#)
let line = """
  A6F0..A6F1    ; Extend # Mn   [2] BAMUM COMBINING MARK KOQNDON..BAMUM \
  COMBINING MARK TUKWENTIS
  """
let match = line.firstMatch(of: regex)!
let output = match.output // `AnyRegexOutput`
output.0            // => (the entire match)
output[0].bounds    // => (the bounds of the entire match)
output[0].substring // => (the entire match)
output[1].substring // => "A6F0"
output[2].substring // => "A6F1"
output[3].substring // => "Extend"
```


rxwei:
Please review b5dd0ef as this is on top another PR.

rxwei:
@swift-ci please test

Sources/_StringProcessing/RegexDSL/AnyRegexOutput.swift
```diff
@@ -0,0 +1,128 @@
...
+public struct AnyRegexOutput {
+  let input: String
+  fileprivate let _elements: [ElementRepresentation]
+
+  /// The underlying representation of the element of a type-erased regex
+  /// output.
+  fileprivate struct ElementRepresentation {
+    /// The depth of `Optioals`s wrapping the underlying value. For example,
+    /// `Substring` has optional depth `0`, and `Int??` has optional depth `2`.
+    let optionalDepth: Int
```
milseman, line 38: Unrelated to this PR, did we decide on an optional nesting story?

milseman (approved):
LGTM, thanks for tackling this!

Sources/_StringProcessing/RegexDSL/AnyRegexOutput.swift
rxwei: There’s no non-factorial solution to optional flattening this year, unfortunately. So we’ll have nested optionals.

PR #226, opened 2022-03-21
Escape sequence + empty char class tweaks

hamishknight:
- Start throwing an error on unknown `a-zA-Z` and non-ASCII non-whitespace escape sequences.
- Start allowing more escape sequences that denote Unicode scalars in custom character class ranges.
- Forbid empty character classes, `]` as the first character is instead treated as literal, for consistency with PCRE, Oniguruma, and ICU.

### DIFF
--- a/Sources/_RegexParser/Regex/AST/Atom.swift
+++ b/Sources/_RegexParser/Regex/AST/Atom.swift
@@ -641,14 +641,64 @@ extension AST.Atom {
     case .scalar(let s):
       return Character(s)
 
-    case .keyboardControl, .keyboardMeta, .keyboardMetaControl:
-      // TODO: Not a character per-say, what should we do?
-      fallthrough
+    case .escaped(let c):
+      switch c {
+      // TODO: Should we separate these into a separate enum? Or move the
+      // specifics of the scalar to the DSL tree?
+      case .alarm:
+        return "\u{7}"
+      case .backspace:
+        return "\u{8}"
+      case .escape:
+        return "\u{1B}"
+      case .formfeed:
+        return "\u{C}"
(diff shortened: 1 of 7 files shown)


Sources/_MatchingEngine/Regex/AST/Atom.swift
```diff
@@ -641,17 +641,67 @@ extension AST.Atom {
     case .scalar(let s):
       return Character(s)
 
+    case .escaped(let c):
+      switch c {
+      // TODO: Should we separate these into a separate enum? Or move the
```
milseman, line 646: LGTM, but feel free to look first at the use sites of `literalCharacterValue` to see if there's really two different questions an AST client is asking. There might be a literal value for the purposes of character class treatment vs a literal value for the purposes of performing the match.

milseman (approved):
LGTM

Sources/_MatchingEngine/Regex/AST/Atom.swift
hamishknight: `literalCharacterValue` is currently only really used for building a model character class, and not for regular matching. It seems like we might want to use these scalars for both though, I actually wonder whether we should convert these escape sequences into a `.scalar` DSL atoms? Because ultimately it seems like these are just syntactic sugar for unicode scalar sequences, and it would allow them to be supported by the matching engine. Any thoughts?

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
Going to land this so I can land more stuff on top of it, I can work on converting to scalars in a follow-up

PR #227, opened 2022-03-24
RegexBuilder module

rxwei:
Move the regex builder DSL (except `RegexComponent`) to a new module named RegexBuilder. The DSL depends on `DSLTree` and a few other supporting types, so those types have been made `_spi(RegexBuilder) public`. The SPI establishes an ABI between `_StringProcessing` and `RegexBuilder`, but I don't think it's a concern because the two modules will co-evolve and both will be rebuilt for every release.


rxwei:
@swift-ci please test

rxwei:
I didn't move `CharacterClass` due to its dependency on `AST`. It might make sense to make it use `DSLTree` instead. I or someone else (@natecook1000?) can probably do it in a separate patch.

rxwei:
@swift-ci please test

natecook1000 (approved):
LGTM! I think we may want to talk a bit further about where things like `extension String: RegexComponent {}` should live, but this should be good for now. It's not ideal that `myString.firstMatch(of: "[0-9]+")` will work out of the box, and resolve to the regex method, even though it's evaluating as a string literal search.

I'll extricate `CharacterClass` in a follow-up with some API revisions.

Sources/RegexBuilder/DSL.swift
```diff
@@ -10,6 +10,16 @@
 //===----------------------------------------------------------------------===//
 
 import _MatchingEngine
+@_spi(RegexBuilder) import _StringProcessing
+import _StringProcessing
```
itingliu, line 14: Do we need this given we already have line 13?

itingliu (approved):

Sources/RegexBuilder/DSL.swift
rxwei: Good catch. Removed.

rxwei:
@swift-ci please test

PR #230, opened 2022-03-28
Remove operators from regex builder.

rxwei:
https://github.com/apple/swift-experimental-string-processing/blob/main/Documentation/Evolution/RegexBuilderDSL.md#operators-for-quantification-and-alternation


rxwei:
@swift-ci please test

PR #232, opened 2022-03-28
Rename _MatchingEngine to _RegexParser and fix CMake.

rxwei:
The _MatchingEngine module no longer contains the matching engine. It has the regex AST and parser. This patch renames it to `_RegexParserSupport`.

Also fix the CMake build which has been broken for a while.


rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

milseman (approved):
LGTM

PR #233, opened 2022-03-29
Change `lookahead` function to `Lookahead` type

rxwei:


rxwei:
@swift-ci please test

PR #234, opened 2022-03-29
Remove `Regex.init<R: RegexComponent>(_: R)`.

rxwei:
`Regex.init<R: RegexComponent>(_: R)` has existed as a way to convert any `RegexComponent` to a `Regex`. However, it is not super useful for the end user. More importantly, its existence seems to cause bad error messages in regex builder closures.

with-init

When this initializer is deleted, error messages become better and more localized. The user can still obtain a `Regex` from any `RegexComponent` just by calling the `regex` property.

without-init


rxwei:
@swift-ci please test

PR #236, opened 2022-03-31
Remove extra const from gestScriptExtensions

etcwilde:
Returning a constant pointer is extraneous and leads to a bunch of
warnings. Since you don't control where the pointer is assigned you
can't really control whether the pointer is const or not. The uint8_t
inside can be const though.


etcwilde:
@swift-ci please test

Azoy (approved):
Thank you!

PR #240, opened 2022-03-31
Remove Normalization and Grapheme data for SPI

Azoy:
In preparation of marking the relevant pieces in the stdlib as `@_spi`, remove the data here from `_StringProcessing`.


rxwei (approved):

Azoy:
@swift-ci please test Linux

PR #242, opened 2022-04-01
Multi-line and extended delimiter literal lexing

hamishknight:
Update the delimiter lexing logic to handle `/.../`, an extended variant that allows an arbitrary number of balanced pounds, e.g `##/.../##`, and a multi-line mode when the extended variant starts with a newline. The multi-line mode uses extended regex syntax by default, allowing for non-semantic whitespace and end-of-line comments.

Additionally, disable the `#|...|#`, `re'...'`, and `rx'...'` spellings for now.


hamishknight:
@swift-ci please test

hamishknight:
Given this patch allows the lexing of `/.../`, we can actually update https://github.com/apple/swift/pull/42119 to lex using that logic (though the heuristics remain on the compiler side for now).

milseman (approved):
LGTM

PR #246, opened 2022-04-04
Expanded options

natecook1000:
This adds methods to `RegexComponent` for the remainder of the regex options, and passes the current `MatchingOptions` further down into the consumers so that the correct behavior can be used.

### DIFF
--- a/Sources/_StringProcessing/Regex/Options.swift
+++ b/Sources/_StringProcessing/Regex/Options.swift
@@ -40,13 +40,26 @@ extension RegexComponent {
     wrapInOption(.asciiOnlyPOSIXProps, addingIf: useASCII)
   }
   
-  /// Returns a regular expression that uses a simpler set of Unicode scalar
-  /// values to detect word boundaries.
-  public func usingSimpleWordBoundaries(_ useASCII: Bool = true) -> Regex<Output> {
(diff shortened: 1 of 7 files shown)


Sources/_StringProcessing/Regex/Options.swift
```diff
@@ -12,17 +12,221 @@
...
+  /// This method corresponds to applying the `J` option in a regular
+  /// expression literal, and only applies to regular expressions specified as
+  /// literals.
+  ///
+  /// - Parameter allowDuplicateNames: A Boolean value indicating whether groups
+  ///   are allowed to share names.
+  public func allowDuplicateNames(_ allowDuplicateNames: Bool = true) -> Regex<Output> {
+    wrapInOption(.allowDuplicateGroupNames, addingIf: allowDuplicateNames)
+  }
+}
```
natecook1000, line 217: @milseman @hamishknight Am I right that we can't provide these at the API level? Would there be any value in providing them as deprecated or something so that we can explain the unavailability vs the other methods?

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #247, opened 2022-04-04
Clean up based on the String Processing Algorithms proposal

itingliu:
- Move functions and types that have not been proposed from public to internal
- Add doc comments for public API
- Add FIXME for API awaiting SE-0346
- Replace `_MatchResult` with `Regex<Output>.Match` and update tests

### DIFF
--- a/Sources/_StringProcessing/Algorithms/Matching/MatchReplace.swift
+++ b/Sources/_StringProcessing/Algorithms/Matching/MatchReplace.swift
@@ -12,7 +12,7 @@
 // MARK: `MatchingCollectionSearcher` algorithms
 
 extension RangeReplaceableCollection {
-  public func replacing<
+  func replacing<
     Searcher: MatchingCollectionSearcher, Replacement: Collection
   >(
     _ searcher: Searcher,
@@ -40,7 +40,7 @@ extension RangeReplaceableCollection {
     return result
   }
 
-  public func replacing<
+  func replacing<
     Searcher: MatchingCollectionSearcher, Replacement: Collection
   >(
     _ searcher: Searcher,
@@ -56,7 +56,7 @@ extension RangeReplaceableCollection {
       maxReplacements: maxReplacements)
   }
 
-  public mutating func replace<
+  mutating func replace<
     Searcher: MatchingCollectionSearcher, Replacement: Collection
   >(
     _ searcher: Searcher,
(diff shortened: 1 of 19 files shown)


Sources/_StringProcessing/Algorithms/Matching/MatchReplace.swift
```diff
@@ -110,4 +110,82 @@ extension RangeReplaceableCollection where SubSequence == Substring {
...
+  ///   - replacement: A closure that receives the full match information,
+  ///   including captures, and returns a replacement collection.
+  ///   - subrange: The range in the collection in which to search for `regex`.
+  ///   - maxReplacements: A number specifying how many occurrences of the
+  ///   sequence matching `regex` to replace. Default is `Int.max`.
+  /// - Returns: A new collection in which all occurrences of subsequence
+  /// matching `regex` are replaced by `replacement`.
+  public func replacing<R: RegexComponent, Replacement: Collection>(
+    _ regex: R,
+    with replacement: (Regex<R.Output>.Match) throws -> Replacement,
```
itingliu, line 127: @milseman this would replace `_MatchResult` boilerplate

rxwei (approved):

itingliu:
@swift-ci please test

Sources/_StringProcessing/Algorithms/Matching/MatchReplace.swift
milseman: That's great!

Tests/RegexBuilderTests/AlgorithmsTests.swift
```diff
@@ -38,13 +40,13 @@ class RegexConsumerTests: XCTestCase {
       int,
       input: "foo 160 bar 99 baz",
       result: "foo 240 bar 143 baz",
-      { match in String(match.result.1, radix: 8) })
+      { match in String(match.output.1, radix: 8) })
```
milseman, line 43: Just to make sure, couldn't this be `match.1`?, etc? No need to synthesize the entire output.

milseman (approved):
Overall LGTM, thanks for internalizing all of this.

Future work would be to see what we should remove, what we should underscore, etc., but getting this out of the module's API is a huge step. Thanks!

Tests/RegexBuilderTests/AlgorithmsTests.swift
itingliu: Yeah you're right that this can just be `match.1`. I was just doing a blind replace.

itingliu:
@swift-ci please test

itingliu:
@swift-ci Please test

hamishknight:
@swift-ci please test

PR #254, opened 2022-04-05
Move `CharacterClass` API into RegexBuilder

natecook1000:
This makes the existing CharacterClass model API internal to _StringProcessing, and creates a new (4th?) CharacterClass type in the RegexBuilder module. With this change, the CharacterClass APIs will only be available when that module's imported.

### DIFF
--- a/Sources/RegexBuilder/CharacterClass.swift
+++ b/Sources/RegexBuilder/CharacterClass.swift
@@ -177,8 +177,14 @@ extension Unicode.GeneralCategory {
 // MARK: - Set algebra methods
 
 extension RegexComponent where Self == CharacterClass {
-  public static func anyOf(_ first: CharacterClass, _ second: CharacterClass, _ rest: CharacterClass...) -> CharacterClass {
-    CharacterClass(.init(members: ([first, second] + rest).map { .custom($0.ccc) }))
+  public init(_ first: CharacterClass, _ rest: CharacterClass...) {
+    if rest.isEmpty {
+      self.init(first.ccc)
+    } else {
+      let members: [DSLTree.CustomCharacterClass.Member] =
+        (CollectionOfOne(first) + rest).map { .custom($0.ccc) }
+      self.init(.init(members: members))
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


natecook1000:
@swift-ci Please test

Sources/RegexBuilder/CharacterClass.swift
```diff
@@ -0,0 +1,209 @@
...
+
+  public static var word: CharacterClass {
+    .init(unconverted: .word)
+  }
+}
+
+extension RegexComponent where Self == CharacterClass {
+  /// Returns a character class that matches any character in the given string
+  /// or sequence.
+  public static func anyOf<S: Sequence>(_ s: S) -> CharacterClass
```
rxwei, line 96: Should the name align with `ChoiceOf`?  Could this be confusing with the `ChoiceOf` type, or perhaps be folded as an initializer under `ChoiceOf` so that the developer won't need to write `CharacterClass.anyOf`?

natecook1000: I had and removed a `ChoiceOf` initializer, since one thing that's distinct about character classes vs. other regex components is that character classes support inversion. Maybe just a `CharacterClass` initializer instead of this, since you often need the type name anyway?

rxwei: Good point... An initializer sounds good.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/RegexBuilder/CharacterClass.swift
```diff
@@ -0,0 +1,215 @@
...
+    guard let extendedCategory = category.extendedGeneralCategory else {
+      fatalError("Unexpected general category")
+    }
+    return CharacterClass(property:
+        .init(.generalCategory(extendedCategory), isInverted: false, isPOSIX: false))
+  }
+}
+
+/// Range syntax for characters in `CharacterClass`es.
+public func ...(lhs: Character, rhs: Character) -> CharacterClass {
```
rxwei, line 125: Would it be useful to define these operators as static methods on extensions?

natecook1000: I don't think(?) we get any benefit from making them static methods — that syntax is to support implementation of protocol requirements, which isn't the case here, and all operator functions end up in the global namespace in any case.

rxwei (approved):

PR #257, opened 2022-04-08
Unicode for String Processing proposal

natecook1000:
Draft of the Unicode for String Processing proposal


Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,677 @@
...
+Proposal: [SE-NNNN](NNNN-filename.md)  
+Authors: [Nate Cook](https://github.com/natecook1000), [Michael Ilseman](https://github.com/milseman)  
+Review Manager: TBD  
+Implementation: [apple/swift-experimental-string-processing][repo]  
+Status: **Draft**
+
+
+## Introduction
+
+This proposal describes `Regex`'s rich Unicode support during regular expression matching, along with the character classes and options and that define that behavior.
```
stephentyrone, line 12: typo: "and that define"

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,677 @@
...
+
+```swift
+let str = "Cafe\u{301}" // "Café"
+str == "Café"           // true
+str.dropLast()          // "Caf"
+str.last == "é"         // true (precomposed e with acute accent)
+str.last == "e\u{301}"  // true (e followed by composing acute accent)
+```
+
+At a regular expression's simplest, without metacharacters or special features, matching should behave like a test for equality. A string should always match a regular expression that simply contains the same characters.
```
stephentyrone, line 28: Throughout this section, you can drop "should". It's not that it should behave this way, it _does_ behave this way.

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,677 @@
...
+
+TK: semantic levels, options for controlling, canonical equivalence, Unicode properties
+
+## Detailed design
+
+First, we'll discuss the options that let you control a regex's behavior, and then explore the character classes that define the your pattern.
+
+### Options
+
+Options can be declared in two different ways: as part of [regular expression literal syntax][literals], or applied as methods when declaring a `Regex`. For example, both of these `Regex`es are declared with case insensitivity:
```
stephentyrone, line 68: I think Michael noted this on Slack already, but they're not part of the _literal_ syntax, they're part of the interior syntax (i.e. they work with string inits too).

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,677 @@
...
+	
+  /// Returns a regular expression that only matches ASCII characters when
+  /// matching character classes.
+  public func usingASCIICharacterClasses(_ useASCII: Bool = true) -> Regex<Output>
+}
+```
+
+#### Use Unicode word boundaries
+
+By default, matching uses the Unicode specification for finding word boundaries for the `\b` and `Anchor.wordBoundary` anchors. Disabling the `w` option switches to finding word boundaries at points in the input where `\b\B` or `\B\b` match, given the other matching options that are enabled, which may be more compatible with other regular expression engines.
```
stephentyrone, line 222: It's Unicode _level 2_, right?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,691 @@
...
+    <<This string
+    uses double-angle-brackets
+    to group text.>>
+    """
+    
+str.firstMatch(of: /<<.+>>/)        // nil
+str.firstMatch(of: /(?s)<<.+>>/)    // "This string\nuses double-angle-brackets\nto group text."
+```
+
+This option also affects the behavior of `CharacterClass.any`, which is designed to match the behavior of the `.` regex literal component.
```
milseman, line 141: I think we should conceptually think of this as being convenience API on top of more general purpose character class customizations.

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,691 @@
...
+```
+
+The `U` option toggles the "eagerness" of quanitifiers, so that quantifiers are reluctant by default, and only become eager when `?` is added to the quantifier.
+
+```swift
+// '(?U)' toggles the eagerness of quantifiers:
+str.firstMatch(of: /(?U)<.+>/)      // "<token>"
+str.firstMatch(of: /(?U)<.+?>/)     // "<token>A value.</token>"
+```
+
```
milseman, line 176: Should we consider a "possessive-by-default", either now or as future API?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,691 @@
...
+**Regular expression syntax:** `(?U)...` or `(?U...)`
+
+**`RegexBuilder` API:**
+
+```swift
+extension RegexComponent {
+  /// Returns a regular expression where quantifiers are reluctant by default
+  /// instead of eager.
+  public func reluctantCaptures(_ useReluctantCaptures: Bool = true) -> Regex<Output>
```
milseman, line 185: What does capturing have to do with anything?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,691 @@
...
+// Prints:
+//  ̀
+//  ́
+//  ̂
+// ...
+```
+
+`Regex` also provides ways to select a specific level of "any" matching, without needing to change semantic levels.
+
+- The **any grapheme cluster** character class is written as `\X` or `CharacterClass.anyGraphemeCluster`, and matches from the current location up to the next grapheme cluster boundary.
```
milseman, line 393: Including newlines?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,691 @@
...
+
+**`RegexBuilder` API:**
+
+```swift
+extension RegexComponent {
+    /// Returns a regular expression that ignores casing when matching.
+    public func ignoringCase(_ ignoreCase: Bool = true) -> Regex<Output>
+}
+```
+
```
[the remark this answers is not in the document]
natecook1000: These are defined on `RegexComponent` instead of `Regex`, so `ignoringCase()` is available on `String` whenever `String` conforms. Unfortunately, it looks like that conformance is in `_StringProcessing` instead of `RegexBuilder`! I'll move it over and explain the way it works in the proposal.

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: We could — that would probably necessitate figuring out some syntactical way of writing it, which could be a bit tricky. Maybe when possessive is the default then `.?+` is greedy? I can add it as another example to the future option directions.

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: ...I'm not sure what I was thinking here — will fix, thanks!

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: Yes, will note.

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,758 @@
...
+The options that `Regex` supports are shown in the table below. Options that affect _matching behavior_ are supported through both regex syntax and APIs, while options that have _structural_ or _syntactic_ effects are only supported through regex syntax.
+
+| **Matching Behavior**        |                |                               |
+|------------------------------|----------------|-------------------------------|
+| Case insensitivity           | `(?i)`         | `ignoringCase()`              |
+| Single-line mode             | `(?s)`         | `dotMatchesNewlines()`        |
+| Multi-line mode              | `(?m)`         | `anchorsMatchNewlines()`      |
+| ASCII-only character classes | `(?DSWP)`      | `usingASCIIDigits()`, etc     |
+| Unicode word boundaries      | `(?w)`         | `usingSimpleWordBoundaries()` |
```
stephentyrone, line 131: If we make this a binary toggle, we'll have to invent another switch if we ever want to support Unicode level 3. Should this take an enum instead? `.identifyingWordBoundaries(with: [.unicodeLevel[123], .defaultRule])` or similar?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,779 @@
...
+| `[:print:]`  | `[[:graph:][:blank:]--[:cntrl:]]` | starts-with          | `[[:graph:] ]`                |
+
+\* The Unicode scalar property definition for `[:word:]` is `[\p{Alphanumeric}\p{Mark}\p{Join_Control}\p{Connector_Punctuation}]`.  
+\*\* The Unicode scalar property definition for `[:cntrl:]` is `[^\p{Space}\p{Control}\p{Surrogate}\p{Unassigned}]`.
+
+#### Custom classes
+
+Custom classes function as the set union of their individual components, whether those parts are individual characters, individual Unicode scalar values, ranges, Unicode property classes or POSIX classes, or other custom classes.
+
+- Individual characters and scalars will be tested using the same behavior as if they were listed in an alternation. That is, a custom character class like `[abc]` is equivalent to `(a|b|c)` under the same options and modes.
```
milseman, line 620: I thought we said this was explicitly _not_ the case, so that you could support `e[\u{301}-\u{305}]` or some such, under the impression that alternation required a grapheme break. Has that changed?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,779 @@
...
+- When in grapheme cluster semantic mode, ranges of characters will test for membership using NFD form (or NFKD when performing caseless matching). This differs from how a `ClosedRange<Character>` would operate its `contains` method, since that depends on `String`'s `Comparable` conformance, but the decomposed comparison better aligns with the canonical equivalence matching used elsewhere in `Regex`.
```
milseman, line 621: How does that compose with concatenation and alternation?

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -0,0 +1,779 @@
...
+- A custom character class will match a maximum of one `Character` or `UnicodeScalar`, depending on the matching semantic level. This means that a custom character class with extended grapheme cluster members may not match anything while using scalar semantics.
```
milseman, line 622: Can you explain and provide rationale? This is a surprise and I thought we were working on clarifying the rules here.

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: Implemented this as `wordBoundaryKind(_:)` — see https://github.com/apple/swift-experimental-string-processing/pull/295

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: That's a good point, the `(a|b|c)` grouping would require a grapheme break, which shouldn't be the case. I'll update this language.

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: I'll clarify this language; this doesn't totally capture the intent.

Documentation/Evolution/UnicodeForStringProcessing.md
milseman: BTW, what's the reason alternation requires a grapheme break again?

natecook1000: Alternation start and end points are set by grouping (or the start/end of the whole pattern), which are the places we're requiring a grapheme break.

milseman: We surely need them for capturing groups and for the overall match, but even for non-capturing groups? The latter is just syntactic scopes. In the builder that's be an embedded `ChoiceOf { ... }`

natecook1000:
@swift-ci Please test

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -77,7 +77,38 @@ str.contains(/.+e\u{301}/)  // true
...
+
+Matching in Unicode scalar mode is analogous to comparing against a string's `UnicodeScalarView` — individual Unicode scalars are matched without combining them into characters or testing for canonical equivalence.
+
+```swift
+str.contains(/Café/.matchingSemantics(.unicodeScalar))
+// false - "e\u{301}" doesn't match with /é/
+str.contains(/Cafe\u{301}/.matchingSemantics(.unicodeScalar))
+// true - "e\u{301}" matches with /e\u{301}/
+```
+
```
milseman, line 110: Somewhere (not sure here or in detailed design) I feel we should have an example using the DSL where we switch modes of a sub-builder. That way we can explain where the implicit grapheme break requirements are inserted (e.g. on entry or on exit?)

Documentation/Evolution/UnicodeForStringProcessing.md
```diff
@@ -743,6 +815,10 @@ An earlier version of this pitch described adding standard library APIs to `Char
 
 A future `Regex` version could support a byte-level semantic mode in addition to grapheme cluster and Unicode scalar semantics. Byte-level semantics would allow matching individual bytes, potentially providing the capability of parsing string and non-string data together.
 
+### More general `CharacterSet` replacement
+
+Foundation's `CharacterSet` type is in some ways similar to the `CharacterClass` type defined in this proposal. `CharacterSet` is primarily a set type that is defined over Unicode scalars, and can therefore sometimes be awkward to use in conjunction with Swift `String`s. The proposed `CharacterClass` type is a `RegexBuilder`-specific type, and as such isn't intended to be a full general purpose replacement. Future work could involve expanding upon the `CharacterClass` API or introducing a different type to fill that role.
+
```
milseman, line 821: Is it in the regex builder module or the stdlib? If the stdlib I feel we'll want something a little more than saying it's regex-specific. It is a great model for character classes (or we should make it be that way), but we can say extra API or extra conversions are future work as we're focusing on regex's needs (which are broad and deep, so it's a good area to focus on)

milseman (approved):
Very quick pass, LGTM and we should try to move to pitch phase ASAP

Documentation/Evolution/UnicodeForStringProcessing.md
natecook1000: It's in the `RegexBuilder` – if/when we decide it's fully generally useful, we could lower to the stdlib.

natecook1000:
@swift-ci Please test

Documentation/Evolution/UnicodeForStringProcessing.md
milseman: We'd need to keep the ABI, so it's still pretty relevant (or just as relevant). I don't know if there's a good way to deprecate it in the future.

PR #261, opened 2022-04-09
Throwing customization hooks

milseman:
Adds test to try out throwing custom code. Adds processor support.


milseman:
@swift-ci please test

milseman:
I would like to add the following, but we don't take `throws` on `Capture` currently (should we?)

```

    struct Poison: Error, Hashable {}
    struct Overflow: Error, Hashable {}

    let addressRegex = Regex {
      "0x"
      Capture(OneOrMore(.hexDigit)) { hex -> Int in
        guard let i = Int(hex, radix: 16) else {
          throw Overflow()
        }
        if i == 0xdeadbeef {
          throw Poison()
        }
        return i
      }
    }
```

rxwei (approved):

rxwei:
> we don't take `throws` on `Capture` currently (should we?)

I don't know. Perhaps we should raise it on the forum?

milseman:
Added support for `throws` everywhere

milseman:
@rxwei it seems you beat me to the second commit by 4 minutes, drat. Care to take a look at the rest?

rxwei:
What's the proposed new behavior? Is it the following:
- Throwing means abort in both `Capture` and `TryCapture`.
- `TryCapture` backtracks when the capture transform returns nil.

milseman:
Yes, that’s right

rxwei:
LGTM

rxwei:
The non-Optional-returning ones on TryCapture could probably be removed.

milseman:
Done.

@swift-ci please test

PR #272, opened 2022-04-13
Allow POSIX character properties outside of custom character classes

hamishknight:
This matches the ICU behavior, and appears to be suggested by UTS#18.


hamishknight:
@swift-ci please test

PR #273, opened 2022-04-13
Throwing matches and update to CustomMatchingRegexComponent

itingliu:
- Update the name `CustomRegexComponent` to `CustomMatchingRegexComponent` per pitch
- Adopt `throws` for `CustomMatchingRegexComponent` as added in #261. Errors thrown by `CustomMatchingRegexComponent`'s conformers will be bubbled up to the engine and surfaced at client-side.

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -302,7 +302,9 @@ extension Compiler.ByteCodeGen {
     // not captured. This may mean we should store
     // an existential instead of a closure...
 
-    let matcher = builder.makeMatcherFunction(matcher)
+    let matcher = builder.makeMatcherFunction { input, start, range in
+      try matcher(input, start, range)
+    }
 
     let valReg = builder.makeValueRegister()
     builder.buildMatcher(matcher, into: valReg)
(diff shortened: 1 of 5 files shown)


itingliu:
Could anyone call out to @swift-ci for me?

rxwei:
@swift-ci please test

itingliu:
Tests are failing because I was still using the old function names before Michael's renaming cleanup

rxwei:
@swift-ci please test

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -293,13 +293,15 @@ extension Compiler.ByteCodeGen {
   mutating func emitMatcher(
     _ matcher: @escaping _MatcherInterface,
     into capture: CaptureRegister? = nil
-  ) {
+  ) throws {
```
rxwei, line 296: Hmm why does the compiler need to throw? It's only the underlying matcher that throws, right?

itingliu: Hmm you're right that it doesn't. I didn't understand how this works well so I thought it was needed. For my information why does the compiler need to throw in the case like `emitTransform` and `emitCharacter`?

rxwei: I think it's because they call `emitNode` which throws. `emitNode` throws because `emitReference` throws. `emitReference` throws because it needs to check the validity of references and throw `RegexCompilationError`.

itingliu:
@swift-ci please test

rxwei:
We need this in ASAP. Merging.

PR #278, opened 2022-04-14
Fix trivia in custom character classes

### LINKED ISSUE #275: Whitespace in CCCs in extended syntax causes trap when matching
The following traps in `Processor.cycle()`, due to the unescaped space inside the custom character class:

```swift
try! Regex(compiling: #"(?xx)[ \t]+"#).matchWhole(" \t ")
```

Removing the space (`[\t]`) or escaping it (`[\ \t]`) resolves the crash.

hamishknight:
Fix trivia matching logic in custom character classes, rather than matching and not advancing the input, we should always return `nil` to never match against the trivia. Resolves #275.

Additionally, fix custom character class parsing logic such that trivia is suitably ignored for set operations and the case where the first semantic member is `]`.

### DIFF
--- a/Tests/RegexTests/MatchTests.swift
+++ b/Tests/RegexTests/MatchTests.swift
@@ -1217,6 +1217,9 @@ extension RegexTests {
     firstMatchTest(#"(?xx)[ \t]+"#, input: " \t ", match: "\t")
     firstMatchTest(#"(?xx)[ \t]+"#, input: " \t\t ", match: "\t\t")
     firstMatchTest(#"(?xx)[ \t]+"#, input: " \t \t", match: "\t")
+
+    firstMatchTest("(?xx)[ a && ab ]+", input: " aaba ", match: "aa")
+    firstMatchTest("(?xx)[ ] a ]+", input: " a]]a ] ", match: "a]]a")
   }
   
   func testASCIIClasses() {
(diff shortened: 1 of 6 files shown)


hamishknight:
@swift-ci please test

Tests/RegexTests/MatchTests.swift
```diff
@@ -1208,6 +1208,19 @@ extension RegexTests {
...
+    firstMatchTest(#" \t "#, input: " \t ", match: " \t ")
+    firstMatchTest(#"(?xx) \t "#, input: " \t ", match: "\t")
+
+    firstMatchTest(#"[ \t]+"#, input: " \t ", match: " \t ")
+    firstMatchTest(#"(?xx)[ \t]+"#, input: " \t ", match: "\t")
+    firstMatchTest(#"(?xx)[ \t]+"#, input: " \t\t ", match: "\t\t")
+    firstMatchTest(#"(?xx)[ \t]+"#, input: " \t \t", match: "\t")
+
+    firstMatchTest("(?xx)[ a && ab ]+", input: " aaba ", match: "aa")
+    firstMatchTest("(?xx)[ ] a ]+", input: " a]]a ] ", match: "a]]a")
```
natecook1000, line 1222: 👍🏻

natecook1000 (approved):
LGTM!

PR #279, opened 2022-04-14
Move RegexComponent conformances to RegexBuilder

natecook1000:
This moves the RegexComponent conformances for String, Substring, etc to the RegexBuilder module. This will remove extra API from strings, such as the option-setting methods, when not writing DSL regexes, and also cut down on confusing usage of regex-oriented APIs with strings or characters.

### DIFF
--- a/Sources/RegexBuilder/DSL.swift
+++ b/Sources/RegexBuilder/DSL.swift
@@ -32,6 +32,40 @@ extension _BuiltinRegexComponent {
   }
 }
 
+// MARK: - Primitive regex components
+
+extension String: RegexComponent {
+  public typealias Output = Substring
+
+  public var regex: Regex<Output> {
+    .init(node: .quotedLiteral(self))
+  }
+}
+
+extension Substring: RegexComponent {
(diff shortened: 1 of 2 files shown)


natecook1000:
@swift-ci Please test

milseman (approved):
LGTM

Big improvement, though I do wonder if the associated type on RegexComponent should be named something like `RegexOutput`, since this type aliases on common types could clearly collide with something else. E.g. some kind of successor to `TextOutputStream` or some other parsing solution.

Otherwise we're waiting for the feature that allows us to use a different name to satisfy the requirement. I really want that feature anyways, so it might not be a big deal.

Regex's generic parameter can still remain `Output`. @rxwei, WDYT?

rxwei (approved):

rxwei:
> Regex's generic parameter can still remain `Output`. @rxwei, WDYT?

Good suggestion. I agree.

natecook1000:
Good call, Michael - going to merge this and follow up with that name change.

PR #280, opened 2022-04-14
Error on unknown character properties

hamishknight:
Previously we would form an `.other` character property kind for any unclassified properties, which crash at runtime as unsupported. Instead, switch to erroring on them. Eventually it would be nice if we could version this based on what the runtime being targeted supports.


hamishknight:
@swift-ci please test

natecook1000 (approved):
Agreed, this is a better approach! 👍🏻

hamishknight:
@swift-ci please test

PR #281, opened 2022-04-14
Rename RegexComponent.Output

natecook1000:
This renames `RegexComponent.Output` to `RegexOutput`, so that it is less likely to collide with other associated type that `String`, etc might conform to.


natecook1000:
@swift-ci Please test

rxwei (approved):

PR #283, opened 2022-04-15
Add remaining availability annotations.

rxwei:
Tests now have flag `-disable-availability-checking` because XCTests always test the just-built modules, rather than the modules in the OS.


rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

PR #286, opened 2022-04-15
Add `wholeMatch` and `prefixMatch`

itingliu:
Add the functions to string processing algorithms proposal and implement the
change.

Move the functions from `String` and `SubString` extensions to
`BidirectionalCollection`. Add tests for `firstMatch`, `wholeMatch`, and
`prefixMatch` that use a custom `BidirectionalCollection` type.

### DIFF
--- a/Documentation/Evolution/StringProcessingAlgorithms.md
+++ b/Documentation/Evolution/StringProcessingAlgorithms.md
@@ -162,8 +162,9 @@ We also propose the following regex-powered algorithms as well as their generic 
 |`replace(:with:subrange:maxReplacements)`| Replaces all occurrences of the sequence matching the given `RegexComponent` or sequence with a given collection |
 |`split(by:)`| Returns the longest possible subsequences of the collection around elements equal to the given separator |
 |`firstMatch(of:)`| Returns the first match of the specified `RegexComponent` within the collection |
+|`wholeMatch(of:)`| Matches the specified `RegexComponent` in the collection as a whole |
+|`prefixMatch(of:)`| Matches the specified `RegexComponent` against the collection at the beginning |
 |`matches(of:)`| Returns a collection containing all matches of the specified `RegexComponent` |
-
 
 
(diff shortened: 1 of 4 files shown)


milseman:
I agree that if these are on String/Substring, they should instead be generic like the others. I weakly hold that they should exist, but I appreciate @natecook1000's perspective, especially if we have anchoring modifier APIs on `Regex`. Let's discuss this at the next sync up.

itingliu:
@swift-ci please test

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -473,7 +484,7 @@ extension RangeReplaceableCollection where SubSequence == Substring {
     /// - Returns: A new collection in which all occurrences of subsequence
     /// matching `regex` in `subrange` are replaced by `replacement`.
     public func replacing<R: RegexComponent, Replacement: Collection>(
-        _ regex: R,
+        _ r: R,
```
milseman, line 487: Why this change? It makes the doc comment above invalid

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -159,32 +159,23 @@ extension Regex {
 }
 
 @available(SwiftStdlib 5.7, *)
-extension String {
+extension BidirectionalCollection where SubSequence == Substring {
+  /// Match a regex in its entirety.
+  /// - Parameter r: The regex to match against.
```
milseman, line 164: Hmm... potential parameter names could be `regex` or `pattern` or `regexComponent`....

WDYT?

milseman (approved):
LGTM, but potentially different parameter names for docs.

Sources/_StringProcessing/Regex/Match.swift
milseman: @natecook1000

PR #287, opened 2022-04-15
Import _RegexParser as implementation only

rxwei:
Use `@_implementationOnly import _RegexParser` in _StringProcessing and RegexBuilder modules.


rxwei:
@swift-ci please test

milseman (commented):
Thanks!

PR #289, opened 2022-04-15
Add Substring algorithms tests

natecook1000:
`firstMatch(of:)` was ignoring the start/endIndex when searching in substrings; this change fixes that issue. Also adds the 'in' label to `Regex.firstMatch(in:Substring)` to match the rest of the related APIs.

### DIFF
--- a/Sources/_StringProcessing/Regex/Match.swift
+++ b/Sources/_StringProcessing/Regex/Match.swift
@@ -123,7 +123,7 @@ extension Regex {
   /// Find the first match in a substring
   ///
   /// Returns `nil` if no match is found and throws on abort
-  public func firstMatch(_ s: Substring) throws -> Regex<Output>.Match? {
+  public func firstMatch(in s: Substring) throws -> Regex<Output>.Match? {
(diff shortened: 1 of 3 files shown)


natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -123,7 +123,7 @@ extension Regex {
   /// Find the first match in a substring
   ///
   /// Returns `nil` if no match is found and throws on abort
-  public func firstMatch(_ s: Substring) throws -> Regex<Output>.Match? {
+  public func firstMatch(in s: Substring) throws -> Regex<Output>.Match? {
```
itingliu, line 126: I thought `Output` was renamed to `RegexOutput`?

itingliu (approved):

Sources/_StringProcessing/Regex/Match.swift
rxwei: Regex's generic parameter is still named `Output`.

PR #292, opened 2022-04-16
Fix release build.

### LINKED ISSUE #288: Fails with release build configuration
The string processing package (SwiftPM) fails to build in release mode because the PEG prototype uses `@testable import _StringProcessing` in order to access the matching engine. It’s not urgent to fix it since we don't ever release using SwiftPM, but i think we should either move PEG to a test target or make the matching engine symbols be SPI for PEG to access.

rxwei:
Move Prototypes to a test target so that it's not built under the release configuration. Disable the PEG prototype in Exercises since it depends on Prototypes.

Resolves #288.


rxwei:
@swift-ci please test

milseman (commented):
Thanks!

PR #293, opened 2022-04-18
Make RegexBuilder quantifiers follow option settings

natecook1000:
This changes the RegexBuilder quantifiers to take an optional behavior, so that by default they can follow the current options setting. An explicitly provided behavior is always used.

This is how the regex syntax operator works:
```swift
let _ = /\w+/           // eager one-or-more operator
let _ = /\w+?/          // reluctant one-or-more operator
// (?U) toggles eager/reluctance
let _ = /(?U)\w+/       // reluctant one-or-more operator
let _ = /(?U)\w+?/      // eager one-or-more operator
```

In `RegexBuilder` regexes, we get support for `(?U)` via the `reluctantQuantifiers()` modifier method:

```swift
let _ = Regex {
    OneOrMore(.word)             // reluctant due to option method below
    OneOrMore(.word, .eagerly)   // eager due to explicit behavior parameter
}.reluctantQuantifiers()


natecook1000:
@swift-ci Please test

rxwei (approved):

PR #294, opened 2022-04-18
Add word boundary kind type

natecook1000:
This updated the word boundary algorithm selection API to accept a named type rather than a Bool, as described in the Unicode for String Processing proposal

PR #295, opened 2022-04-18
Nominalize option methods

natecook1000:
This subsumes #294, and renames the remaining option methods to use a nominal naming scheme instead of gerunds.


natecook1000:

rxwei (approved):

PR #299, opened 2022-04-19
Untangle `_RegexParser` from `RegexBuilder`

natecook1000:
With the change to make `_StringProcessing` import `_RegexParser` as implementation only, no `AST` types can be used by the `RegexBuilder` module. This adds wrapper types in `_StringProcessing` to hide those types, particularly as associated values in `DSLTree.Node`. It also provides a wrapper type for `DSLTree.Node` that conforms to `_TreeNode` for the purpose of building up the tree's capture structure. (It sure would be nice if implementation-only-imported protocols were just treated as internal.)


rxwei (approved):

Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -108,8 +108,8 @@ extension PrettyPrinter {
       print("/* TODO: conditional */")
 
     case let .quantification(amount, kind, child):
-      let amount = amount._patternBase
-      let kind = kind._patternBase
+      let amount = amount.ast._patternBase
+      let kind = (kind.ast ?? .eager)._patternBase
```
milseman, line 112: Does this depend on options?

Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -400,11 +400,6 @@ extension AST.Quantification.Kind {
 
 extension DSLTree.QuantificationKind {
   var _patternBase: String {
-    switch self {
-    case .explicit(let kind), .syntax(let kind):
-      return kind._patternBase
-    case .default:
-      return ".eager"
-    }
+    (ast ?? .eager)._patternBase
```
milseman, line 403: Does this depend on options?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -134,6 +142,12 @@ extension DSLTree {
       self.isInverted = isInverted
     }
     
+    public static func generalCategory(_ category: Unicode.GeneralCategory) -> Self {
```
milseman, line 145: Is this API? What proposal is it in?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -455,3 +452,225 @@ public struct CaptureTransform: Hashable, CustomStringConvertible {
...
+
+// MARK: AST wrapper types
+//
+// These wrapper types are required because even @_spi-marked public APIs can't
+// include symbols from implementation-only dependencies.
+
+extension DSLTree {
+  /// Presents a wrapped version of `DSLTree.Node` that can provide an internal
+  /// `_TreeNode` conformance.
+  struct _Tree: _TreeNode {
```
milseman, line 464: What is this used for? Why isn't this `DSLTree`?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -455,3 +452,225 @@ public struct CaptureTransform: Hashable, CustomStringConvertible {
...
+
+      case .customCharacterClass, .atom, .trivia, .empty,
+          .quotedLiteral, .consumer, .characterPredicate:
+        return .empty
+      }
+    }
+  }
+
+  @_spi(RegexBuilder)
+  public enum _AST {
```
milseman, line 554: Why is the AST relevant at all to DSLTree? Is it because we don't have or want to have versions of these enums?

Sources/_StringProcessing/_CharacterClassModel.swift
```diff
@@ -294,7 +298,17 @@ extension _CharacterClassModel: CustomStringConvertible {
 }
 
 extension _CharacterClassModel {
-  public func makeAST() -> AST.Node? {
+  public func makeDSLTreeCharacterClass() -> DSLTree.CustomCharacterClass? {
```
milseman, line 301: Why is this public? Should it be SPI?

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: This is more or less a compiler limitation workaround. `_TreeNode` is a protocol defined in `_RegexParser`, and since `DSLTree` is public (as SPI), its conformance to `_TreeNode` is also public. However, `_RegexParser` is imported as implementation only, so its symbols can't/shouldn't be included in the public interface of `_StringProcessing`. So this type basically provides that `DSLTree: _TreeNode` conformance by composition, and since it's internal, it doesn't cause the issue with exported implementation-only symbols.

Sources/_StringProcessing/PrintAsPattern.swift
natecook1000: It does in two ways:
- If this quantifier was written in the DSL, it can have a default value that gets drawn from the current options
- If this quantifier was written in syntax, `.eager` and `.reluctant` can get flipped, depending on the options (the AST is just the syntax)

We defer both of those effects until byte code generation — we'd need to maintain option state in a similar way here.

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: These are the things that we hadn't duplicated into the DSLTree layer, so we were just storing AST symbols in DSLTree nodes. My goal with this change was just to wrap those, not necessarily replace them wholesale. At some point it would be great to have better separation between the two modules.

Sources/_StringProcessing/PrintAsPattern.swift
natecook1000: Yes, see above.

Sources/_StringProcessing/_CharacterClassModel.swift
natecook1000: `_CharacterClassModel` is SPI, so this is as well — I should probably at least add a comment so it isn't jarring to see this in an extension.

PR #301, opened 2022-04-19
Don't parse a character property containing a backslash

hamishknight:
Add backslash to the list of characters we don't consider valid for a character property name (previous rules implemented in https://github.com/apple/swift-experimental-string-processing/pull/269). This means that we'll bail when attempting to lex a POSIX character property and instead lex a custom character class if we see a `\`. This allows e.g `[:\Q :] \E]` to be lexed as a custom character class. For `\p{...}` this just means we'll emit a truncated invalid property error, which is arguably more inline with what the user was expecting..

I noticed this when digging through the ICU source code. It will bail out of parsing a POSIX character property if it encounters one of its known escape sequences (e.g `\a`, `\e`, `\f`, ...). Interestingly this doesn't cover character property escapes e.g `\d`, but it's not clear that is intentional. Given backslash is not a valid character property character anyway, it seems reasonable to broaden this behavior to bail on any backslash.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
@@ -1176,6 +1176,14 @@ extension Source {
         // character property name anyway, and it's nice not to have diverging
         // logic for these cases.
         return true
+      case "\\":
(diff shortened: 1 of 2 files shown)


hamishknight:
@swift-ci please test

PR #305, opened 2022-04-19
Convenience quoting

milseman:


milseman:
@swift-ci please test

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -187,3 +187,10 @@ extension Substring {
     try? r.regex.prefixMatch(in: self)
   }
 }
+
+@available(SwiftStdlib 5.7, *)
+extension Regex {
+  public init(quoting string: String) {
```
rxwei, line 193: IMO this should go to Core.swift rather than Match.swift

PR #311, opened 2022-04-21
Allow setting any of the three quant behaviors

natecook1000:
This also moves `QuantificationBehavior` from the `RegexBuilder` module down to `_StringProcessing,` and renames it `RegexQuantificationBehavior`.

### DIFF
--- a/Sources/_RegexParser/Regex/AST/MatchingOptions.swift
+++ b/Sources/_RegexParser/Regex/AST/MatchingOptions.swift
@@ -41,7 +41,11 @@ extension AST {
       case graphemeClusterSemantics // X
       case unicodeScalarSemantics   // u
       case byteSemantics            // b
+      
+      // Swift-only default possessive quantifier
+      case possessiveByDefault      // t.b.d.
     }
+    
     public var kind: Kind
     public var location: SourceLocation
 
--- a/Sources/_StringProcessing/MatchingOptions.swift
+++ b/Sources/_StringProcessing/MatchingOptions.swift
@@ -63,6 +63,16 @@ extension MatchingOptions {
     stack.last!.contains(.reluctantByDefault)
   }
   
+  var defaultQuantificationKind: AST.Quantification.Kind {
+    if stack.last!.contains(.possessiveByDefault) {
+      return .possessive
+    } else if stack.last!.contains(.reluctantByDefault) {
+      return .reluctant
+    } else {
+      return .eager
+    }
+  }
+  
   var dotMatchesNewline: Bool {
     stack.last!.contains(.singleLine)
   }
@@ -150,6 +160,9 @@ extension MatchingOptions {
     case unicodeScalarSemantics
     case byteSemantics
     
+    // Swift-only default possessive quantifier
+    case possessiveByDefault
+
     init?(_ astKind: AST.MatchingOption.Kind) {
       switch astKind {
       case .caseInsensitive:
@@ -184,6 +197,8 @@ extension MatchingOptions {
         self = .unicodeScalarSemantics
       case .byteSemantics:
         self = .byteSemantics
+      case .possessiveByDefault:
+        self = .possessiveByDefault
         
       // Whitespace options are only relevant during parsing, not compilation.
       case .extended, .extraExtended:
@@ -218,6 +233,9 @@ extension MatchingOptions {
       }
       if Self.textSegmentOptions.contains(opt.representation) {
         remove(.textSegmentOptions)
+      }
+      if Self.quantificationBehaviors.contains(opt.representation) {
+        remove(.quantificationBehaviors)
       }
 
       insert(opt.representation)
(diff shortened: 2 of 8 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/_RegexParser/Regex/AST/MatchingOptions.swift
```diff
@@ -41,7 +41,11 @@ extension AST {
       case graphemeClusterSemantics // X
       case unicodeScalarSemantics   // u
       case byteSemantics            // b
+      
+      // Swift-only default possessive quantifier
+      case possessiveByDefault      // t.b.d.
```
milseman, line 46: In general, it's fine to model swift extensions (even ones that we haven't assigned letters to) in the options here. But, I'm curious if or why it's necessary.

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -63,6 +63,16 @@ extension MatchingOptions {
     stack.last!.contains(.reluctantByDefault)
   }
   
+  var defaultQuantificationKind: AST.Quantification.Kind {
+    if stack.last!.contains(.possessiveByDefault) {
```
milseman, line 67: Is it the stack's last entry or any entry in the stack? Or are we merging options as we push?

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -150,6 +160,9 @@ extension MatchingOptions {
     case unicodeScalarSemantics
     case byteSemantics
     
+    // Swift-only default possessive quantifier
+    case possessiveByDefault
```
milseman, line 164: Does this need to be modeled in this way? It seems like options would just have a quantification kind in it.

Sources/_StringProcessing/MatchingOptions.swift
```diff
@@ -274,7 +292,15 @@ extension MatchingOptions.Representation {
...
     [.graphemeClusterSemantics, .unicodeScalarSemantics, .byteSemantics]
   }
-    
+  
+  // Quantification behavior options
+  static var reluctantByDefault: Self { .init(.reluctantByDefault) }
+  static var possessiveByDefault: Self { .init(.possessiveByDefault) }
+
+  static var quantificationBehaviors: Self {
+    [.reluctantByDefault, .possessiveByDefault]
```
milseman, line 301: Couldn't someone switch the default back to eager via nesting?

Sources/_StringProcessing/Regex/Options.swift
```diff
@@ -183,6 +189,46 @@ public struct RegexWordBoundaryKind: Hashable {
   }
 }
 
+/// Specifies how much to attempt to match when using a quantifier.
+@available(SwiftStdlib 5.7, *)
+public struct RegexQuantificationBehavior: Hashable {
```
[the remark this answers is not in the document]
natecook1000: There's no unspecified case — `eager` is the default, and then you can switch to another. DSL types like `OneOf` take an optional behavior, with `nil` falling back to the behavior set via `quantificationBehavior`.

Sources/_StringProcessing/MatchingOptions.swift
natecook1000: Yes, I don't think I have a test for that…

Sources/_RegexParser/Regex/AST/MatchingOptions.swift
natecook1000: It's just more of our existing AST leakage. I think it's okay since we'll want it in the AST eventually, but we still do need to audit and remove AST values from non-AST types.

Sources/_StringProcessing/MatchingOptions.swift
natecook1000: We merge options on every push, so all of these just look at the top of the stack.

Sources/_StringProcessing/MatchingOptions.swift
natecook1000: This is a bitset internally, so it's modeling the tripartite state with two Booleans and an invariant check.

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Regex/Options.swift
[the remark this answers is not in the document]
natecook1000: The default isn't contextual, it's `.eager` — if someone wants to override that default for a chunk of their declaration they can, but there isn't an "unspecified" value for any of the options.

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Regex/Options.swift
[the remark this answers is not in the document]
milseman: And how would they have a non-overridable `.eager` subexpression within there?

PR #318, opened 2022-04-22
Remove redundant declarations from builder module

milseman:


milseman:
For some reason triggers an error in a test:

```
    let match = try XCTUnwrap("aaa".wholeMatch { OneOrMore { "a" } })
    error: Type '() -> OneOrMore<Substring>' cannot conform to 'RegexComponent'
```

PR #321, opened 2022-04-22
Preparation for location aware diagnostics in the compiler

rintaro:
Changes for https://github.com/apple/swift/pull/42583

Moving `libswiftLexRegexLiteral()` and `libswiftParseRegexLiteral()` to the compiler repository because these functions are basically just bridging the compiler to the actual lexing/parsing function.

 * Make some Lexing error APIs public.
 * Make LocatedErrorProtocol public and expose the `location` property
 * Shift the location of `LocatedError` in `parseWithDelimiters` so the client can get the valid string indices of the passed-in literal string.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/Parse.swift
+++ b/Sources/_RegexParser/Regex/Parse/Parse.swift
@@ -583,5 +583,15 @@ public func parseWithDelimiters<S: StringProtocol>(
   _ regex: S
 ) throws -> AST where S.SubSequence == Substring {
   let (contents, delim) = droppingRegexDelimiters(String(regex))
-  return try parse(contents, defaultSyntaxOptions(delim, contents: contents))
+  do {
+    return try parse(contents, defaultSyntaxOptions(delim, contents: contents))
+  } catch let error as LocatedErrorProtocol {
+    // Convert the range in 'contents' to the range in 'regex'.
+    let delimCount = delim.opening.count
+    let offsets = contents.offsets(of: error.location.range)
+    let startIndex = regex.index(atOffset: delimCount + offsets.lowerBound)
+    let endIndex = regex.index(atOffset: delimCount + offsets.upperBound)
+
+    throw error._typeErasedError.addingLocation(startIndex..<endIndex)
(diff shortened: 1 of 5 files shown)


rintaro:
@swift-ci please test

milseman:
Should this code live in the regex parser repo? @hamishknight

milseman (approved):
Overall LGTM but get a check from @hamishknight

milseman:
And make sure the cherry-picking aligns with the compiler-side, ideally avoiding lock-step or integration risk

Sources/_RegexParser/Regex/Parse/Parse.swift
```diff
@@ -583,5 +583,15 @@ public func parseWithDelimiters<S: StringProtocol>(
...
+    return try parse(contents, defaultSyntaxOptions(delim, contents: contents))
+  } catch let error as Source.LocatedError<ParseError> {
+    // Convert the range in 'contents' to the range in 'regex'.
+    let delimCount = delim.opening.count
+    let offsets = contents.offsets(of: error.location.range)
+    let startIndex = regex.index(atOffset: delimCount + offsets.lowerBound)
+    let endIndex = regex.index(atOffset: delimCount + offsets.upperBound)
+
+    throw error.error.addingLocation(startIndex..<endIndex)
+  }
```
hamishknight, line 596: Would it be worth generalizing this to any `LocatedErrorProtocol`? Just to make sure if a different kind of error is thrown, we don't silently use the wrong location? Doesn't have to be done in this PR tho.

hamishknight (approved):

Sources/_RegexParser/Regex/Parse/Parse.swift
rintaro: > Just to make sure if a different kind of error is thrown, we don't silently use the wrong location? 

Good catch! Although we currently don't use other underlying errors than `ParserError`, to make it safe, I generalized it in this PR.

rintaro:
@swift-ci Please test

PR #324, opened 2022-04-22
More updates for algorithms proposal

natecook1000:
This updates the Sequence/Collection constraints for several methods' parameters, adds missing parameters to the two `split` variations, and adopts opaque parameter types where allowed today.

### DIFF
--- a/Documentation/Evolution/StringProcessingAlgorithms.md
+++ b/Documentation/Evolution/StringProcessingAlgorithms.md
@@ -592,9 +592,23 @@ extension BidirectionalCollection where SubSequence == Substring {
 }
 ```
 
+**Note:** We plan to adopt the new generics features enabled by [SE-0346][] for these proposed methods when the standard library adopts primary associated types, [pending a forthcoming proposal][stdlib-pitch]. For example, the first method in the _Replacement_ section above would instead be:
 
+```swift
+extension RangeReplaceableCollection where Element: Equatable {
+    /// Returns a new collection in which all occurrences of a target sequence
+    /// are replaced by another collection.
+    public func replacing(
+        _ other: some Collection<Element>,
+        with replacement: some Collection<Element>,
+        subrange: Range<Index>,
+        maxReplacements: Int = .max
+    ) -> Self
+}
+```
 
-
+[SE-0346]: https://github.com/apple/swift-evolution/blob/main/proposals/0346-light-weight-same-type-syntax.md
+[stdlib-pitch]: https://forums.swift.org/t/pitch-primary-associated-types-in-the-standard-library/56426
 
 ## Alternatives considered
 
--- a/Documentation/Evolution/StringProcessingAlgorithms.md
+++ b/Documentation/Evolution/StringProcessingAlgorithms.md
@@ -234,7 +234,7 @@ extension Collection where Element: Equatable {
     /// - Parameter other: A sequence to search for within this collection.
     /// - Returns: `true` if the collection contains the specified sequence,
     /// otherwise `false`.
-    public func contains<S: Sequence>(_ other: S) -> Bool
+    public func contains<C: Collection>(_ other: C) -> Bool
         where S.Element == Element
 }
 
@@ -281,7 +281,7 @@ extension Collection where SubSequence == Self {
     /// - Parameter predicate: A closure that takes an element of the sequence
     /// as its argument and returns a Boolean value indicating whether the
     /// element should be removed from the collection.
-    public mutating func trimPrefix(while predicate: (Element) throws -> Bool)
+    public mutating func trimPrefix(while predicate: (Element) throws -> Bool) rethrows
 }
 
 extension RangeReplaceableCollection {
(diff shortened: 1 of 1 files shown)


Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -234,7 +234,7 @@ extension Collection where Element: Equatable {
     /// - Parameter other: A sequence to search for within this collection.
     /// - Returns: `true` if the collection contains the specified sequence,
     /// otherwise `false`.
-    public func contains<S: Sequence>(_ other: S) -> Bool
+    public func contains<C: Collection>(_ other: C) -> Bool
```
milseman, line 237: Would this be `some Collection<Element>`?

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -399,17 +399,17 @@ extension BidirectionalCollection where SubSequence == Substring {
...
     /// - Parameter r: The regex to match against.
     /// - Returns: The match if there is one, or `nil` if none.
-    public func wholeMatch<R: RegexComponent>(of r: R) -> Regex<R.Output>.Match? 
+    public func wholeMatch<R: RegexComponent>(of regex: R) -> Regex<R.RegexOutput>.Match? 
     
     /// Match part of the regex, starting at the beginning.
     /// - Parameter r: The regex to match against.
     /// - Returns: The match if there is one, or `nil` if none.
-    public func prefixMatch<R: RegexComponent>(of r: R) -> Regex<R.Output>.Match?
+    public func prefixMatch<R: RegexComponent>(of regex: R) -> Regex<R.RegexOutput>.Match?
```
milseman, line 412: Update the doc comment for `r` to be `regex`.

milseman (approved):
This LGTM, but want to make sure @itingliu can take a look.

Can you follow up with making sure the implementation matches?

Documentation/Evolution/StringProcessingAlgorithms.md
natecook1000: Right, all of the sequence/collection parameters can change to that. Where there's still a `<R: RegexComponent>` parameter, it's because we need the component's output parameter. Those will change to e.g. `func firstMatch<Output>(of regex: some RegexComponent<Output>) -> Regex<Output>.Match?` when we have the primary associated type.

milseman: Is the primary associate type tracked anywhere? Do we know when we need to integrate that?

itingliu: It's waiting on stdlib's adoption of lightweight generics. Filed #325 to track this

Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -574,27 +574,65 @@ extension RangeReplaceableCollection where SubSequence == Substring {
...
+    ///   - omittingEmptySubsequences: If `false`, an empty subsequence is 
+    ///     returned in the result for each consecutive pair of matches
+    ///     and for each match at the start or end of the collection. If
+    ///     `true`, only nonempty subsequences are returned.
     /// - Returns: A collection of substrings, split from this collection's
-    /// elements.
-    public func split<R: RegexComponent>(by separator: R) -> some Collection<Substring>
+    ///   elements.
+    public func split(
+        by separator: some RegexComponent,
```
itingliu, line 612: Do we want to drop `by` and call it 
```swift 
public func split(separator... 
```

That would match the existing `String.split`

natecook1000: I'm a little nervous about some of the overloads we already have in the stdlib — let me take a quick look.

natecook1000: Okay, this should be safe — when someone has a type-unconstrained function reference, they'll still get the current `split(separator:maxSplits:omittingEmptySubsequences:)` method, because it's more tightly specified (i.e., to `Character`, instead of a generic collection). I don't _think_ adopting the lightweight generics syntax will change the story here.

Documentation/Evolution/StringProcessingAlgorithms.md
natecook1000: And the `RegexComponent` primary associated type is here: https://github.com/apple/swift-experimental-string-processing/pull/207

itingliu (approved):

itingliu:
@swift-ci please test

Documentation/Evolution/StringProcessingAlgorithms.md
milseman: Even if this is from `_StringProcessing` instead of `Swift`?

natecook1000: I added tests to make sure you still get the stdlib versions here: https://github.com/apple/swift-experimental-string-processing/pull/330/files#diff-3d2dc8728c9cb47deca496d7122ae9cd2e49118aa79f65f3b3c730c1adc51078R139-R143

PR #330, opened 2022-04-23
Update algorithms methods to match proposal

natecook1000:
- [x] Sequence/Collection constraints
- [x] `@escaping`/`rethrows` closures
- [x] `split` parameters
- [x] More exhaustive testing
- [ ] ~Documentation comments~ _Coming in a separate PR_

### DIFF
--- a/Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift
+++ b/Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift
@@ -28,16 +28,16 @@ extension Collection where Element: Equatable {
   /// - Returns: `true` if the collection contains the specified sequence,
   /// otherwise `false`.
   @available(SwiftStdlib 5.7, *)
-  public func contains<S: Sequence>(_ other: S) -> Bool
-    where S.Element == Element
+  public func contains<C: Collection>(_ other: C) -> Bool
+    where C.Element == Element
   {
     firstRange(of: other) != nil
   }
 }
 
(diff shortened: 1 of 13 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift
```diff
@@ -46,6 +46,20 @@ extension BidirectionalCollection where Element: Comparable {
...
+  @available(SwiftStdlib 5.7, *)
+  public func contains(_ other: String) -> Bool {
+    firstRange(of: other) != nil
+  }
+  
+  @available(SwiftStdlib 5.7, *)
+  public func contains(_ other: Substring) -> Bool {
+    firstRange(of: other) != nil
+  }
+}
```
itingliu, line 61: Do we need to add these to the proposal?

itingliu (approved):
Looking great!

Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift
natecook1000: I don't think so - we just need to make sure these are winning the overload resolution vs Foundation's overlay version. However we do that is just an implementation detail.

PR #337, opened 2022-04-23
[5.7] Add ~= overloads

milseman:


milseman:
@swift-ci please test

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -184,3 +184,15 @@ extension Substring {
     try? r.regex.prefixMatch(in: self)
   }
 }
+
+@available(SwiftStdlib 5.7, *)
+public func ~=<Output>(regex: Regex<Output>, input: String) -> Bool {
```
rxwei, line 189: Should this take a generic `RegexComponent` instead? 

```swift
switch "aaa" {
case OneOrMore("a"):
  ...
}
```

rxwei: https://github.com/apple/swift-experimental-string-processing/pull/339

PR #339, opened 2022-04-23
Generic `~=` operator

rxwei:
Make `~=` generic over `RegexComponent` so that one can use builder components, e.g. `OneOrMore(...)`, or custom-consuming regex components as patterns.

### DIFF
--- a/Sources/_StringProcessing/Regex/Core.swift
+++ b/Sources/_StringProcessing/Regex/Core.swift
@@ -63,6 +63,13 @@ public struct Regex<Output>: RegexComponent {
 
 @available(SwiftStdlib 5.7, *)
 extension Regex {
+  public init(quoting string: String) {
+    self.init(node: .quotedLiteral(string))
+  }
+}
+
+@available(SwiftStdlib 5.7, *)
+extension Regex {
   /// A program representation that caches any lowered representation for
   /// execution.
   internal class Program {
--- a/Sources/_StringProcessing/Regex/Match.swift
+++ b/Sources/_StringProcessing/Regex/Match.swift
@@ -180,20 +180,12 @@ extension BidirectionalCollection where SubSequence == Substring {
 }
 
 @available(SwiftStdlib 5.7, *)
-extension Regex {
-  public init(quoting string: String) {
-    self.init(node: .quotedLiteral(string))
+extension RegexComponent {
+  public static func ~=(regex: Self, input: String) -> Bool {
+    input.wholeMatch(of: regex) != nil
   }
-}
 
-@available(SwiftStdlib 5.7, *)
-public func ~=<Output>(regex: Regex<Output>, input: String) -> Bool {
-  guard let _ = try? regex.wholeMatch(in: input) else { return false }
-  return true
-}
-
-@available(SwiftStdlib 5.7, *)
-public func ~=<Output>(regex: Regex<Output>, input: Substring) -> Bool {
-  guard let _ = try? regex.wholeMatch(in: input) else { return false }
-  return true
+  public static func ~=(regex: Self, input: Substring) -> Bool {
+    input.wholeMatch(of: regex) != nil
+  }
 }
--- a/Tests/RegexBuilderTests/AlgorithmsTests.swift
+++ b/Tests/RegexBuilderTests/AlgorithmsTests.swift
@@ -104,4 +104,61 @@ class RegexConsumerTests: XCTestCase {
        result: "9+16, 3, 10, 99+1")
     )
   }
+
+  func testSwitches() {
+    // Failure cases
+    do {
+      switch "abcde" {
+      case Regex {
+        "a"
+        ZeroOrMore(.any)
+        "f"
+      }:
+        XCTFail()
+
+      case "abc":
+        XCTFail()
+
+      case Regex {
+        "a"
+        "b"
+        "c"
+      }:
+        XCTFail()
+
+      default:
+        break
+      }
+    }
+    // Success cases
+    do {
+      let input = "abcde"
+
+      switch input {
+      case Regex {
+        "a"
+        ZeroOrMore(.any)
+        "e"
+      }:
+        break
+
+      default:
+        XCTFail()
+      }
+
+      guard case Regex({
+        "a"
+        ZeroOrMore(.any)
+        "e"
+      }) = input else {
+        XCTFail()
+        return
+      }
+
+      guard case OneOrMore(.word) = input else {
+        XCTFail()
+        return
+      }
+    }
+  }
 }


rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -180,20 +180,13 @@ extension BidirectionalCollection where SubSequence == Substring {
 }
 
 @available(SwiftStdlib 5.7, *)
-extension Regex {
```
rxwei, line 183: Moving this away from Match.swift as `Regex` initializers are declared in Core.swift.

Tests/RegexBuilderTests/AlgorithmsTests.swift
```diff
@@ -104,4 +104,38 @@ class RegexConsumerTests: XCTestCase {
...
+  func testSwitches() {
+    switch "abcde" {
+    case Regex {
+      "a"
+      ZeroOrMore(.any)
+      "f"
+    }:
+      XCTFail()
+
+    case "abc":
```
natecook1000, line 117: Is there a way to test that this is hitting the `Equatable` version instead of the `RegexComponent` version?

Tests/RegexBuilderTests/AlgorithmsTests.swift
```diff
@@ -104,4 +104,38 @@ class RegexConsumerTests: XCTestCase {
...
+
+    case Regex {
+      "a"
+      ZeroOrMore(.any)
+      "e"
+    }:
+      break // success
+
+    case OneOrMore(.word):
+      break // success
```
natecook1000, line 135: Should these be in two different tests? If this is  just to test compilation, no worries.

Tests/RegexBuilderTests/AlgorithmsTests.swift
rxwei: I don't think so. Any suggestions?

Tests/RegexBuilderTests/AlgorithmsTests.swift
rxwei: Good point. Fixed.

rxwei:
@swift-ci please test

Tests/RegexBuilderTests/AlgorithmsTests.swift
natecook1000: I don't have an idea that doesn't require making both sides generic, which I don't think we want to do. :/ We can check temporarily by printing / breaking in the operator function, but there's not really a way to test it properly.

natecook1000 (approved):

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -180,20 +180,13 @@ extension BidirectionalCollection where SubSequence == Substring {
...
+@_disfavoredOverload // disambiguate from equatable patterns
+public func ~=<R: RegexComponent>(regex: R, input: String) -> Bool {
+  input.wholeMatch(of: regex) != nil
 }
 
 @available(SwiftStdlib 5.7, *)
-public func ~=<Output>(regex: Regex<Output>, input: Substring) -> Bool {
-  guard let _ = try? regex.wholeMatch(in: input) else { return false }
-  return true
+@_disfavoredOverload // disambiguate from equatable patterns
```
xedin, line 189: I think it would make sense to make them both static func in `extension RegexComponent`. Why did you add `@_disfavoredOverload`?

rxwei: When there's no `@_disfavoredOverload`, some tests failed as string patterns had two ambiguous overloads.  When I moved the operator inside `extension RegexComponent`, I no longer see those errors so that's good. Is that the effect of making an operator a static method?

Sources/_StringProcessing/Regex/Core.swift
```diff
@@ -61,6 +61,13 @@ public struct Regex<Output>: RegexComponent {
   }
 }
 
+@available(SwiftStdlib 5.7, *)
+extension Regex {
+  public init(quoting string: String) {
+    self.init(node: .quotedLiteral(string))
+  }
+}
+
```
milseman, line 70: I have a PR in flight soon that's adding doc comments, so we might be racing on this.

Sources/_StringProcessing/Regex/Match.swift
xedin: Yes, there is some logic we have to optimize that in the solver. If you know exactly what used to fail it would be great to reduce and add it to our test suite to make sure that doesn't regress in the future.

xedin: In general, if possible, it's preferred to add protocol related operators into extensions of protocols since it helps to group and filter overloads in the solver.

Tests/RegexBuilderTests/AlgorithmsTests.swift
```diff
@@ -104,4 +104,49 @@ class RegexConsumerTests: XCTestCase {
...
+
+      case Regex {
+        "a"
+        "b"
+        "c"
+      }:
+        XCTFail()
+
+      default:
+        return
```
milseman, line 130:
```suggestion
        break
```

milseman: Otherwise the other tests in this function won't run. Also, would be nice to have a successful test case somewhere for switches.

milseman (approved):

Sources/_StringProcessing/Regex/Match.swift
rxwei:
```swift
public func ~=<R: RegexComponent>(regex: R, input: String) -> Bool {
  input.wholeMatch(of: regex) != nil
}

public func ~=<R: RegexComponent>(regex: R, input: Substring) -> Bool {
  input.wholeMatch(of: regex) != nil
}

func testSwitches() {
  switch "abcde" {
  case "abc": return // error: Ambiguous use of operator '~='
  }
}
```

Tests/RegexBuilderTests/AlgorithmsTests.swift
rxwei: Fixed.

rxwei:
@swift-ci please test

milseman (approved):
LGTM

rxwei:
@swift-ci please test

Sources/_StringProcessing/Regex/Match.swift
xedin: Hm, tried adding this into Match.swift and instead of ambiguity I get `switch must be exhaustive` with 04/22 snapshot of 5.7 branch.

rxwei: Updated the example.  It's missing a `default: break`.

Sources/_StringProcessing/Regex/Core.swift
rxwei: sorry you are gonna have a conflict...

Sources/_StringProcessing/Regex/Match.swift
xedin: Added `default` and tried with 04/23 snapshot of trunk and the project built successfully, rebased the project with origin/main just in case too.

PR #349, opened 2022-04-25
API stubs for casting and named captures

milseman:


Sources/_StringProcessing/Regex/AnyRegexOutput.swift
```diff
@@ -205,3 +205,28 @@ extension Regex.Match where Output == AnyRegexOutput {
     fatalError("FIXME: Not implemented")
   }
 }
+
+@available(SwiftStdlib 5.7, *)
+extension Regex where Output == AnyRegexOutput {
+  /// Returns whether a named-capture with `name` exists
+  public func contains(namedCapture name: String) -> Bool {
```
rxwei, line 212: Would `contains(captureNamed: "foo")` be more fluent?
```suggestion
  public func contains(captureNamed name: String) -> Bool {
```

milseman: Argh, I like both!

milseman: `myRegex.contains(captureNamed: "account")` vs `myRegex.contains(namedCapture: "account")`. I ever so slightly prefer the latter but I'm not good with names

rxwei: `namedCapture` reads as if the argument is a named capture, since nominal phrases, when used as an argument name, tend to describe the exact role of the argument.

milseman: Isn't that how we want it to read?

milseman: Is the question whether `namedCapture` means the capture or the name of the capture? I was reading it "the regex contains the named capture account".

rxwei: > Isn't that how we want it to read?

The argument is the name of a capture, not a named capture.  So "named capture" is IMO not accurately describing the role of the argument.

milseman: SGTM

milseman:
@swift-ci please test

PR #351, opened 2022-04-26
Algorithm cleanup

milseman:


Documentation/Evolution/StringProcessingAlgorithms.md
```diff
@@ -285,6 +308,8 @@ extension BidirectionalCollection where SubSequence == Substring {
 
 #### Trim prefix
 
+We propose generic `trimPrefix` for collections taking a `Element` predicate or a specific `Element`.
```
natecook1000, line 311:
```suggestion
We propose generic `trimmingPrefix` and `trimPrefix` methods for collections that trim elements matching a predicate or a possible prefix sequence.
```

natecook1000 (commented):
Looks good — as far as I can tell these hit all our edits. One note, and then I'll follow up later with info about empty matches.

milseman:
@swift-ci please test

itingliu (approved):
Looks good. I like separating regexcomponent-backed ones from the collection ones. Also I'm glad that we can land on `Collection` instead of `BidirectionalCollection`

PR #352, opened 2022-04-26
Add a section describing 'find empty' behavior

natecook1000:
We can do a quick review here, and then I can add this to https://github.com/apple/swift-evolution/pull/1635.


milseman:
@swift-ci please test

PR #353, opened 2022-04-26
Fix empty.split w/ empty separator

natecook1000:
This change corrects the behavior here:

```swift
"".split(separator: "", omittingEmptySubsequences: false)
// [""]  --  was ["", ""]

"".split(separator: "")
// []
```


milseman (approved):
LGTM

PR #359, opened 2022-04-26
Update PatternConverter

Azoy:
This updates the pattern converter to generate DSLs that are updated with the current proposed (and implemented) API. There are some character classes that are "unprintable" which kind of complicate the printing process.

```swift
// For example: '[\d[:upper:]abc](a|b)'

Regex {
  #/\d[:upper:]abc]/#
  Capture {
    ChoiceOf {
      "a"
      "b"
    }
  }
}
```

That being said, character classes that we support print pretty well!

```swift
// For example: [a-z--[^a-c]]

Regex {
  ("a"..."z")
    .subtraction(("a"..."c").inverted)
}
```

### DIFF
--- a/Sources/_StringProcessing/PrintAsPattern.swift
+++ b/Sources/_StringProcessing/PrintAsPattern.swift
@@ -57,7 +57,17 @@ extension PrettyPrinter {
   mutating func printAsPattern(_ ast: AST) {
     // TODO: Handle global options...
     let node = ast.root.dslTreeNode
-    printAsPattern(convertedFromAST: node)
+    
+    // If we have any named captures, create referencs to those above the regex.
+    let namedCaptures = node.getNamedCaptures()
+    
+    for namedCapture in namedCaptures {
+      print("let \(namedCapture) = Reference(Substring.self)")
+    }
+    
+    printBlock("Regex") { printer in
+      printer.printAsPattern(convertedFromAST: node)
+    }
   }
 
   // FIXME: Use of back-offs like height and depth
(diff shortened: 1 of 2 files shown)


Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -180,11 +190,118 @@ extension PrettyPrinter {
...
+      if terminateLine {
+        self.terminateLine()
+      }
+      
+      return
+    }
+    
+    var charMembers = ""
+    
+    let nonCharMembers = ccc.members.filter {
```
milseman, line 229: What is this doing or for? Add a comment, and consider if it's worth giving the concept a name e.g. a computed var

milseman (approved):
Quick look over and LGTM

Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -57,7 +57,17 @@ extension PrettyPrinter {
   mutating func printAsPattern(_ ast: AST) {
     // TODO: Handle global options...
     let node = ast.root.dslTreeNode
-    printAsPattern(convertedFromAST: node)
+    
+    // If we have any named captures, create referencs to those above the regex.
```
DaveEwing, line 61: typo "referencs"

Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -109,41 +122,38 @@ extension PrettyPrinter {
...
+      }
+      
+      if let pattern = a._patternBase(&self) {
+        print(pattern)
       }
 
     case .trivia:
-      // What should we do? Maybe keep comments, etc?
-      print("")
+      // We never print trivia
```
DaveEwing, line 155: Why not print them as a comment (since that's what they are)?

DaveEwing (approved):

PR #360, opened 2022-04-27
Simplify capture representations

milseman:
Make `CaptureStructure` semi-vestigial and introduce an array-of-structs `CaptureList` to use instead. Flipping from singly to multiply-nested captures is just flipping a `Bool`, though we still need to feed that info into the existentially stored captures.

### DIFF
--- a/Sources/PatternConverter/PatternConverter.swift
+++ b/Sources/PatternConverter/PatternConverter.swift
@@ -30,9 +30,6 @@ struct PatternConverter: ParsableCommand {
   @Flag(help: "Whether to show canonical regex literal")
   var showCanonical: Bool = false
 
-  @Flag(help: "Whether to show capture structure")
-  var showCaptureStructure: Bool = false
-
   @Flag(help: "Whether to skip result builder DSL")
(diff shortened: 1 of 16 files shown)


hamishknight:
I believe those `libswift...` functions can be removed now that they've been moved into the compiler repo (cc @rintaro)

milseman:
But what does that mean WRT the interface? We really wanted to keep this logic in the library because it's super easy to break otherwise. Case in point, I think this PR would break it because it makes `CaptureStructure` internal

milseman:
At the very least we need a comment convention for what is a compiler-locked API and what isn't. We could do that via `_spi` actually, which would be nice.

hamishknight:
Yeah I think it makes sense to add back complier entry points, which should be the only thing the compiler calls into (and yeah we could make them SPI). We can keep the bridging gunk on the compiler side tho.

milseman:
Do you think you can commit those changes to this PR? I'm not the most up to date.

hamishknight (approved):

milseman:
@swift-ci please test

PR #361, opened 2022-04-27
Add `compactingCaptureN` methods

natecook1000:
This adds a set of `compactingCaptureN` methods, which let you collapse a double-optional at a specific position in a regex capture list to a single optional. This position-specific approach only requires `arity^2` overloads instead of the `arity!` that a general solution would require.

```swift
let regex = Regex {
    ":"
    ChoiceOf {
        Capture {
            OneOrMore(.word)
        }
        Capture {
            OneOrMore(.digit)
        } transform: { Int($0) }
    }
    ":"
}
// type(of: regex) == Regex<(Substring, Substring?, Int??)>
let flattenedRegex = regex.compactingCapture2()
// type(of: flattenedRegex) == Regex<(Substring, Substring?, Int?)>
```

Note that `compactingCapture2()` could alternatively be added to `ChoiceOf` instead of the outer regex.

### DIFF
--- a/Sources/RegexBuilder/Variadics.swift
+++ b/Sources/RegexBuilder/Variadics.swift
@@ -4172,6 +4172,421 @@ extension TryCapture {
   }
 }
 
+@available(SwiftStdlib 5.7, *)
+extension RegexComponent {
+  public func compactingCapture1<C1>()
+    -> Regex<(Substring, C1?)>
+    where RegexOutput == (Substring, C1??)
+  {
+    Regex(castingCaptures: self.regex, to: (Substring, C1?).self)
+  }
+}
+
+@available(SwiftStdlib 5.7, *)
+extension RegexComponent {
+  public func compactingCapture1<C1, C2>()
+    -> Regex<(Substring, C1?, C2)>
+    where RegexOutput == (Substring, C1??, C2)
+  {
+    Regex(castingCaptures: self.regex, to: (Substring, C1?, C2).self)
+  }
+
+  public func compactingCapture2<C1, C2>()
+    -> Regex<(Substring, C1, C2?)>
+    where RegexOutput == (Substring, C1, C2??)
+  {
+    Regex(castingCaptures: self.regex, to: (Substring, C1, C2?).self)
+  }
+}
+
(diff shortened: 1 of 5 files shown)


Sources/RegexBuilder/Variadics.swift
```diff
@@ -4172,6 +4172,421 @@ extension TryCapture {
   }
 }
 
+@available(SwiftStdlib 5.7, *)
+extension RegexComponent {
```
milseman, line 4176: Do we want this on every `RegexComponent`? What about an internal-ish protocol for constructions that add a layer of optionality, so that these don't appear on `String`, `Regex`, etc?

Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -783,4 +789,38 @@ struct VariadicsGenerator: ParsableCommand {
 
       """)
   }
+  
+  func emitCompactingCapture(arity: Int) {
+    output("""
+      \(defaultAvailableAttr)
```
milseman, line 795: Availability?

natecook1000: I thought that's what this was for — does `Variadics.swift` not have the correct availability annotations?

milseman: I am blind

Sources/_StringProcessing/Regex/Core.swift
```diff
@@ -55,6 +55,12 @@ public struct Regex<Output>: RegexComponent {
     // in libswiftParseRegexLiteral.
     self.init(ast: try! parseWithDelimiters(pattern))
   }
+  
+  @_spi(RegexBuilder)
+  public init<T>(castingCaptures regex: Regex<T>, to outputType: Output.Type = Output.self) {
```
rxwei, line 60: It's really about casting the "output", not just captures, right? Should it be called `castingOutput:`?

natecook1000:
Closing this as the current plan is to implement `mapOutput` instead.

PR #364, opened 2022-04-27
Introduce new compiler interface

hamishknight:
Add the SPI interface `swiftCompilerLexRegexLiteral` and `swiftCompilerParseRegexLiteral` for the Swift compiler to call into. This allows us to avoid depending on other library API on the compiler side, while letting it keep the bridging gunk.

While we're here, add an extra `String` return for the parsing function that could allow us to change the regex emission format in the future. This still needs to be plumbed through on the complier side though.


hamishknight:
@swift-ci please test

rintaro (approved):
lgtm

milseman (approved):
LGTM, thanks! Anything need to be done compiler-side for this?

hamishknight:
The compiler-side changes need to be applied with the next main -> swift/main merge. Unfortunately it seems a RegexBuilder change is currently blocking CI for that (https://ci.swift.org/job/swift-PR-Linux/1558). We could do possibly do a cherry-pick of this in the mean time though, as I don't believe it should conflict with anything.

PR #328, opened 2022-04-22
[Integration] main (06f40f6) -> swift/main

rintaro:
This should not require swift changes. But [89b80bf](https://github.com/apple/swift-experimental-string-processing/pull/328/commits/89b80bfe180ea3940ae9a35bb5765f69f448ec4e) in this PR is required by https://github.com/apple/swift/pull/42583

PR #365, opened 2022-04-27
main -> swift/main

milseman:


rxwei:
We shouldn't create a PR from `main` directly as main's HEAD is constantly moving, and any PR that merges into `main` will invalidate the tests here. Instead, we should create a branch off of `main` and start a PR from _that_ to `swift/main`, for example: https://github.com/apple/swift-experimental-string-processing/pull/328.  Detailed workflow: https://github.com/apple/swift-experimental-string-processing#integration-workflow

PR #367, opened 2022-04-28
Replace opaque results with generic parameters in Algorithms.swift

hamishknight:
Work around the compiler bug rdar://92459215 by using generic parameters instead of opaque result types for a couple of methods.


hamishknight:
@swift-ci please test

PR #371, opened 2022-05-01
Add tests for AnyRegexOutput

milseman:


milseman:
@swift-ci please test

Tests/RegexTests/AnyRegexOutputTests.swift
```diff
@@ -0,0 +1,120 @@
+
+import _StringProcessing
+import XCTest
+
```
rxwei, line 4: There are already some tests on `AnyRegexOutput` in `RegexDSLTests.testDynamicCaptures()`. I'd suggest moving those tests here now that we have a dedicated file.

milseman:
@swift-ci please test

PR #379, opened 2022-05-05
Implement semantic diagnostics

### LINKED ISSUE #116: AST post-processing

### LINKED ISSUE #264: Require unique names for named captures
According to regex101, PCRE2 requires a unique name for each capture, but the current implementation of `Regex.init(compiling:)` doesn't throw an error when there's duplicate names.

### LINKED ISSUE #312: Reject quantifiers on zero-width assertions
A repeated assertion, like `\b+` causes an infinite loop, since it checks the same position over and over without advancing in the input. We should reject regexes with this pattern and prevent `RegexBuilder` regexes from creating them.

### LINKED ISSUE #357: Parser should error out for unsupported runtime features
Right now we will error out during regex-compilation for unsupported features or combinations, but the parser should also surface these errors for swift-compilation time.

hamishknight:
Start emitting errors for unsupported constructs, and other semantic errors such as duplicate group names.

Once we start emitting bytecode for regex at compile time, these errors could potentially be subsumed into the bytecode generator. But for now, implement them as a separate pass.

Resolves #357
Resolves #264
Resolves #116
Resolves #312

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
+++ b/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
@@ -45,7 +45,6 @@ enum ParseError: Error, Hashable {
 
   case cannotReferToWholePattern
 
-  case notQuantifiable
   case quantifierRequiresOperand(String)
 
   case backtrackingDirectiveMustHaveName(String)
@@ -83,6 +82,8 @@ enum ParseError: Error, Hashable {
   case duplicateNamedCapture(String)
   case invalidCharacterClassRangeOperand
   case invalidQuantifierRange(Int, Int)
+  case invalidCharacterRange(from: Character, to: Character)
+  case notQuantifiable
 }
 
 extension IdentifierKind {
@@ -125,8 +126,6 @@ extension ParseError: CustomStringConvertible {
       return "invalid escape sequence '\\\(c)'"
     case .cannotReferToWholePattern:
       return "cannot refer to whole pattern here"
-    case .notQuantifiable:
-      return "expression is not quantifiable"
     case .quantifierRequiresOperand(let q):
       return "quantifier '\(q)' must appear after expression"
     case .backtrackingDirectiveMustHaveName(let b):
@@ -191,6 +190,10 @@ extension ParseError: CustomStringConvertible {
       return "group named '\(str)' already exists"
     case let .invalidQuantifierRange(lhs, rhs):
       return "range lower bound '\(lhs)' must be less than or equal to upper bound '\(rhs)'"
+    case let .invalidCharacterRange(from: lhs, to: rhs):
+      return "character '\(lhs)' must compare less than or equal to '\(rhs)'"
+    case .notQuantifiable:
+      return "expression is not quantifiable"
     }
   }
 }
--- a/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
+++ b/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
@@ -15,6 +15,8 @@ enum ParseError: Error, Hashable {
   // TODO: I wonder if it makes sense to store the string.
   // This can make equality weird.
 
+  // MARK: Syntactic Errors
+
   case numberOverflow(String)
   case expectedNumDigits(String, Int)
   case expectedNumber(String, kind: RadixKind)
@@ -55,7 +57,6 @@ enum ParseError: Error, Hashable {
   case cannotRemoveMatchingOptionsAfterCaret
 
   case expectedCustomCharacterClassMembers
-  case invalidCharacterClassRangeOperand
 
   case emptyProperty
   case unknownProperty(key: String?, value: String)
@@ -73,6 +74,15 @@ enum ParseError: Error, Hashable {
   case cannotRemoveExtendedSyntaxInMultilineMode
 
   case expectedCalloutArgument
+
+  // MARK: Semantic Errors
+
+  case unsupported(String)
+  case deprecatedUnicode(String)
+  case invalidReference(Int)
+  case duplicateNamedCapture(String)
+  case invalidCharacterClassRangeOperand
+  case invalidQuantifierRange(Int, Int)
 }
 
 extension IdentifierKind {
@@ -88,6 +98,7 @@ extension IdentifierKind {
 extension ParseError: CustomStringConvertible {
   var description: String {
     switch self {
+    // MARK: Syntactic Errors
     case let .numberOverflow(s):
       return "number overflow: \(s)"
     case let .expectedNumDigits(s, i):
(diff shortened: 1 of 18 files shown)


Sources/_RegexParser/Regex/Parse/Diagnostics.swift
```diff
@@ -73,6 +74,15 @@ enum ParseError: Error, Hashable {
...
   case expectedCalloutArgument
+
+  // MARK: Semantic Errors
+
+  case unsupported(String)
+  case deprecatedUnicode(String)
+  case invalidReference(Int)
+  case duplicateNamedCapture(String)
+  case invalidCharacterClassRangeOperand
+  case invalidQuantifierRange(Int, Int)
```
milseman, line 85: Same enum or separate enum? (I haven't thought about it)

Sources/_RegexParser/Regex/Parse/Parse.swift
```diff
@@ -588,14 +609,23 @@ fileprivate func defaultSyntaxOptions(
   }
 }
 
+@available(*, deprecated, renamed: "parseWithDelimiters(_:_:)")
```
milseman, line 612: Are these compiler interfaces?

milseman (approved):
LGTM, unless there's a break between compiler/library interfaces. We really need to keep those stable and migrate incrementally.

Sources/_RegexParser/Regex/Parse/Diagnostics.swift
hamishknight: I initially added them as a separate enum, but it seemed cleaner to do it this way as they share all the same logic as other parser errors for e.g printing and `catch` block handling. This for example means they can use the same testing logic as other parsing tests. We could split them out in the future, but for now at least I think this is the simplest way to go.

hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/Sema.swift
```diff
@@ -0,0 +1,396 @@
...
+        .notHorizontalWhitespace, .verticalTab, .notVerticalTab,
+        // '\N' needs to be emitted using 'emitAny'.
+        .notNewline:
+      throw error(.unsupported("'\\\(esc.character)'"), at: loc)
+
+    // Character classes.
+    case .decimalDigit, .notDecimalDigit, .whitespace, .notWhitespace,
+        .wordCharacter, .notWordCharacter, .graphemeCluster, .trueAnychar:
+      // TODO: What about scalar matching mode for .graphemeCluster? We
+      // currently crash at runtime.
```
natecook1000, line 195: The behavior in the comment is fixed elsewhere.

Sources/_RegexParser/Regex/Parse/Sema.swift
```diff
@@ -0,0 +1,396 @@
...
+      throw error(.unsupported("Unicode block property"), at: loc)
+    }
+  }
+
+  func validateEscaped(
+    _ esc: AST.Atom.EscapedBuiltin, at loc: SourceLocation
+  ) throws {
+    switch esc {
+    case .resetStartOfMatch, .singleDataUnit, .horizontalWhitespace,
+        .notHorizontalWhitespace, .verticalTab, .notVerticalTab,
```
natecook1000, line 186: Support for `horizontalWhitespace` and `notHorizontalWhitespace` is implemented in #384.

Sources/_RegexParser/Regex/Parse/Sema.swift
```diff
@@ -0,0 +1,396 @@
...
+      throw error(.unsupported("script run"), at: kind.location)
+
+    case .changeMatchingOptions(let opts):
+      try validateMatchingOptions(opts)
+    }
+    try validateNode(group.child)
+  }
+
+  func validateQuantification(_ quant: AST.Quantification) throws {
+    try validateNode(quant.child)
```
natecook1000, line 343: Is it possible to validate that the child isn't a zero-width assertion here? e.g. we want to reject `\b+`. Tracked in #312.

Sources/_RegexParser/Regex/Parse/Sema.swift
```diff
@@ -0,0 +1,396 @@
...
+      break
+
+    case .expandsOnNFC, .expandsOnNFD, .expandsOnNFKD, .expandsOnNFKC:
+      throw error(.deprecatedUnicode(prop.rawValue.quoted), at: loc)
+
+    case .bidiControl, .compositionExclusion, .emojiComponent,
+        .extendedPictographic, .graphemeLink, .hyphen, .otherAlphabetic,
+        .otherDefaultIgnorableCodePoint, .otherGraphemeExtended,
+        .otherIDContinue, .otherIDStart, .otherLowercase, .otherMath,
+        .otherUppercase, .prependedConcatenationMark:
```
natecook1000, line 157: For these `.other*` properties, we don't implement b/c they're included in e.g. `\p{isAlphabetic}`. Do you think we should redirect people to the corresponding properties?

Sources/_RegexParser/Regex/Parse/Sema.swift
```diff
@@ -0,0 +1,396 @@
...
+  }
+
+  func validateEscaped(
+    _ esc: AST.Atom.EscapedBuiltin, at loc: SourceLocation
+  ) throws {
+    switch esc {
+    case .resetStartOfMatch, .singleDataUnit, .horizontalWhitespace,
+        .notHorizontalWhitespace, .verticalTab, .notVerticalTab,
+        // '\N' needs to be emitted using 'emitAny'.
+        .notNewline:
```
natecook1000, line 188: Still to implement, but I'll fix up the validation then.

hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/Sema.swift
hamishknight: Makes sense, and I can take care of that in a follow-up

Sources/_RegexParser/Regex/Parse/Sema.swift
hamishknight: Implemented logic to check for escape sequences that aren't quantifiable, how does it look?

Sources/_RegexParser/Regex/Parse/Sema.swift
hamishknight: Went ahead and marked them supported

Sources/_RegexParser/Regex/Parse/Sema.swift
```diff
@@ -0,0 +1,396 @@
...
+        .unsupported("character class range operand"), at: lhs.location)
+    }
+
+    guard rhs.literalCharacterValue != nil else {
+      throw error(
+        .unsupported("character class range operand"), at: rhs.location)
+    }
+
+    // TODO: Validate lhs <= rhs? That may require knowledge of case
+    // insensitivity though.
```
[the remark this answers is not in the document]
hamishknight: Makes sense, updated to check range regardless of case sensitivity

natecook1000 (approved):
LGTM!

PR #383, opened 2022-05-06
Use full case folding outside of character classes

natecook1000:
This should get us part way to proper full case folding support. To handle case folding that results in multiple characters, we'll need to switch to more of a sequence-based comparison than just matching a single character.

### DIFF
--- a/Sources/_StringProcessing/Unicode/CaseConversion.swift
+++ b/Sources/_StringProcessing/Unicode/CaseConversion.swift
@@ -9,6 +9,24 @@
 //
 //===----------------------------------------------------------------------===//
 
+@_spi(_Unicode) import Swift
 
 // TODO
 
+extension Character {
+  /// Whether this character and `c` are equal when case folded.
+  func caseFoldedEquals(_ c: Character) -> Bool {
+    guard #available(SwiftStdlib 5.7, *) else { fatalError() }
+    let foldedSelf = unicodeScalars.map(\.properties._caseFolded).joined()
+    let foldedOther = c.unicodeScalars.map(\.properties._caseFolded).joined()
+    return foldedSelf == foldedOther
+  }
+}
+
+extension UnicodeScalar {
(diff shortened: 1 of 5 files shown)


Azoy (approved):
LGTM!

Sources/_StringProcessing/Unicode/CaseConversion.swift
```diff
@@ -9,6 +9,24 @@
...
 
 // TODO
 
+extension Character {
+  /// Whether this character and `c` are equal when case folded.
+  func caseFoldedEquals(_ c: Character) -> Bool {
+    guard #available(SwiftStdlib 5.7, *) else { fatalError() }
+    let foldedSelf = unicodeScalars.map(\.properties._caseFolded).joined()
+    let foldedOther = c.unicodeScalars.map(\.properties._caseFolded).joined()
+    return foldedSelf == foldedOther
```
milseman, line 22: This makes a lot of intermediary arrays, which is unfortunate because nearly all Characters fit in the small-string form, and those that don't are usually case invariant. Does `.lazy.map(...).elementsEqual(` work?

natecook1000: I think what we really want is a stdlib API that can do the case-folded comparison — it's wasteful even to UTF-8 encode these when we should be able to convert, canonicalize, and compare character-by-character.

milseman: We can write the code here or in the stdlib (I agree stdlib SPI makes sense). But in the meantime, can we write the code we want the stdlib to have, or at least approximate it by making the mapping lazy?

milseman: Otherwise, it looks like case-insensitive matching could do multiple array allocations for every single character in the input

natecook1000:
@swift-ci Please test

stephentyrone:
Nate, what's the status of this PR now?

PR #386, opened 2022-05-09
Introduce scalar sequences `\u{AA BB CC}`

hamishknight:
Allow a whitespace-separated list of scalars within the `\u{...}` syntax. This is syntactic sugar that gets implicitly splatted out, for example `\u{A B C}` becomes `\u{A}\u{B}\u{C}`.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
@@ -290,10 +290,54 @@ extension Source {
     return try Source.validateUnicodeScalar(str, .hex)
   }
 
+  /// Try to lex a seqence of hex digit unicode scalars.
+  ///
+  ///     UniScalarSequence   -> Whitespace? UniScalarSequencElt+
+  ///     UniScalarSequencElt -> HexDigit{1...} Whitespace?
+  ///
+  mutating func expectUnicodeScalarSequence(
+    eating ending: Character
+  ) throws -> AST.Atom.Kind {
+    try recordLoc { src in
+      var scalars = [AST.Atom.Scalar]()
+      var trivia = [AST.Trivia]()
+
+      // Eat up any leading whitespace.
+      if let t = src.lexWhitespace() { trivia.append(t) }
+
+      while true {
+        let str = src.lexUntil { src in
+          // Hit the ending, stop lexing.
+          if src.isEmpty || src.peek() == ending {
+            return true
+          }
+          // Eat up trailing whitespace, and stop lexing to record the scalar.
+          if let t = src.lexWhitespace() {
+            trivia.append(t)
+            return true
+          }
+          // Not the ending or trivia, must be a digit of the scalar.
+          return false
+        }
+        guard !str.value.isEmpty else { break }
+        scalars.append(try Source.validateUnicodeScalar(str, .hex))
+      }
+      guard !scalars.isEmpty else {
+        throw ParseError.expectedNumber("", kind: .hex)
+      }
+      try src.expect(ending)
+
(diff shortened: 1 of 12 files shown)


hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -289,49 +347,60 @@ extension Source {
...
       // Hex numbers.
-      case "u" where src.tryEat("{"), "x" where src.tryEat("{"):
-        let str = try src.lexUntil(eating: "}").value
-        return try Source.validateUnicodeScalar(str, .hex)
+      case "u" where src.tryEat("{"):
+        return try src.expectUnicodeScalarSequence(eating: "}")
+
+      case "x" where src.tryEat("{"):
+        let str = try src.lexUntil(eating: "}")
+        return .scalar(try Source.validateUnicodeScalar(str, .hex))
```
hamishknight, line 366: @milseman Just to double check, we don't want this to apply to `\x{...}`?

milseman: I don't think so, lets stick with `\u{xx xx xx}`

Tests/RegexTests/ParseTests.swift
```diff
@@ -600,6 +616,19 @@ extension RegexTests {
...
+      #"[\u{AA}-\u{BB}]"#,
+      charClass(range_m(scalar_a("\u{AA}"), scalar_a("\u{BB}")))
+    )
+    parseTest(
+      #"[\u{AA BB}-\u{CC}]"#,
+      charClass(range_m(scalarSeq_a("\u{AA}", "\u{BB}"), scalar_a("\u{CC}")))
+    )
+    parseTest(
+      #"[\u{CC}-\u{AA BB}]"#,
+      charClass(range_m(scalar_a("\u{CC}"), scalarSeq_a("\u{AA}", "\u{BB}")))
```
milseman, line 629: What does a scalar sequence in a range entail?

milseman: CC @natecook1000 on whether we want these inside custom ccs and whether this would do the whole NFD thing

milseman (approved):

Tests/RegexTests/ParseTests.swift
milseman: Let's error/reject

natecook1000: Yeah, this was my frustration from the last time we discussed these. It would be great if this syntax could be used for writing a multi-scalar character in a custom CC, but that's not how it's defined at all. I agree that we should error.

milseman: > Yeah, this was my frustration from the last time we discussed these. It would be great if this syntax could be used for writing a multi-scalar character in a custom CC, but that's not how it's defined at all. I agree that we should error.

Can we can define it that way? That makes _way_ more sense to me. But would it be a sequence again in scalar-semantics mode? Would we error out if it's more than one grapheme cluster normally?

natecook1000: I think we should reject altogether in CCs for now, so that we can discuss and decide on whether there's a reasonable behavior that works for different modes.

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

Tests/RegexTests/ParseTests.swift
hamishknight: Updated to reject as unsupported in a custom character class for now

PR #393, opened 2022-05-10
Tighten up some syntax rules

### LINKED ISSUE #303: Regex parser should be wary of combining characters
The regex parser is based over `Character`, which is fine, but means that some programs could put combining scalars following meta characters and those will not compare equal. We have a few options:

- We process scalars instead
- We error out for any multi-scalar grapheme cluster that starts with a metacharacter scalar

The latter seems simpler and there's an easy (and highly advisable!) fall back path of representing the combining scalar through an escape.

hamishknight:
- Ban numeric escapes in custom character classes e.g `[\7]`. In PCRE, this is an octal sequence, but we require a `0` prefix.
- Ban multiple-scalar non-letter non-digit ASCII characters, as they may be confusable with metacharacters.
- Reserve `<{...}>` for a future interpolation syntax.

Resolves #303

### DIFF
--- a/Tests/RegexTests/ParseTests.swift
+++ b/Tests/RegexTests/ParseTests.swift
@@ -466,14 +466,6 @@ extension RegexTests {
     parseTest(#"[\08]"#, charClass(scalar_m("\u{0}"), "8"))
     parseTest(#"[\0707]"#, charClass(scalar_m("\u{1C7}")))
 
-    // TODO: These are treated as octal sequences by PCRE, we should warn and
-    // suggest user prefix with 0.
-    parseTest(#"[\1]"#, charClass("1"))
-    parseTest(#"[\123]"#, charClass("1", "2", "3"))
-    parseTest(#"[\101]"#, charClass("1", "0", "1"))
-    parseTest(#"[\7777]"#, charClass("7", "7", "7", "7"))
-    parseTest(#"[\181]"#, charClass("1", "8", "1"))
-
     // We take *up to* the first two valid digits for \x. No valid digits is 0.
     parseTest(#"\x"#, scalar("\u{0}"))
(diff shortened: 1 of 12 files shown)


hamishknight:
@swift-ci please test

Tests/RegexTests/ParseTests.swift
```diff
@@ -2555,6 +2576,27 @@ extension RegexTests {
...
+
+    diagnosticTest("[\u{301}]", .confusableCharacter("[\u{301}"))
+    diagnosticTest("(\u{358})", .confusableCharacter("(\u{358}"))
+    diagnosticTest("{\u{35B}}", .confusableCharacter("{\u{35B}"))
+    diagnosticTest(#"\\#u{35C}"#, .confusableCharacter(#"\\#u{35C}"#))
+    diagnosticTest("^\u{35D}", .confusableCharacter("^\u{35D}"))
+    diagnosticTest("$\u{35E}", .confusableCharacter("$\u{35E}"))
+    diagnosticTest(".\u{35F}", .confusableCharacter(".\u{35F}"))
+    diagnosticTest("|\u{360}", .confusableCharacter("|\u{360}"))
+    diagnosticTest(" \u{361}", .confusableCharacter(" \u{361}"))
```
hamishknight, line 2589: @milseman Should we include whitespace characters for this diagnostic? Not really metacharacters, but also you'd probably want to write out explicitly with `\u{...}` anyway.

milseman: Combining characters with whitespace are pretty nasty and satisfy (one of) Unicode's definition of degenerate. So I'm good forcing them to use explicit scalar escapes too.

hamishknight:
@swift-ci please test

PR #396, opened 2022-05-10
Improve Regex's playground display.

rxwei:
Conform `Regex` to `CustomPlaygroundDisplayConvertible` to prevent its guts from being displayed in a playground.


rxwei:
@swift-ci please test

milseman (approved):
LGTM

PR #397, opened 2022-05-11
Disable resilience on _RegexParser

rxwei:
_RegexParser does not need resilience as it's only ever going to be used by _StringProcessing and RegexBuilder.


rxwei:
@swift-ci please test

milseman:
Do we want to get this in for 5.7?

rxwei:
Yes. Haven’t got a chance yet

PR #398, opened 2022-05-11
[5.7] Tidy up build flags and fix implicit import circular dependency (#392)

rxwei:
Cherry-pick of #392

---

- Explicitly ask the compiler not to implicitly import _StringProessing. This is to avoid a circular dependency when `-enable-experimental-string-processing` is enabled by default.
- Unify the build flags for modules that are built in the compiler repo into a `stdlibSettings` value.
- Disable implicit _Concurrency import as well since it is how it's built in the compiler repo. This helps us catch errors before we integrate with the compiler repo.
- Remove `-enable-experimental-pairwise-build-block` since SE-0348 has been implemented and enabled.
- Update the minimum toolchain requirement to 2022-04-20.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -10,6 +10,13 @@ let availabilityDefinition = PackageDescription.SwiftSetting.unsafeFlags([
     #"SwiftStdlib 5.7:macOS 9999, iOS 9999, watchOS 9999, tvOS 9999"#,
 ])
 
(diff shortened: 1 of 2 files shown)


rxwei:
@swift-ci please test

milseman (approved):
LGTM, assuming it isn't superseded by https://github.com/apple/swift-experimental-string-processing/pull/397

stephentyrone (approved):

PR #404, opened 2022-05-12
Ban `]` as literal first character of custom character class

hamishknight:
PCRE, Oniguruma, and ICU allow `]` to appear as the first member of a custom character class, and treat it as literal, due to empty character classes being invalid.

However this behavior isn't particularly intuitive, and makes lexing heuristics such as https://github.com/apple/swift/pull/58835 harder to implement properly. Instead, reject such character classes as being empty, and require escaping if `]` is meant as the first character.


hamishknight:
@swift-ci please test

milseman (approved):

PR #410, opened 2022-05-12
Wrap character classes around One

Azoy:
This PR wraps character classes like `.digit` around `One(.digit)` in contexts where it will be treated as a member reference. I also made the change to force syntax like:
```swift
OneOrMore(.digit)
```
to avoid the inevitable:
```swift
OneOrMore {
  One(.digit)
}
```

(DSL printing needs a refactor)


DaveEwing (approved):

Azoy:
@swift-ci please test

PR #411, opened 2022-05-13
Update contains and split with different parameter names

natecook1000:
This is an alternative resolution to overload resolution, contrasted with #402.


Sources/_StringProcessing/Algorithms/Algorithms/Split.swift
```diff
@@ -309,11 +309,11 @@ extension Collection where Element: Equatable {
   ///   elements.
   @available(SwiftStdlib 5.7, *)
   public func split<C: Collection>(
-    separator: C,
+    pattern: C,
```
rxwei, line 312: For `split`, I think calling this parameter `pattern` would make its role unclear. Have you considered `separatorPattern`?

natecook1000:
Closing in favor of the solution in #402.

PR #412, opened 2022-05-13
Keep track of initial options in compiled program

natecook1000:
This allows us to fix an issue where searching via `firstMatch` or `matches(of:)` would only _start_ searches at a character index, even when a regex has Unicode scalar semantics.

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -35,6 +35,9 @@ extension Compiler.ByteCodeGen {
       builder.buildUnresolvedReference(id: id)
 
     case let .changeMatchingOptions(optionSequence):
+      if !builder.hasReceivedInstructions {
+        builder.initialOptions.apply(optionSequence.ast)
+      }
       options.apply(optionSequence.ast)
 
     case let .unconverted(astAtom):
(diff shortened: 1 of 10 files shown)


Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -35,6 +35,9 @@ extension Compiler.ByteCodeGen {
       builder.buildUnresolvedReference(id: id)
 
     case let .changeMatchingOptions(optionSequence):
+      if !builder.hasReceivedInstructions {
+        builder.initialOptions.apply(optionSequence.ast)
+      }
```
milseman, line 40: I think it's a much cleaner model if we get the initial set by checking the tree for global options. That is, whatever creates `ByteCodeGen` struct sets them.

milseman: That is, resolve the TODO in https://github.com/apple/swift-experimental-string-processing/blob/4f8f67a653d363d56fab477a6205481919cd6529/Sources/_StringProcessing/Compiler.swift#L29

milseman (commented):
Thanks for tackling this! I think we can do this in a way that avoids adding temporal variance.

Sources/_StringProcessing/ByteCodeGen.swift
natecook1000: We discussed — going to merge this as is and look into computing this directly from `DSLTree` in the future.

natecook1000:
@swift-ci Please test

PR #413, opened 2022-05-13
Handle atoms as things to be wrapped in One

Azoy:
standalone atoms, like `\s` get converted to `.whitespace`, so make sure to wrap it around `One(.whitespace)` wherever possible.

### DIFF
--- a/Sources/_StringProcessing/PrintAsPattern.swift
+++ b/Sources/_StringProcessing/PrintAsPattern.swift
@@ -127,7 +127,7 @@ extension PrettyPrinter {
       print("/* TODO: conditional */")
 
     case let .quantification(amount, kind, child):
-      let amount = amount.ast._patternBase
+      let amountStr = amount.ast._patternBase
       var kind = kind.ast?._patternBase ?? ""
       
       // If we've updated our quantification behavior, then use that. This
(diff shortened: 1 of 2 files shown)


DaveEwing (approved):

DaveEwing (approved):

Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -163,23 +177,42 @@ extension PrettyPrinter {
...
-          return
-        }
-        
-        break
-      case let .convertedRegexLiteral(.customCharacterClass(ccc), _):
-        if ccc.isSimplePrint {
-          printSimpleCCC(ccc)
-          return
+      // We can only do this for Optionally, ZeroOrMore, and OneOrMore. Cannot
+      // do it right now for Repeat.
```
natecook1000, line 181: What's going on with `Repeat`? Does it not have the right one-`RegexComponent` inits for use with this?

Azoy: Oh, it looks like it does. I was trying to use it like `Repeat(count: 2, ...)` but it seems its `Repeat(..., count: 2)` (which reads better now that I'm thinking about it.)

natecook1000: Yeah, it's that weird pattern where the closure-based version has to move the "thing to repeat" to the end, but it otherwise comes first.

Azoy:
@swift-ci please test

Sources/_StringProcessing/PrintAsPattern.swift
Azoy: going to merge as is and will fix this in another PR

Azoy:
@swift-ci please test

PR #414, opened 2022-05-13
Underscore internal algorithms methods

natecook1000:
When working on overload resolution, it's trickier with these unnecessary name collisions. This includes a few symbols that might be public eventually, but we can de-underscore them at that point.


natecook1000:
@swift-ci Please test

PR #415, opened 2022-05-13
Don't get stuck on empty matches

natecook1000:
This fixes an issue where calling `matches(of:)` with an pattern that matches an empty substring gets stuck searching the same position over and over.


natecook1000:
@swift-ci Please test

PR #416, opened 2022-05-15
Remove the last SPI use of _RegexParser symbols

natecook1000:
PatternConverter uses one last AST-based API defined in _StringProcessing, which is technically not allowed but not
problematic b/c PatternConverter also imports _RegexParser. However, this can have compile-time problems later on, so this changes the entry point for PatternConverter to just use an Any parameter that is checked at runtime to actually be AST.


natecook1000:
@swift-ci Please test

rxwei (approved):

PR #417, opened 2022-05-15
[5.7] Remove the last SPI use of _RegexParser symbols (#416)

natecook1000:
PatternConverter uses one last AST-based API defined in _StringProcessing, which is technically not allowed but not
problematic b/c PatternConverter also imports _RegexParser. However, this can have compile-time problems later on, so this changes the entry point for PatternConverter to just use an Any parameter that is checked at runtime to actually be AST.

Cherry pick of #416, resolves rdar://92876793.


natecook1000:
@swift-ci Please test

nkcsgexi (approved):

PR #418, opened 2022-05-16
[5.7] Algorithm correctness cherry-picks

natecook1000:
This is a cherry-pick of three algorithms implementation improvements:
- #415 resolves an infinite loop when `matches(of:)` is called with a regex that matches an empty subsequence
- #414 is an NFC change that underscores internal algorithms methods to simplify disambiguation
- #412 adds a step during regex compilation that tracks the initial options, so that searching can honor the regex's semantic level


natecook1000:
@swift-ci Please test

stephentyrone (approved):

Issue #401, opened 2022-05-11
Digit matching behaving as intended?


stephentyrone:
Reposted from the Swift forums: https://forums.swift.org/t/bad-digit-matching-bugreport-regarding-se-0354-regex-literals/57262/1

Problem: Some digit character groups match number-like grapheme clusters.

```
// this matches:
try /[1-2]/.wholeMatch(in: "1️⃣")

// still matches:
try /[1-2]/.asciiOnlyDigits().wholeMatch(in: "1️⃣")

// does not match:
try /[12]/.wholeMatch(in: "1️⃣")
```

Above described behavior seems inconsistent and difficult to predict. Shouldn't [1-2] and [12] be identical? Should they match anything outside of ascii?

Note: :one: is U+0031 (ascii digit 1) U+FE0F (VARIATION SELECTOR-16) U+20E3 (COMBINING ENCLOSING KEYCAP)

Same is true for 1︎⃣: U+0031 (ascii digit 1) U+FE0E (VARIATION SELECTOR-15) U+20E3 (COMBINING ENCLOSING KEYCAP)

rdar://96898279

NikolaiRuhe:
If there are questions about my report: I'm following this issue.

milseman:
Note that the emoji compares between `"1"` and `"2"` as a Character. @natecook1000

NikolaiRuhe:
> Note that the emoji compares between `"1"` and `"2"` as a Character. @natecook1000

This explains a lot of above behavior—at least technically. On the other hand it scares me away from using regular expressions when writing code that should sanitize arbitrary input.

Maybe an ascii-only mode would help? In my code I'm manually checking input now:

`input.utf8.allSatisfy { $0 <= 127 }`.

natecook1000:
Thanks for the example, @NikolaiRuhe! It looks like custom character class ranges will need something a bit different than `Character.<`, at least by default.

milseman:
Note that in scalar semantics you should get your desired behavior. It's a reasonable interpretation to extend `A-Z` to `"A"..."Z"`, but probably undesirable and counter intuitive. The Unicode pitch was arguing for using NFD for ranges, but that wouldn't affect this example.

PR #422, opened 2022-05-18
Limit custom character class ranges to single scalars

natecook1000:
As shown by issue #401, the standard lexicographic-ordering-based comparisons for characters yield very unexpected results when matching with ranges in custom character classes. This change resolves that unexpected behavior by only matching single-scalar characters within ranges and only allowing single-scalar characters to be range endpoints.

Multi-scalar characters are still allowed within custom character classes, and meta characters and built-in character classes continue to function in the same way as before. With this change, we have the following behavior:

```swift
try /[1-2]/.wholeMatch(in: "1")     // "1"
try /[12]/.wholeMatch(in: "1")      // "1"
try /[1-2]/.wholeMatch(in: "1️⃣")    // nil
try /[12]/.wholeMatch(in: "1️⃣")     // nil

try /\d/.wholeMatch(in: "1️⃣")                      // "1️⃣"
try /\d/.asciiOnlyDigits().wholeMatch(in: "1️⃣")    // nil
try /[\d]/.wholeMatch(in: "1️⃣")                    // "1️⃣"
try /[\d]/.asciiOnlyDigits().wholeMatch(in: "1️⃣")  // nil

try /[🇦🇫-🇿🇼]/.wholeMatch(in: "Flags! 🇬🇭🇰🇷")       // error: invalid character class range
```

### DIFF
--- a/Sources/_RegexParser/Regex/AST/Atom.swift
+++ b/Sources/_RegexParser/Regex/AST/Atom.swift
@@ -771,7 +771,7 @@ extension AST.Atom {
   /// range.
   public var isValidCharacterClassRangeBound: Bool {
     // If we have a literal character value for this, it can be used as a bound.
-    if literalCharacterValue != nil { return true }
+    if literalCharacterValue?.hasExactlyOneScalar == true { return true }
     switch kind {
     // \cx, \C-x, \M-x, \M-\C-x, \N{...}
     case .keyboardControl, .keyboardMeta, .keyboardMetaControl, .namedCharacter:
--- a/Sources/_RegexParser/Utility/Misc.swift
+++ b/Sources/_RegexParser/Utility/Misc.swift
@@ -19,6 +19,13 @@ extension Substring {
   var string: String { String(self) }
 }
 
+extension Character {
+  /// Whether this character is made up of exactly one Unicode scalar value.
+  public var hasExactlyOneScalar: Bool {
+    unicodeScalars.index(after: unicodeScalars.startIndex) == unicodeScalars.endIndex
+  }
+}
+
 extension CustomStringConvertible {
   @_alwaysEmitIntoClient
   public var halfWidthCornerQuoted: String {
--- a/Sources/_StringProcessing/ConsumerInterface.swift
+++ b/Sources/_StringProcessing/ConsumerInterface.swift
@@ -248,10 +248,10 @@ extension DSLTree.CustomCharacterClass.Member {
       return c
     case let .range(low, high):
       // TODO:
-      guard let lhs = low.literalCharacterValue else {
+      guard let lhs = low.literalCharacterValue, lhs.hasExactlyOneScalar else {
         throw Unsupported("\(low) in range")
       }
-      guard let rhs = high.literalCharacterValue else {
+      guard let rhs = high.literalCharacterValue, rhs.hasExactlyOneScalar else {
         throw Unsupported("\(high) in range")
       }
 
@@ -262,6 +262,7 @@ extension DSLTree.CustomCharacterClass.Member {
         return { input, bounds in
           // TODO: check for out of bounds?
           let curIdx = bounds.lowerBound
+          guard input[curIdx].hasExactlyOneScalar else { return nil }
           if (lhsLower...rhsLower).contains(input[curIdx].lowercased()) {
             // TODO: semantic level
             return input.index(after: curIdx)
@@ -273,6 +274,7 @@ extension DSLTree.CustomCharacterClass.Member {
         return { input, bounds in
           // TODO: check for out of bounds?
           let curIdx = bounds.lowerBound
+          guard input[curIdx].hasExactlyOneScalar else { return nil }
           if (lhs...rhs).contains(input[curIdx]) {
             // TODO: semantic level
             return input.index(after: curIdx)
--- a/Sources/_StringProcessing/ConsumerInterface.swift
+++ b/Sources/_StringProcessing/ConsumerInterface.swift
@@ -60,24 +60,53 @@ extension DSLTree.Atom {
     _ opts: MatchingOptions
   ) throws -> MEProgram<String>.ConsumeFunction? {
     let isCaseInsensitive = opts.isCaseInsensitive
-    
+    let isCharacterSemantics = opts.semanticLevel == .graphemeCluster
+
     switch self {
     case let .char(c):
-      // TODO: Match level?
       return { input, bounds in
-        let low = bounds.lowerBound
+        let nextIndex = isCharacterSemantics
+          ? input.index(after: bounds.lowerBound)
+          : input.unicodeScalars.index(after: bounds.lowerBound)
+
+        var curIdx = bounds.lowerBound
         if isCaseInsensitive && c.isCased {
-          return input[low].lowercased() == c.lowercased()
-            ? input.index(after: low)
-            : nil
+          if isCharacterSemantics {
+            return input[curIdx].lowercased() == c.lowercased()
+              ? nextIndex
+              : nil
+          } else {
+            // FIXME: How do multi-scalar characters match in case insensitive mode?
+            return input.unicodeScalars[curIdx].properties.lowercaseMapping == c.lowercased()
+              ? nextIndex
+              : nil
+          }
         } else {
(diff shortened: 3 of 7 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -60,24 +60,53 @@ extension DSLTree.Atom {
...
+        var curIdx = bounds.lowerBound
         if isCaseInsensitive && c.isCased {
-          return input[low].lowercased() == c.lowercased()
-            ? input.index(after: low)
-            : nil
+          if isCharacterSemantics {
+            return input[curIdx].lowercased() == c.lowercased()
+              ? nextIndex
+              : nil
+          } else {
```
milseman, line 78: Remove else

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -60,24 +60,53 @@ extension DSLTree.Atom {
...
+          }
         } else {
-          return input[low] == c
-            ? input.index(after: low)
-            : nil
+          if isCharacterSemantics {
+            return input[curIdx] == c
+              ? nextIndex
+              : nil
+          } else {
```
milseman, line 89: Remove else

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -60,24 +60,53 @@ extension DSLTree.Atom {
...
         if isCaseInsensitive && c.isCased {
-          return input[low].lowercased() == c.lowercased()
-            ? input.index(after: low)
-            : nil
+          if isCharacterSemantics {
+            return input[curIdx].lowercased() == c.lowercased()
+              ? nextIndex
+              : nil
+          } else {
+            // FIXME: How do multi-scalar characters match in case insensitive mode?
```
milseman, line 79: Do you have an issue or something to clarify this?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -247,40 +276,48 @@ extension DSLTree.CustomCharacterClass.Member {
...
-          }
+      let isCaseInsensitive = opts.isCaseInsensitive
+      let isCharacterSemantic = opts.semanticLevel == .graphemeCluster
+      
+      return { input, bounds in
+        // TODO: check for out of bounds?
+        let curIdx = bounds.lowerBound
+        let nextIndex = isCharacterSemantic
+          ? input.index(after: curIdx)
+          : input.unicodeScalars.index(after: curIdx)
```
milseman, line 297: Really feel like this should be a helper function or fully refactored. It seems ripe for bugs if vigilance is required to remember to support scalar semantics

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -247,40 +276,48 @@ extension DSLTree.CustomCharacterClass.Member {
...
+        if isCharacterSemantic && !input[curIdx].hasExactlyOneScalar {
```
milseman, line 298: What does this do to non-NFC characters?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -247,40 +276,48 @@ extension DSLTree.CustomCharacterClass.Member {
...
+        if !isCaseInsensitive {
           return nil
         }
+        
+        let stringRange = String(lhs)...String(rhs)
+        if (scalar.properties.changesWhenLowercased
+            && stringRange.contains(scalar.properties.lowercaseMapping))
+          || (scalar.properties.changesWhenUppercased
+            && stringRange.contains(scalar.properties.uppercaseMapping)) {
+          return nextIndex
```
milseman, line 315: Isn't this going _back_ to lexicographical contains?

Sources/_StringProcessing/_CharacterClassModel.swift
```diff
@@ -103,6 +103,11 @@ public struct _CharacterClassModel: Hashable {
           return c == character
         }
       case .range(let range):
+        // Ranges can be formed with single-scalar characters, and can only
+        // match as such.
+        // FIXME: Convert to canonical composed version before testing?
+        guard character.hasExactlyOneScalar else { return false }
```
milseman, line 109: This is just wrong and breaks canonical equivalence.

Tests/RegexTests/MatchTests.swift
```diff
@@ -1507,31 +1539,28 @@ extension RegexTests {
...
-              xfail: true)
-    firstMatchTest(#"e[\u{300}-\u{320}]$"#, input: eDecomposed, match: eDecomposed,
-              xfail: true)
-    firstMatchTest(#"[a-z][\u{300}-\u{320}]$"#, input: eDecomposed, match: eDecomposed,
-              xfail: true)
+    // Grapheme cluster semantics
+    firstMatchTest(#"e[\u{301}]$"#, input: eComposed, match: nil)
+    firstMatchTest(#"e[\u{300}-\u{320}]$"#, input: eComposed, match: nil)
+    firstMatchTest(#"[e][\u{300}-\u{320}]$"#, input: eComposed, match: nil)
+    firstMatchTest(#"[a-z][\u{300}-\u{320}]$"#, input: eComposed, match: nil)
```
milseman, line 1563: We also need positive tests for grapheme-semantics under canonical equivalence.

Sources/_RegexParser/Regex/AST/Atom.swift
```diff
@@ -771,7 +771,7 @@ extension AST.Atom {
   /// range.
   public var isValidCharacterClassRangeBound: Bool {
     // If we have a literal character value for this, it can be used as a bound.
-    if literalCharacterValue != nil { return true }
+    if literalCharacterValue?.hasExactlyOneScalar == true { return true }
```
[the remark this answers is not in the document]
natecook1000: It does for sure, it would help if we normalized all characters as we stored them in the custom character class model.

Sources/_RegexParser/Utility/Misc.swift
```diff
@@ -19,6 +19,13 @@ extension Substring {
   var string: String { String(self) }
 }
 
+extension Character {
+  /// Whether this character is made up of exactly one Unicode scalar value.
+  public var hasExactlyOneScalar: Bool {
```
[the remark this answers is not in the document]
natecook1000: This was previously in the _StringProcessing module; we need it _RegexParser for the compile-time validation.

Sources/_RegexParser/Regex/AST/Atom.swift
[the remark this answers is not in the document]
milseman: Which normalization?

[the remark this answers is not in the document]
natecook1000: NFC would allow us to permit as many characters as possible to act as endpoints.

Sources/_StringProcessing/ConsumerInterface.swift
natecook1000: I'll make it a helper function, but the real issue is that the two views share the same index type, and we can't enforce that you call the helper function. Maybe we could look at designing a wrapper type that would distinguish between the two? Seems like it add a lot of friction…

```swift
struct ViewLockedIndex<View> {
    var index: String.Index
}

extension String {
    func index(after i: ViewLockedIndex<String>) -> ViewLockedIndex<String> {
        .init(index: index(after: i.index))
    }

    func index(after i: ViewLockedIndex<String.UnicodeScalarView>) -> ViewLockedIndex<String.UnicodeScalarView>
        .init(index: self.unicodeScalars.index(after: i.index))
    }
}
```

milseman: What I mean is rather than have a bunch of inner-most ifs, what would the code look like to have an outer-if? That is a more bottoms-up refactoring

hamishknight:
Picking up this work in https://github.com/apple/swift-experimental-string-processing/pull/570

stephentyrone:
Is this obsoleted by #570?

PR #428, opened 2022-05-23
[5.7] Tighten up some syntax rules

### LINKED ISSUE #303: Regex parser should be wary of combining characters

hamishknight:
*5.7 cherry-pick of https://github.com/apple/swift-experimental-string-processing/pull/393*

- Ban numeric escapes in custom character classes e.g `[\7]`. In PCRE, this is an octal sequence, but we require a `0` prefix.
- Ban multiple-scalar non-letter non-digit ASCII characters, as they may be confusable with metacharacters.
- Reserve `<{...}>` for a future interpolation syntax.

Resolves #303

### DIFF
--- a/Sources/_RegexParser/Regex/AST/AST.swift
+++ b/Sources/_RegexParser/Regex/AST/AST.swift
@@ -52,6 +52,9 @@ extension AST {
     /// Comments, non-semantic whitespace, etc
     case trivia(Trivia)
 
+    /// Intepolation `<{...}>`, currently reserved for future use.
+    case interpolation(Interpolation)
+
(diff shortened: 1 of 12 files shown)


hamishknight:
@swift-ci please test

PR #431, opened 2022-05-24
Some trivia fixes

### LINKED ISSUE #387: Non-semantic whitespace doesn't work for character class ranges
The following doesn't parse correctly:

```
(?x)[ a - c ]
```

### LINKED ISSUE #429: Allow empty inline comments
We currently reject `(?#)`, but it should be allowed.

### LINKED ISSUE #430: Allow end-of-line comments in custom character classes
We currently don't parse end-of-line comments as comments in a character class such as:

```
let r = #/
[
  a # comment
]
/#
```

However ICU supports this.

hamishknight:
- Allow empty inline comments (resolves #429)
- Parse end-of-line comments in character classes (resolves #430)
- Parse trivia between character class range operands (resolves #387)
- Lex whitespace in range quantifiers allowing for e.g `x{2, 3}`. This differs from PCRE, which would treat such a sequence as literal, but it's more intuitive to allow whitespace.

### DIFF
--- a/Sources/_RegexParser/Regex/AST/CustomCharClass.swift
+++ b/Sources/_RegexParser/Regex/AST/CustomCharClass.swift
@@ -51,11 +51,16 @@ extension AST {
       public var lhs: Atom
       public var dashLoc: SourceLocation
       public var rhs: Atom
+      public var trivia: [AST.Trivia]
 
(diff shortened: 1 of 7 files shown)


hamishknight:
@swift-ci please test

milseman (approved):
LGTM

PR #432, opened 2022-05-24
Allow unbounded quoted sequences `\Q...`

hamishknight:
PCRE and ICU both support quoted sequences that don't have a terminating `\E`. Update the parsing to allow this. Additionally, allow empty quoted sequences outside of custom character classes, which is consistent with ICU. Finally, don't allow quoted sequences to span multiple lines in extended syntax literals.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
@@ -579,7 +579,7 @@ extension Source {
 
   /// Try to consume quoted content
   ///
-  ///     Quote -> '\Q' (!'\E' .)* '\E'
+  ///     Quote -> '\Q' (!'\E' .)* '\E'?
   ///
   /// With `SyntaxOptions.experimentalQuotes`, also accepts
   ///
@@ -592,9 +592,24 @@ extension Source {
   mutating func lexQuote(context: ParsingContext) throws -> AST.Quote? {
     let str = try recordLoc { src -> String? in
       if src.tryEat(sequence: #"\Q"#) {
-        return try src.expectQuoted(endingWith: #"\E"#).value
+        let contents = src.lexUntil { src in
+          src.isEmpty || src.tryEat(sequence: #"\E"#)
+        }.value
+
+        // In multi-line literals, the quote may not span multiple lines.
+        if context.syntax.contains(.multilineExtendedSyntax),
+            contents.spansMultipleLinesInRegexLiteral {
+          throw ParseError.quoteMayNotSpanMultipleLines
+        }
+
+        // The sequence must not be empty in a custom character class.
+        if context.isInCustomCharacterClass && contents.isEmpty {
+          throw ParseError.expectedNonEmptyContents
+        }
+        return contents
       }
       if context.experimentalQuotes, src.tryEat("\"") {
+        // TODO: Can experimental quotes be empty?
         return try src.expectQuoted(endingWith: "\"", ignoreEscaped: true).value
       }
       return nil
--- a/Tests/RegexTests/ParseTests.swift
+++ b/Tests/RegexTests/ParseTests.swift
@@ -754,6 +754,14 @@ extension RegexTests {
     // This follows the PCRE behavior.
     parseTest(#"\Q\\E"#, quote("\\"))
 
+    // ICU allows quotes to be empty outside of custom character classes.
+    parseTest(#"\Q\E"#, quote(""))
+
+    // Quotes may be unterminated.
+    parseTest(#"\Qab"#, quote("ab"))
+    parseTest(#"\Q"#, quote(""))
+    parseTest("\\Qab\\", quote("ab\\"))
+
     parseTest(#"a" ."b"#, concat("a", quote(" ."), "b"),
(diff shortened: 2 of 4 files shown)


hamishknight:
@swift-ci please test

hamishknight:
Ah actually it's probably better to continue reject `\Q\E` as `[\Q\E]` should definitely be illegal

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -578,9 +578,21 @@ extension Source {
...
+          src.isEmpty || src.starts(with: #"\E"#)
+        }.value
+
+        // If we have an ending, the sequence shouldn't be empty. This isn't a
+        // particulary useful thing to express, and [\Q\E] should definitely be
+        // illegal. The unbounded case may however be empty.
+        let ending = src.tryEat(sequence: #"\E"#)
+        if ending && contents.isEmpty {
+          throw ParseError.expectedNonEmptyContents
+        }
```
hamishknight, line 591: @milseman ICU does allow empty `\Q\E` outside of a character class, do you think it's worth supporting that? PCRE doesn't support it, and IMO it doesn't seem particularly useful, but we could support it while rejecting empty `\Q\E` within a character class. Any thoughts?

milseman: Weakly in favor of allowing it unless you can think of any downsides?

milseman (approved):

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
hamishknight: Seems reasonable to me, I can't think of any downsides. I've updated the PR to allow them, and also added the rule to reject quotes that span multiple lines.

hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -592,9 +592,24 @@ extension Source {
...
+
+        // The sequence must not be empty in a custom character class.
+        if context.isInCustomCharacterClass && contents.isEmpty {
+          throw ParseError.expectedNonEmptyContents
+        }
+        return contents
       }
       if context.experimentalQuotes, src.tryEat("\"") {
+        // TODO: Can experimental quotes be empty?
```
milseman, line 612: I suspect we'll be dropping experimental mode in favor of `(?"quoted")`

Tests/RegexTests/ParseTests.swift
```diff
@@ -2711,6 +2720,26 @@ extension RegexTests {
...
+        \Qabc
+          \E
+      /#
+      """#, .quoteMayNotSpanMultipleLines)
+
+    diagnosticWithDelimitersTest(#"""
+      #/
+        \Q
+      /#
+      """#, .quoteMayNotSpanMultipleLines)
```
milseman, line 2741: Should this just be the regex `/\Q/`, i.e. it's not spanning multiple lines because there's no additional line? I'm ok either way, and perhaps reporting an error is helpful.

milseman (approved):

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -579,7 +579,7 @@ extension Source {
 
   /// Try to consume quoted content
   ///
-  ///     Quote -> '\Q' (!'\E' .)* '\E'
+  ///     Quote -> '\Q' (!'\E' .)* '\E'?
```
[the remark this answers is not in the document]
hamishknight: It has to be terminated by the end of the regex yeah, I can update the grammar to be explicit about that. I don't believe we ever added a convention for it, `<end-of-input>` seems reasonable to me tho.

Tests/RegexTests/ParseTests.swift
hamishknight: Good point! Currently we don't do any whitespace stripping, so if we just allowed it as-is, it would parse as `\Q\n       \E`. That being said, it seems reasonable enough to strip everything after-and-including the last newline given we enforce that the closing delimiter must appear on a newline. Does that sound reasonable to you, or should be leave it as an error for now?

milseman: I could go either way, what do you think?

hamishknight: I think we should probably have a consistent behavior between `(?-x)` and unterminated `\Q` whichever way we go. For now at least, I'm tempted to say we should keep everything using the rule where semantic whitespace cannot span multiple lines anywhere in the literal, so `(?-x)` and unterminated `\Q` are always invalid. So this test case would remain as-is. Given this rule only affects literals, we'd be free to lift in the future if we wanted. Does that sound reasonable?

hamishknight:
@swift-ci please test

Tests/RegexTests/ParseTests.swift
milseman: SGTM

PR #436, opened 2022-05-24
Escape backslashes properly when converting to the DSL

DaveEwing:
Fix an issue where we weren't properly escaping backslashes in quote strings when converting to the DSL.

### DIFF
--- a/Sources/_StringProcessing/PrintAsPattern.swift
+++ b/Sources/_StringProcessing/PrintAsPattern.swift
@@ -578,7 +578,7 @@ extension PrettyPrinter {
 extension String {
   // TODO: Escaping?
   fileprivate var _quoted: String {
-    "\"\(self._replacing("\"", with: "\\\""))\""
+    "\"\(self._replacing("\\", with: "\\\\")._replacing("\"", with: "\\\""))\""
   }
 }
 
--- a/Sources/_StringProcessing/PrintAsPattern.swift
+++ b/Sources/_StringProcessing/PrintAsPattern.swift
@@ -578,7 +578,7 @@ extension PrettyPrinter {
 extension String {
   // TODO: Escaping?
   fileprivate var _quoted: String {
-    "\"\(self._replacing("\\", with: "\\\\")._replacing("\"", with: "\\\""))\""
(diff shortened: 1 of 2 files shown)


DaveEwing:
@swift-ci please test

Sources/_StringProcessing/PrintAsPattern.swift
```diff
@@ -578,7 +578,7 @@ extension PrettyPrinter {
 extension String {
   // TODO: Escaping?
   fileprivate var _quoted: String {
-    "\"\(self._replacing("\"", with: "\\\""))\""
+    "\"\(self._replacing("\\", with: "\\\\")._replacing("\"", with: "\\\""))\""
```
milseman, line 581:
```suggestion
    "\"\(self._replacing(#"\"#, with: #"\\"#)._replacing(#"""#, with: #"\""#))\""
```

Tests/RegexTests/RenderDSLTests.swift
```diff
@@ -114,4 +114,12 @@ extension RenderDSLTests {
         """)
     }
   }
+  
+  func testQuoting() throws {
+    try testConversion(#"\\\"a\""#, #"""
+      Regex {
+        "\\\"a\""
```
milseman, line 121: Is this the right conversion? The input is raw, so IIUC it has 3 full backslashes, then a quote then a full backslash and another quote.

Sources/_StringProcessing/PrintAsPattern.swift
milseman: Or we can emit it as a raw string if it has any backslashes or quotes inside it. We could do a quick check to count the number of hashes prior to a backslash and emit a raw string with one additional one.

Tests/RegexTests/RenderDSLTests.swift
DaveEwing: I'll update the test. The `\`s in `\"` in the regex to convert is ignored when parsing, but not necessary.

DaveEwing:
@swift-ci please test

Azoy (approved):

milseman (approved):

PR #439, opened 2022-05-26
Parse .NET character class subtractions

hamishknight:
This is similar to `--`, but is spelled `-`, and only takes effect when the RHS is a nested custom character class. It additionally must appear as the last member of the custom character class.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
+++ b/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
@@ -60,6 +60,7 @@ enum ParseError: Error, Hashable {
   case cannotRemoveMatchingOptionsAfterCaret
 
   case expectedCustomCharacterClassMembers
+  case dotNetSubtractionMustBeLast
 
   case emptyProperty
   case unknownProperty(key: String?, value: String)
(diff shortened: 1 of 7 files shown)


hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/Diagnostics.swift
```diff
@@ -173,6 +174,8 @@ extension ParseError: CustomStringConvertible {
       return "expected ASCII for '\(c)'"
     case .expectedCustomCharacterClassMembers:
       return "expected custom character class members"
+    case .dotNetSubtractionMustBeLast:
+      return "character class members may not appear after .NET subtraction"
```
milseman, line 178: Do we actually want to tell users something about `.NET` here? Diagnostics should probably be phrased in terms of Swift regex syntax and violations there-of.

milseman (approved):
Since this would otherwise be invalid, I'm ok supporting this. We could also warn/fixit for literals.

I'd also be ok saying this is a weird syntactic corner case to support and excluding it due to its obscurity and how it complicates the syntax.

hamishknight:
Talked offline, we're going to error on this for now

PR #440, opened 2022-05-26
Additional character property parsing

hamishknight:
Formalize the parsing of Unicode block properties (previously they were the remaining Oniguruma special properties), including allowing the shorthand spelling e.g `\p{inBasicLatin}`, which is supported by Perl and Oniguruma (though Perl discourages it).

Additionally, start parsing the Java character properties.


hamishknight:
@swift-ci please test

natecook1000 (approved):
LGTM, we can implement the java special property support, but we'll need the block Unicode data to implement those (cc @Azoy).

PR #448, opened 2022-05-27
Use inits instead of as methods

milseman:


milseman:
@swift-ci please test

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
```diff
@@ -232,21 +182,81 @@ extension Regex where Output == AnyRegexOutput {
...
-  public func `as`<Output>(
-    _ type: Output.Type = Output.self
-  ) -> Regex<Output>? {
-    let result = Regex<Output>(node: root)
-    
-    guard result._verifyType() else {
+  /// Use this initializer to create a strongly typed regex from
+  /// one that was created from a string. Returns `nil` if the types
+  /// don't match.
+  public init?(_ dynamic: Regex<AnyRegexOutput>) {
```
natecook1000, line 195: It feels like we should have a `as: Output.Type = Output.self` parameter to match the run-time string-based initializer. e.g.

```swift
let pattern = "a+(b+)c+"
let a = try Regex(pattern, as: (Substring, Substring).self))
let b: Regex<(Substring, Substring)> = try Regex(pattern)

let dynamic = try Regex(pattern)
let c = Regex(dynamic, as: (Substring, Substring).self)!
let d: Regex<(Substring, Substring)> = Regex(dynamic)!
```

milseman: Argh that syntax highlighting. Any ideas for a better argument name than `dynamic`?

milseman:
@swift-ci please test

PR #454, opened 2022-05-28
Add primary associated type to RegexComponent

natecook1000:
This PR makes the `RegexOutput` associated type primary for `RegexComponent`, and adopts the new lightweight generics syntax across the algorithms methods. With this change, methods that don't rely on the specific output type no longer have an explicit generic parameter:

```diff
-  public func contains<R: RegexComponent>(_ regex: R) -> Bool {
+  public func contains(_ regex: some RegexComponent) -> Bool {
```

And methods that do are generic over that type instead of over the regex component type:

```diff
-  public func firstMatch<R: RegexComponent>(
-    of r: R
-  ) -> Regex<R.RegexOutput>.Match? {
+  public func firstMatch<Output>(
+    of r: some RegexComponent<Output>
+  ) -> Regex<Output>.Match? {
```

(Technically these are all just sugared versions of the old ones, but they do look nicer.)

### DIFF
--- a/Sources/RegexBuilder/Algorithms.swift
+++ b/Sources/RegexBuilder/Algorithms.swift
@@ -23,9 +23,9 @@ extension BidirectionalCollection where SubSequence == Substring {
   /// - Parameter content: A closure that returns a regex to match against.
   /// - Returns: The match if there is one, or `nil` if none.
   @available(SwiftStdlib 5.7, *)
-  public func wholeMatch<R: RegexComponent>(
-    @RegexComponentBuilder of content: () -> R
-  ) -> Regex<R.RegexOutput>.Match? {
+  public func wholeMatch<Output>(
+    @RegexComponentBuilder of content: () -> some RegexComponent<Output>
+  ) -> Regex<Output>.Match? {
     wholeMatch(of: content())
   }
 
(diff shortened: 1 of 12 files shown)


rxwei (approved):

natecook1000:
@swift-ci Please test

Sources/RegexBuilder/Algorithms.swift
```diff
@@ -49,8 +49,8 @@ extension BidirectionalCollection where SubSequence == Substring {
   /// - Returns: `true` if the regex returned by `content` matched anywhere in
   ///   this collection, otherwise `false`.
   @available(SwiftStdlib 5.7, *)
-  public func contains<R: RegexComponent>(
-    @RegexComponentBuilder _ content: () -> R
+  public func contains(
+    @RegexComponentBuilder _ content: () -> some RegexComponent
   ) -> Bool {
     contains(content())
   }
```
hamishknight, line 56: Unfortunately I believe these will fail to integrate with the current Swift compiler due to rdar://92459215, see the comment at the top of the file (I don't know if `some RegexComponent<Output>` works tho)

hamishknight: Opened https://github.com/apple/swift-experimental-string-processing/pull/458 to verify

hamishknight: Hmm, seems to be fixed now (other test failure unrelated)

PR #455, opened 2022-05-28
Implement Regex.mapOutput

Azoy:
This implements `mapOutput` on Regex from: https://github.com/apple/swift-evolution/blob/d2b03bf572c7d13c575067641a5d9d1d52cb1750/proposals/0351-regex-builder.md#mapping-output

### DIFF
--- a/Tests/RegexTests/MapOutputTests.swift
+++ b/Tests/RegexTests/MapOutputTests.swift
@@ -1,0 +1,51 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+@testable import _RegexParser
+
+import XCTest
+@testable import _StringProcessing
+
+enum Transaction: String {
+  case credit
+  case debit
+}
+
+extension RegexTests {
+  func testMapOutput() throws {
+    let regex0: Regex<(Substring, Substring)> = try Regex(#"Transaction: (credit|debit)"#)
+    let string0 = "Transaction: credit"
+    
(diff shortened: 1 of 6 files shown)


Azoy:
@swift-ci please test

Tests/RegexTests/MapOutputTests.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+    
+    let match1 = try XCTUnwrap(string0.firstMatch(of: regex2)?.output)
+    XCTAssertEqual(match1, .credit)
+    
+    let regex3: Regex<Substring> = try Regex(#"Hello"#)
+    let string1 = "Hello"
+    
+    let regex4 = regex3.mapOutput {
+      "\($0) world!"[...]
```
natecook1000, line 45: Curious here — does this cause a problem if you leave off the `[...]`?

Azoy: Previously if the output type was just `Substring` it would've returned only the whole match instead of actually transforming the output, so I'm testing that even with just a `Substring` output type we still attempt to transform.

natecook1000: Can we also add a test of the behavior when dropping an output-mapped regex in a builder block?

milseman (commented):
This is a short-term band aid (which might be necessary). What we need for `mapOutput` is a DSLTree node that will transform / map the actual outputs. E.g. `contains(captureNamed:)` would honor any tuple labels that were added or removed.

We can merge something like this (but with more tests) as a stop-gap, open an issue, and publish a release note about this being a known problem. Should we also fatal-error if those APIs are called or if the mapping is nested in a way we don't fully support?

milseman:
I.e. `mapOutput` should close the semantic gap between the DSL and the literals, enabling these tests: https://github.com/apple/swift-experimental-string-processing/blob/main/Tests/RegexBuilderTests/AnyRegexOutputTests.swift#L252

wwalexander:
Hey, I've been lurking this issue for a little while. @Azoy, do you have any info on the status of this feature?

PR #456, opened 2022-05-29
Add primary associated type to RegexComponent (#454)

natecook1000:
This makes the `RegexOutput` associated type primary for `RegexComponent`, and adopts the new lightweight generics syntax across the algorithms methods. With this change, methods that don't rely on the specific output type no longer have an explicit generic parameter and methods that do rely on `RegexOutput` are generic over that type instead of over the component type.


natecook1000:
@swift-ci Please test

natecook1000:
Passed local and Swift CI ✅ 
cc @stephentyrone

PR #435, opened 2022-05-24
Add additional Unicode API to RegexBuilder.CharacterClass

natecook1000:
This includes revisions to the options API and additional `CharacterClass` type to bring it into alignment with the functionality that we're offering through regex literals. For example, `/\p{NumericValue=1}/` can be written in RegexBuilder syntax as `CharacterClass.numericValue(1)`.

PR #460, opened 2022-05-30
Add regex-specific Matches and Ranges collections

natecook1000:
This prepares for adopting an opaque result type for `matches(of:)`. The old, `CollectionConsumer`-based model moves index-by-index, and isn't aware of the regex's semantic level, which results in inaccurate results for regexes that match at a mid-character index.

The lazy range collection had the same issue, so this includes a wrapper for the new `RegexMatchesCollection` that just returns ranges, and modifies the `replace`/`replacing` methods to operate over a collection of ranges instead of a collection searcher.

### DIFF
--- a/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
+++ b/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
@@ -222,23 +222,22 @@ extension RegexMatchesCollection: Sequence {
     let base: RegexMatchesCollection
     
     // Because `RegexMatchesCollection` eagerly computes the first match for
-    // its `startIndex`, the iterator begins with this current match populated.
-    // For subsequent calls to `next()`, this value is `nil`, and `nextStart`
-    // is used to search for the next match.
-    var currentMatch: Regex<Output>.Match?
+    // its `startIndex`, the iterator can use that match for its initial
+    // iteration. For subsequent calls to `next()`, this value is `false`, and
+    // `nextStart` is used to search for the next match.
+    var initialIteration = true
     var nextStart: String.Index?
     
     init(_ matches: RegexMatchesCollection) {
       self.base = matches
-      self.currentMatch = matches.startIndex.match
-      self.nextStart = currentMatch.flatMap(base.searchIndex(after:))
+      self.nextStart = base.startIndex.match.flatMap(base.searchIndex(after:))
     }
     
     mutating func next() -> Regex<Output>.Match? {
       // Initial case with pre-computed first match
-      if let match = currentMatch {
-        currentMatch = nil
-        return match
+      if initialIteration {
+        initialIteration = false
+        return base.startIndex.match
       }
       
       // `nextStart` is `nil` when iteration has completed
--- a/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
+++ b/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
@@ -246,7 +246,7 @@ extension RegexMatchesCollection: Sequence {
       }
       
       // Otherwise, find the next match (if any) and compute `nextStart`
-      let match = try! base.regex.firstMatch(in: base.input[start...])
+      let match = try? base.regex.firstMatch(in: base.input[start...])
       nextStart = match.flatMap(base.searchIndex(after:))
       return match
     }
--- a/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
+++ b/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
@@ -310,7 +310,7 @@ extension RegexMatchesCollection: Collection {
     
     guard
       let start = searchIndex(after: currentMatch),
-      let nextMatch = try! regex.firstMatch(in: input[start...])
+      let nextMatch = try? regex.firstMatch(in: input[start...])
     else {
       return .end
     }
--- a/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
+++ b/Sources/_StringProcessing/Algorithms/Matching/Matches.swift
@@ -183,13 +183,106 @@ extension BidirectionalCollection {
 
 // MARK: Regex algorithms
 
+@available(SwiftStdlib 5.7, *)
+struct RegexMatchesCollection<Output> {
+  let base: Substring
+  let regex: Regex<Output>
+  let startIndex: Index
+  
+  init(base: Substring, regex: Regex<Output>) {
+    self.base = base
+    self.regex = regex
+    self.startIndex = base.firstMatch(of: regex).map(Index.match) ?? .end
+  }
+}
+
+@available(SwiftStdlib 5.7, *)
+extension RegexMatchesCollection: Collection {
+  enum Index: Comparable {
+    case match(Regex<Output>.Match)
+    case end
+    
+    static func == (lhs: Self, rhs: Self) -> Bool {
+      switch (lhs, rhs) {
+      case (.match(let lhs), .match(let rhs)):
+        return lhs.range == rhs.range
+      case (.end, .end):
+        return true
+      case (.end, .match), (.match, .end):
+        return false
+      }
+    }
+    
+    static func < (lhs: Self, rhs: Self) -> Bool {
+      switch (lhs, rhs) {
+      case (.match(let lhs), .match(let rhs)):
+        // This implementation uses a tuple comparison so that an empty
+        // range `i..<i` will be ordered before a non-empty range at that
+        // same starting point `i..<j`. As of 2022-05-30, `Regex` does not
+        // return matches of this kind, but that is one behavior under
+        // discussion for regexes like /a*|b/ when matched against "b".
+        return (lhs.range.lowerBound, lhs.range.upperBound)
+          < (rhs.range.lowerBound, rhs.range.upperBound)
+      case (.match, .end):
+        return true
+      case (.end, .match), (.end, .end):
+        return false
+      }
+    }
+  }
+  
+  var endIndex: Index {
+    Index.end
+  }
+  
+  func index(after i: Index) -> Index {
+    let currentMatch: Element
+    switch i {
+    case .match(let match):
+      currentMatch = match
+    case .end:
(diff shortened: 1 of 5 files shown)


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,106 @@ extension BidirectionalCollection {
...
+    self.base = base
+    self.regex = regex
+    self.startIndex = base.firstMatch(of: regex).map(Index.match) ?? .end
+  }
+}
+
+@available(SwiftStdlib 5.7, *)
+extension RegexMatchesCollection: Collection {
+  enum Index: Comparable {
+    case match(Regex<Output>.Match)
```
milseman, line 202: Storing the non-trivial element in the index is a design smell, albeit with the stdlib's reliance on `Collection` instead of iterators or true sequences. I don't know of a better workaround though.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,106 @@ extension BidirectionalCollection {
...
+  }
+  
+  func index(after i: Index) -> Index {
+    let currentMatch: Element
+    switch i {
+    case .match(let match):
+      currentMatch = match
+    case .end:
+      fatalError("Can't advance past the 'endIndex' of a match collection.")
+    }
```
milseman, line 245: I think a lot of these switches can be replaced with a `var match: Match?` on the Index.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,106 @@ extension BidirectionalCollection {
...
+    
+    let start: String.Index
+    if currentMatch.range.isEmpty {
+      if currentMatch.range.lowerBound == base.endIndex {
+        return .end
+      }
```
milseman, line 251: Should this check be hoisted and/or asserted upon from non-empty ranges?

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
natecook1000: Agreed

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
milseman: If the current match's lower bound is the `endIndex`, what are some valid values for the upper bound?

natecook1000: Oh, you're right - lowerBound == endIndex definitely signifies that we're at the end of matches.

natecook1000: But even then it doesn't seem necessary to hoist that check - when the range is non empty the upper bound is always the departure point for the next search, and this condition isn't possible. This is just to prevent advancing past the end, and it goes away if we switch to the other "next match after empty" behavior

milseman: IIUC, if you hoist it, it simplifies the code (it's an early exit) and you get the opportunity to add an internal consistency check, which is always very nice to have.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
natecook1000: I added a custom iterator so we won't have to pay this cost in most contexts.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,156 @@ extension BidirectionalCollection {
...
+    
+    init(_ matches: RegexMatchesCollection) {
+      self.base = matches
+      self.currentMatch = matches.startIndex.match
+      self.nextStart = currentMatch.flatMap(base.searchIndex(after:))
+    }
+    
+    mutating func next() -> Regex<Output>.Match? {
+      // Initial case with pre-computed first match
+      if let match = currentMatch {
```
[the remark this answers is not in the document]
natecook1000: Ah, nice 👍🏻

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,156 @@ extension BidirectionalCollection {
...
+  enum Index: Comparable {
+    case match(Regex<Output>.Match)
+    case end
+    
+    var match: Regex<Output>.Match? {
+      switch self {
+      case .match(let match): return match
+      case .end: return nil
+      }
+    }
```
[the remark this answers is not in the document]
natecook1000: Yeah, this is another super common pattern.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
natecook1000: The iterator doesn't hold on to more than the first match, so it could/should be more efficient than `IndexingIterator`.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,155 @@ extension BidirectionalCollection {
...
+
+  struct Iterator: IteratorProtocol {
+    let base: RegexMatchesCollection
+    
+    // Because `RegexMatchesCollection` eagerly computes the first match for
+    // its `startIndex`, the iterator can use that match for its initial
+    // iteration. For subsequent calls to `next()`, this value is `false`, and
+    // `nextStart` is used to search for the next match.
+    var initialIteration = true
+    var nextStart: String.Index?
```
milseman, line 229: What's the difference between `initialIterator == true` and `nextStart == nil`? Do we have 4 cases, 3 cases, or 2? What's the difference between `nextStart` is nil vs endIndex?

natecook1000: There are three cases:
1. `initialIteration == true`: We use the match found (or not) that's stored in `base.startIndex`
2. `nextStart != nil`: We search for the next match starting at `nextStart`
3. `nextStart == nil`: End of iteration, no need to search

`nextStart == endIndex` is a valid search-starting index, just like any other, so it can't represent the "finished iterating" state.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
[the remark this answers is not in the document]
natecook1000: Just remembered; `Match?` doesn't work as an index b/c it would require `Comparable` conformance.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,155 @@ extension BidirectionalCollection {
...
+    if match.range.lowerBound == input.endIndex {
+      return nil
+    }
+    
+    switch regex.initialOptions.semanticLevel {
+    case .graphemeCluster:
+      return input.index(after: match.range.upperBound)
+    case .unicodeScalar:
+      return input.unicodeScalars.index(after: match.range.upperBound)
+    }
```
milseman, line 218: I've seen this code in a few places. I wonder if it should be an underscored helper method somewhere that takes the semantic level as a parameter?

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,155 @@ extension BidirectionalCollection {
...
+        return base.startIndex.match
+      }
+      
+      // `nextStart` is `nil` when iteration has completed
+      guard let start = nextStart else {
+        return nil
+      }
+      
+      // Otherwise, find the next match (if any) and compute `nextStart`
+      let match = try! base.regex.firstMatch(in: base.input[start...])
```
milseman, line 249: Should this be `try!` or `try?`?

milseman (approved):

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
natecook1000: Yep, I'll add it as part of #435.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
natecook1000: `try?` seems like it would be fine, here.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -183,13 +183,155 @@ extension BidirectionalCollection {
...
+  }
+  
+  func index(after i: Index) -> Index {
+    guard let currentMatch = i.match else {
+      fatalError("Can't advance past the 'endIndex' of a match collection.")
+    }
+    
+    guard
+      let start = searchIndex(after: currentMatch),
+      let nextMatch = try! regex.firstMatch(in: input[start...])
```
milseman, line 313: `try?`?

milseman (approved):
LGTM after you fix up the `try!`.

natecook1000:
@swift-ci Please test

natecook1000:
This is strange…
```
/build/swift-experimental-string-processing/Sources/_StringProcessing/Algorithms/Algorithms/Replace.swift:27:18: error: ambiguous use of 'prefix'
    for range in ranges.prefix(maxReplacements) {
                 ^
Swift.Collection:6:40: note: found this candidate
    @inlinable public __consuming func prefix(_ maxLength: Int) -> Self.SubSequence
                                       ^
Swift.Sequence:7:40: note: found this candidate
    @inlinable public __consuming func prefix(_ maxLength: Int) -> PrefixSequence<Self>
                                       ^
```

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #462, opened 2022-05-31
[5.7] Use inits instead of as methods

milseman:
Use inits instead of as methods

Add ARO tests


stephentyrone (approved):

rxwei:
@swift-ci please test

PR #464, opened 2022-05-31
Implement template-string based replacement methods

natecook1000:
These alternative `replace`/`replacing` methods let you pass a template string with `$`-prefixed replacement tokens.

```swift
let str = "a:b c:d"
let swapped = str.replacing(/(\w+):(\w+)/, withTemplate: "$2:$1")
// swapped = "b:a d:c"
```

Tokens can be written as:
- `/\$\d+/`: one or more consecutive digits are read into a single number
- `/\${\d+}/`: you can use this if the token is followed by literal numbers
- `/\${\w+}/`: using a capture by name

Here's surely an uncontroversial design decision: Unrecognized tokens are silently omitted from the output. Other approaches are to include them in the output, or to trap.

`\$` escapes the dollar sign into the output; needs to be written as `\\$` in a regular string literal, where `\$` is an invalid escape sequence.

### DIFF
--- a/Sources/_StringProcessing/Algorithms/Matching/MatchReplace.swift
+++ b/Sources/_StringProcessing/Algorithms/Matching/MatchReplace.swift
@@ -210,7 +210,6 @@ extension RangeReplaceableCollection where SubSequence == Substring {
     
     let replacements = templateString.matches(of: captureReplacementRegex)
       .compactMap { match -> (Range<String.Index>, CaptureLookup)? in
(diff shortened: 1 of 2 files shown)


stephentyrone:
This looks like a pretty good starting point, but I think there's a bunch of open design questions to resolve before we land this on main. Can we spin up a feature branch for templates?

natecook1000:
Sounds good! We can iterate on the `feature/replace_template` branch.

PR #467, opened 2022-06-01
[5.7] Make the `Regex` type Sendable

natecook1000:
This change makes `Regex`, `RegexComponent`, and its component types `Sendable`.

Regex stores a `Program` instance, which lazily lowers the DSLTree into a compiled program. Without synchronization, this lazy compilation is unsafe under concurrency. This change uses atomic initialization for the compiled program.

This is a cherry-pick of #457.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -62,7 +62,10 @@ let package = Package(
             dependencies: []),
         .target(
             name: "_StringProcessing",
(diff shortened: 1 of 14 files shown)


natecook1000:
@swift-ci Please test

PR #469, opened 2022-06-04
Obtain match output elements without materializing the output.

### LINKED ISSUE #267: `Regex.Match` element accessors should not materialize the whole output
Currently `Regex.Match` element accessors are implemented this way:
```swift
  /// Lookup a capture by name or number
  public subscript<T>(dynamicMember keyPath: KeyPath<Output, T>) -> T {
    output[keyPath: keyPath]
  }

  // Allows `.0` when `Match` is not a tuple.
  @_disfavoredOverload
  public subscript(
    dynamicMember keyPath: KeyPath<(Output, _doNotUse: ()), Output>
  ) -> Output {
    output
  }
```

This is not correct as we should not materialize the entire output.

rxwei:
Resolves #267.

### DIFF
--- a/Sources/_RegexParser/Utility/TypeConstruction.swift
+++ b/Sources/_RegexParser/Utility/TypeConstruction.swift
@@ -139,3 +139,62 @@ public enum TypeConstruction {
     return _openExistential(childType, do: helper)
   }
 }
+
+extension TypeConstruction {
+  public static func optionalType<Base>(
+    of base: Base.Type, depth: Int = 1
+  ) -> Any.Type {
+    switch depth {
+    case 0: return base
+    case 1: return Base?.self
+    case 2: return Base??.self
+    case 3: return Base???.self
+    case 4: return Base????.self
+    default:
+      return optionalType(of: Base????.self, depth: depth - 4)
+    }
+  }
+}
+
+extension MemoryLayout {
+  /// Returns the element index that corresponnds to the given tuple element key
+  /// path.
+  /// - Parameters:
+  ///   - keyPath: The key path from a tuple to one of its elements.
+  ///   - elementTypes: The element type of the tuple type.
+  // TODO: It possible to get element types from the type metadata, but it's
+  // more efficient to pass them in since we already know them in the matching
+  // engine.
+  public static func tupleElementIndex<ElementTypes: Collection>(
+    of keyPath: PartialKeyPath<T>,
+    elementTypes: ElementTypes
+  ) -> Int? where ElementTypes.Element == Any.Type {
+    guard let byteOffset = offset(of: keyPath) else {
+      return nil
+    }
+    if byteOffset == 0 { return 0 }
+    var currentOffset = 0
+    for (index, type) in elementTypes.enumerated() {
+      func sizeAndAlignMask<T>(_: T.Type) -> (Int, Int) {
+        (MemoryLayout<T>.size, MemoryLayout<T>.alignment - 1)
+      }
+      // The ABI of an offset-based key path only stores the byte offset, so
+      // this doesn't work if there's a 0-sized element, e.g. `Void`,
+      // `(Void, Void)`. (rdar://63819465)
+      if size == 0 {
+        return nil
+      }
+      let (size, alignMask) = _openExistential(type, do: sizeAndAlignMask)
+      // Align up the offset for this type.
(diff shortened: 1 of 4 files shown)


rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

Sources/_RegexParser/Utility/TypeConstruction.swift
```diff
@@ -139,3 +139,45 @@ public enum TypeConstruction {
...
+    guard let byteOffset = offset(of: keyPath) else {
+      return nil
+    }
+    if byteOffset == 0 { return 0 }
+    var currentOffset = 0
+    for (index, type) in elementTypes.enumerated() {
+      func sizeAndAlignMask<T>(_: T.Type) -> (Int, Int) {
+        (MemoryLayout<T>.size, MemoryLayout<T>.alignment - 1)
+      }
+      // The ABI of an offset-based key path only stores the byte offset, so
```
rxwei, line 165: @lorentey just reminded me of this... 😭
```console
> \(Void, Void).0 == \(Void, Void).1
$R0: Bool = true
```

@Azoy, do you have the radar number for this issue I can put in the comment?

rxwei:
@swift-ci please test

Sources/_RegexParser/Utility/TypeConstruction.swift
Azoy: rdar://63819465

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -37,33 +41,48 @@ extension Regex.Match {
...
+  var wholeMatchType: Any.Type {
+    value.map { type(of: $0) } ?? Substring.self
+  }
+
   /// Accesses a capture by its name or number.
   public subscript<T>(dynamicMember keyPath: KeyPath<Output, T>) -> T {
-    output[keyPath: keyPath]
+    // Note: We should be able to get the element offset from the key path
+    // itself even at compile time. We need a better way of doing this.
+    guard let outputTupleOffset = MemoryLayout.tupleElementIndex(
```
rxwei, line 76: When we store the whole match inside the capture list, we will be able to precompute these offsets for all elements in a `Regex.Match`. I'll do that next.

rxwei:
@swift-ci please test

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
```diff
@@ -62,9 +62,11 @@ extension AnyRegexOutput: RandomAccessCollection {
...
     /// The captured value, `nil` for no-capture
     public var value: Any? {
-      // FIXME: Should this return the substring for default-typed
-      // values?
-      representation.value
+      representation.value ?? substring
+    }
+
+    internal var type: Any.Type {
+      value.map { Swift.type(of: $0) } ?? Substring.self
```
milseman, line 69: What about optional Substring?

Sources/_StringProcessing/Regex/Match.swift
```diff
@@ -37,33 +41,48 @@ extension Regex.Match {
...
       let typeErasedMatch = anyRegexOutput.existentialOutput(
-        from: anyRegexOutput.input[range]
+        from: input[range]
       )
       return typeErasedMatch as! Output
     }
   }
 
+  var wholeMatchType: Any.Type {
+    value.map { type(of: $0) } ?? Substring.self
```
milseman, line 69: What if it's optional? Also, are we tracking nested optionals correctly here (I.e. do they get turned into values or are they counted)?

milseman (commented):
It might be useful to write tests where a capture transform produces a `Substring` by e.g. returning `.dropFirst()`. Similarly for when that capture-transform appears inside an `Optionally` or is adding or removing a layer of Substring optionality.

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
rxwei: Good point. Fixed.

Sources/_StringProcessing/Regex/Match.swift
rxwei: When there's no dedicated value for whole match, it can only be a `Substring`.

PR #470, opened 2022-06-05
Fully generalize "whole match" in the engine and enable transforming custom types

rxwei:
* Track the whole match as an element of the "capture list" in the matching engine. Do so by emitting code as an implicit `capture` around the root node.
* No longer handle `matcher` as a special case within `capture` lowering, because the matcher can be arbitrarily nested within "output-forwarding" nodes, such as a `changeMatchingOptions` non-capturing group. Instead, make the bytecode emitter carry a result value so that a custom output can be propagated through any forwarding nodes.
  ```swift
  Regex {
    Capture(
      SemanticVersionParser()
        .ignoringCase()
        .matchingSemantics(.unicodeScalar)
    ) // This would not work previously.
  }
  ```
* Collapse DSLTree node `transform` into `capture`, because a transform can never be standalone (without a `capture` parent). This greatly simplifies `capture` lowering.
* Make the bytecode's capture transform use type `(Input, _StoredCapture) -> Any` so that it can transform any whole match, not just `Substring`. This means you can now transform any captured value, including a custom-consuming regex component's result!
  ```swift
  Regex {
    "version:"
    OneOrMore(.whitespace)
    Capture {
      SemanticVersionParser() // Regex<SemanticVersion>
    } transform: {
      // (SemanticVersion) -> SomethingElse
    }
  }
  ```
  The transforms of `Capture` and `TryCapture` are now generalized from taking `Substring` to taking generic parameter `W` (the whole match).
* Fix an issue where initial options were applied based solely on whether the bytecode had any instructions, failing examples such as `((?i:.))`. It now checks whether the first matchable atom has been emitted.

### DIFF
--- a/Sources/_StringProcessing/Regex/AnyRegexOutput.swift
+++ b/Sources/_StringProcessing/Regex/AnyRegexOutput.swift
@@ -15,7 +15,7 @@
 @available(SwiftStdlib 5.7, *)
 public struct AnyRegexOutput {
   internal let input: String
-  internal let _elements: [ElementRepresentation]
+  internal var _elements: [ElementRepresentation]
 }
 
 @available(SwiftStdlib 5.7, *)
@@ -37,7 +37,7 @@ extension AnyRegexOutput {
     as type: Output.Type = Output.self
   ) -> Output? {
     let elements = map {
-      $0.existentialOutputComponent(from: input[...])
+      $0.existentialOutputComponent(from: input)
     }
     return TypeConstruction.tuple(of: elements) as? Output
   }
@@ -52,7 +52,7 @@ extension AnyRegexOutput: RandomAccessCollection {
 
     /// The range over which a value was captured. `nil` for no-capture.
     public var range: Range<String.Index>? {
-      representation.bounds
+      representation.content?.range
     }
 
     /// The slice of the input over which a value was captured. `nil` for no-capture.
@@ -60,12 +60,12 @@ extension AnyRegexOutput: RandomAccessCollection {
       range.map { input[$0] }
     }
 
-    /// The captured value, `nil` for no-capture
+    /// The captured value, `nil` for no-capture.
     public var value: Any? {
-      representation.value ?? substring
+      representation.value(forInput: input)
     }
 
-    internal var type: Any.Type {
+    public var type: Any.Type {
       representation.type
     }
 
@@ -168,7 +168,7 @@ extension Regex {
 
   /// Returns whether a named-capture with `name` exists
   public func contains(captureNamed name: String) -> Bool {
-    program.tree.root._captureList.captures.contains(where: {
+    program.tree.captureList.captures.contains(where: {
       $0.name == name
     })
   }
@@ -196,8 +196,7 @@ extension Regex.Match where Output == AnyRegexOutput {
   public init<Output>(_ match: Regex<Output>.Match) {
     self.init(
       anyRegexOutput: match.anyRegexOutput,
-      range: match.range,
-      value: match.value
+      range: match.range
     )
   }
 }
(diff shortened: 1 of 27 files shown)


rxwei:
@swift-ci please test

rxwei:
@swift-ci please test Linux

rxwei:
For some reason the CI failure only occurs on Linux, while both platforms have the same compiler.
```console
[252/256] Compiling RegexBuilderTests RegexDSLTests.swift
/build/swift-experimental-string-processing/Tests/RegexBuilderTests/RegexDSLTests.swift:586:18: error: the compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions
    let regex3 = Regex {
                 ^~~~~~~
/build/swift-experimental-string-processing/Tests/RegexBuilderTests/RegexDSLTests.swift:798:11: warning: initialization of immutable value 'input' was never used; consider replacing with assignment to '_' or removing it
      let input = "      "
      ~~~~^~~~~
      _
```

```console
Apple Swift version 5.8-dev (LLVM 278d67f38c6a910, Swift ee312bc1e20eb01)
Target: x86_64-apple-macosx11.0
```
```console
Swift version 5.8-dev (LLVM 278d67f38c6a910, Swift ee312bc1e20eb01)
Target: x86_64-unknown-linux-gnu
```

rxwei:
@swift-ci please test

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
```diff
@@ -229,17 +226,14 @@ extension AnyRegexOutput {
     /// `Substring` has optional depth `0`, and `Int??` has optional depth `2`.
     let optionalDepth: Int
 
-    /// The bounds of the output element.
-    let bounds: Range<String.Index>?
+    /// The element bounds and the value (if available).
+    let rawValue: (range: Range<String.Index>, value: Any?)?
```
milseman, line 230: The name `rawValue` is used pretty heavily with `RawRepresentable`, but IIUC this tuple alone isn't sufficient for constructing the type. Any particular reason we're storing the tuple instead of having a `range` and `value` field?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -387,47 +371,100 @@ public struct ReferenceID: Hashable, Sendable {
...
+    case .substringFailable(let transform):
+      guard let result = try transform(input as! Substring) else {
         return nil
       }
       assert(type(of: result) == resultType)
       return result
     }
   }
 
+  func callAsFunction(_ input: Substring) throws -> Any? {
```
milseman, line 455: Out of curiosity, why are these closures all union-ed together into a run-time value?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -524,9 +558,43 @@ extension DSLTree.Node {
 
   var _captureList: CaptureList {
     var list = CaptureList()
+    list.append(.init(type: wholeMatchType, optionalDepth: 0, .fake))
```
milseman, line 561: Again, this would argue it should be on DSLTree and not node.

milseman (approved):
Overall LGTM. I'm not sure why transform is back on capture, it seems to complicate the model and make it harder to drop the capture enum. Otherwise it looks good.

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
rxwei: The name is definitely up for debate; any suggestions?

I made it a tuple because when `range`  is non-nil, `value` must be non-nil.  This is so that we can avoid asserting `value != nil` in a bunch of places.  I could also make `range` and `value` be a struct.

rxwei: Renamed the property to `content`, as it represents the content of the capture, whereas other properties like `optionalDepth`, `name`, `referenceID` are metadata.

Sources/_StringProcessing/Regex/DSLTree.swift
rxwei: I suppose we can define a separate `CaptureTransform` for every capture type, e.g. `CaptureTransform` and `FailableCaptureTransform`. But they'd also end up getting unioned as different instruction cases.

rxwei:
> I'm not sure why transform is back on capture, it seems to complicate the model and make it harder to drop the capture enum.

IMO it's the opposite.  I don't think a standalone `transform` tree node is a good fit because transforms are only ever specified in a `Capture` or a `TryCapture`. Before this patch, `transform` was handled as an unreachable case in Processor.swift because it doesn't make sense when it's not part of a capture; that is, `transform` does not have a capture index to correspond to unless its parent is a `capture`.

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

PR #471, opened 2022-06-07
Add additional capture transform tests.

rxwei:
- Add a test where the capture transform produecs a `Substring` from a `Substring`.
- Add a test where the capture transform wraps a `Substring` in an `Optional`.


rxwei:
From #469:
> Similarly for when that capture-transform appears inside an Optionally or is adding or removing a layer of Substring optionality.

In this patch, it's impossible to write a test that removes a layer of Substring optionality because we don't have #470 yet. I'll test in that PR.

rxwei:
@swift-ci please test

PR #477, opened 2022-06-09
Disable Prototypes to work around a CI failure

rxwei:
Work around rdar://94763190


rxwei:
@swift-ci please test

PR #479, opened 2022-06-13
Allow CustomConsuming types to match w/ zero width

natecook1000:
We previously asserted if a custom consuming type matches with zero width, but that isn't necessary or good. A custom type can implement a lookaround assertion or act as a tracer.

### DIFF
--- a/Sources/_StringProcessing/Engine/Processor.swift
+++ b/Sources/_StringProcessing/Engine/Processor.swift
@@ -117,11 +117,15 @@ extension Processor {
     return true
   }
 
-  mutating func advance(to nextIndex: Input.Index) {
-    assert(nextIndex >= bounds.lowerBound)
-    assert(nextIndex <= bounds.upperBound)
-    assert(nextIndex >= currentPosition)
-    currentPosition = nextIndex
+  /// Continue matching at the specified index.
+  ///
+  /// - Precondition: `bounds.contains(index) || index == bounds.upperBound`
+  /// - Precondition: `index >= currentPosition`
+  mutating func resume(at index: Input.Index) {
+    assert(index >= bounds.lowerBound)
+    assert(index <= bounds.upperBound)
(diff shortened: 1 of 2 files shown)


natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -120,7 +120,7 @@ extension Processor {
   mutating func advance(to nextIndex: Input.Index) {
     assert(nextIndex >= bounds.lowerBound)
     assert(nextIndex <= bounds.upperBound)
-    assert(nextIndex > currentPosition)
+    assert(nextIndex >= currentPosition)
```
milseman, line 123: Is this the better change, or is it better to not call `advance(to:)` if there's nowhere to advance to?

natecook1000: It doesn't seem necessary to require the precondition for this method, given that a zero-width "advance" is a reasonable output for a consumer/other regex component. Would it sound better to you to rename this to `continue(at:)`?

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -117,11 +117,15 @@ extension Processor {
...
-    assert(nextIndex >= currentPosition)
-    currentPosition = nextIndex
+  /// Continue matching at the specified index.
+  ///
+  /// - Precondition: `bounds.contains(index) || index == bounds.upperBound`
+  /// - Precondition: `index >= currentPosition`
+  mutating func resume(at index: Input.Index) {
+    assert(index >= bounds.lowerBound)
+    assert(index <= bounds.upperBound)
+    assert(index >= currentPosition)
```
milseman, line 127: Why is this assertion here?

PR #480, opened 2022-06-13
Remove (?X) and (?u) for now

Azoy:
Remove these flags for now in the regex literal parsing until we hone down on the exact semantics for them.

Resolves: rdar://94347937


Azoy:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -767,8 +767,10 @@ extension Source {
         return opt
 
       // Swift semantic level options
-      case "X": return advanceAndReturn(.graphemeClusterSemantics)
-      case "u": return advanceAndReturn(.unicodeScalarSemantics)
+      // These are disabled for now until we figure out the right semantics for
+      // them
+      // case "X": return advanceAndReturn(.graphemeClusterSemantics)
+      // case "u": return advanceAndReturn(.unicodeScalarSemantics)
```
hamishknight, line 773: We could still parse them, but mark them unsupported in Sema (see `validateMatchingOption`), what do you think?

hamishknight (approved):

Azoy:
@swift-ci please test

PR #481, opened 2022-06-13
Recover from parser errors

### LINKED ISSUE #449: Better parser recovery
Currently we `throw` parser errors. While this is very convenient, we need to start recovering where we can, and still produce an AST.

### LINKED ISSUE #472: `\o` crashes the parser
`/\o/` hits a fatal error in the parser

hamishknight:
Currently we use Swift error handling for parser errors. While this is convenient, it has a number of drawbacks:

- Any AST parsed gets thrown away as soon as we encounter an error. This prevents clients from being able to get any useful information from invalid AST (rdar://93677069).

- Multiple diagnostics cannot be issued, meaning that e.g a basic syntactic error could obscure a more useful semantic error.

- It doesn't extend nicely to e.g warning diagnostics, meaning that we'd eventually end up with 2 ways of emitting diagnostics.

- The thrown errors relied on `recordLoc` blocks to annotate them with source location info, which could lead to errors without location info if we forgot to add the appropriate `recordLoc` calls. Additionally, in some cases we want a more fine grained location info than the block would give us.

Therefore this PR removes the use of Swift error handling throughout the parser. The parser is now a total function that _always_ returns an AST. If errors are encountered while parsing, they are recorded, and are attached to the resulting AST by the parser. The parser attempts to recover as much of the AST it can when encountering an error. As such, there is now are now `.invalid` atom and character property kinds. Sema then runs and can attach more diagnostics onto the AST.

For now, the compiler interface remains the same, and we pick a single error to `throw`, but this will be changed in a later PR to allow multiple errors and warnings, as well as AST recovery. This also means we can better preserve the capture type in the presence of parser errors.

Fortunately, in most cases, this is quite a mechanical transformation. It entails:

- Moving the lexical analysis methods onto the `Parser`. We were already passing `ParsingContext` parameters for most of them, so it's not clear they were benefitting from the isolation that `Source` offered. Effectively this means that all parsing has access to the context and diagnostics.

- Converting error throwing statements into calls to the parser's `error` method (or `unreachable` method for unreachables).

This PR also updates the parser tests to be able to be able to match against multiple diagnostics.

Part of the fix for rdar://93677069
Resolves #449

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
@@ -253,18 +253,18 @@ extension Source {
   ///
   /// Throws on overflow
   ///
-  private mutating func lexNumber<Num: FixedWidthInteger>(
-    _ ty: Num.Type, _ kind: RadixKind
-  ) throws -> Located<Num>? {
+  private mutating func lexNumber(
+    _ kind: RadixKind
+  ) throws -> AST.Atom.Number? {
     try recordLoc { src in
-      guard let str = src.tryEatPrefix(kind.characterFilter)?.string else {
+      guard let str = src.tryEatLocatedPrefix(kind.characterFilter) else {
         return nil
       }
-      guard let i = Num(str, radix: kind.radix) else {
-        throw ParseError.numberOverflow(str)
+      guard let i = Int(str.value, radix: kind.radix) else {
+        throw ParseError.numberOverflow(str.value)
       }
-      return i
-    }
+      return .init(i, at: str.location)
+    }.value
   }
 
   /// Try to eat a number off the front.
@@ -273,11 +273,11 @@ extension Source {
   ///
   /// Throws on overflow
   ///
-  mutating func lexNumber() throws -> Located<Int>? {
-    try lexNumber(Int.self, .decimal)
+  mutating func lexNumber() throws -> AST.Atom.Number? {
+    try lexNumber(.decimal)
   }
 
-  mutating func expectNumber() throws -> Located<Int> {
+  mutating func expectNumber() throws -> AST.Atom.Number {
     guard let num = try lexNumber() else {
       throw ParseError.expectedNumber("", kind: .decimal)
     }
@@ -488,9 +488,10 @@ extension Source {
 
         if let t = src.lexWhitespace() { trivia.append(t) }
 
-        let upperOpt = try src.lexNumber()?.map { upper in
+        var upperOpt = try src.lexNumber()
+        if closedRange == false {
           // If we have an open range, the upper bound should be adjusted down.
-          closedRange == true ? upper : upper - 1
+          upperOpt?.value? -= 1
         }
 
         if let t = src.lexWhitespace() { trivia.append(t) }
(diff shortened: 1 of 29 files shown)


Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -16,178 +16,259 @@ concerns upon request.
...
-  func lookahead<T>(_ body: (inout Source) throws -> T) rethrows -> T {
-    var src = self
-    return try body(&src)
+  /// Eat a character from the input, emitting an unreachable error if the end
+  /// has been reached.
+  mutating func eat() -> Char {
+    guard let char = tryEat() else {
+      unreachable("Cannot eat at the end!")
+      return "\u{FFFD}" // U+FFFD Replacement Character (�)
+    }
```
hamishknight, line 175: @milseman We used to `assert` and `unsafelyUnwrapped` here. In the spirit of trying not to crash where we can (ala `throw Unreachable`), I changed it to record an `unreachable` error and return U+FFFD. Does that make sense?

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
milseman: Would the caller just crash downstream, or what would happen?

hamishknight: It would be up to the client, it gets added as a `.fatalError` to the diagnostics, and e.g for the existing clients that expect an error to be thrown (such as the runtime `Regex` initializer), it gets materialized as a thrown error (similar to `Unreachable`). It could be argued that the compiler ought to crash if it gets handed back a `.fatalError`, though we could also emit a `"please file a bug report"` diagnostic.

milseman: AFAICT, the call to `eat()` is guarded by a source not empty, so it could be changed to just call `tryEat()`. The site that uses `eatWithLoc()` seems like returning `U+FFFD` would fall through to a default switch case, but arguably should be properly handled.

What do you think about just removing `eat()`? Callers can use `tryEat()` and if they want a substitute they can supply it explicitly.

hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
hamishknight: > What do you think about just removing `eat()`? Callers can use `tryEat()` and if they want a substitute they can supply it explicitly.

Seems reasonable to me! Updated to remove it

milseman:
We'll want to figure out if we want to land this in a feature branch in the short term or if we'll try to get this into 5.7

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

PR #482, opened 2022-06-14
Remove `buildEither`.

rxwei:
`buildEither` was removed from the regex builder DSL proposal. See apple/swift-evolution#1634.


rxwei:
@swift-ci please test

Azoy (approved):

PR #484, opened 2022-06-15
Allow scoped removal of extended syntax in multi-line literals

hamishknight:
Relax the ban on unsetting extended syntax in a multi-line literal such that it does not apply to a scoped unset e.g `(?-x:...)`, as long as it does not span multiple lines.

This PR also bans the use of `(?^)` in a multi-line literal, unless it is scoped and does not span multiple lines. Instead, `(?^x)` must be written, as PCRE defines `(?^)` to be equivalent to `(?-imnsx)`.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
@@ -597,7 +597,7 @@ extension Source {
         }.value
 
         // In multi-line literals, the quote may not span multiple lines.
-        if context.syntax.contains(.multilineExtendedSyntax),
+        if context.syntax.contains(.multilineCompilerLiteral),
             contents.spansMultipleLinesInRegexLiteral {
           throw ParseError.quoteMayNotSpanMultipleLines
         }
@@ -841,7 +841,7 @@ extension Source {
           throw ParseError.cannotRemoveSemanticsOptions
         }
         // Extended syntax may not be removed if in multi-line mode.
-        if context.syntax.contains(.multilineExtendedSyntax) &&
+        if context.syntax.contains(.multilineCompilerLiteral) &&
             opt.isAnyExtended {
           throw ParseError.cannotRemoveExtendedSyntaxInMultilineMode
         }
--- a/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
+++ b/Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
@@ -840,11 +840,6 @@ extension Source {
         if opt.isSemanticMatchingLevel {
           throw ParseError.cannotRemoveSemanticsOptions
         }
-        // Extended syntax may not be removed if in multi-line mode.
-        if context.syntax.contains(.multilineCompilerLiteral) &&
-            opt.isAnyExtended {
-          throw ParseError.cannotRemoveExtendedSyntaxInMultilineMode
-        }
         removing.append(opt)
       }
       return .init(caretLoc: nil, adding: adding, minusLoc: ateMinus.location,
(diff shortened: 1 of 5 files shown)


hamishknight:
@swift-ci please test

Sources/_RegexParser/Regex/Parse/Parse.swift
```diff
@@ -289,12 +289,16 @@ extension Parser {
...
+  ) throws {
     func mapOption(_ option: SyntaxOptions,
                    _ pred: (AST.MatchingOption) -> Bool) {
       if opts.resetsCurrentOptions {
-        context.syntax.remove(option)
+        // Extended syntax is never reset in a multi-line literal.
+        if !(option == .extendedSyntax &&
+             context.syntax.contains(.multilineLiteral)) {
+          context.syntax.remove(option)
+        }
```
hamishknight, line 301: @milseman Just to double check, `(?^:...)` shouldn't unset extended syntax in a multi-line literal? This is consistent with the behavior of `(?^)` in a multi-line literal, as otherwise we'd need to ban it outright.

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
```diff
@@ -597,7 +597,7 @@ extension Source {
         }.value
 
         // In multi-line literals, the quote may not span multiple lines.
-        if context.syntax.contains(.multilineExtendedSyntax),
+        if context.syntax.contains(.multilineLiteral),
```
milseman, line 600: Just to double check. There's a difference between a multi-line source string containing regex syntax and a multi-line literal. The former is fine, it's just that the lines in a literal are treated differently

Sources/_RegexParser/Regex/Parse/Parse.swift
milseman: If we can, I think `(?^)` meaning reset to default (and the default is determined by literal / run time string vs multi-line literal) would be best

Sources/_RegexParser/Regex/Parse/SyntaxOptions.swift
```diff
@@ -58,9 +58,14 @@ public struct SyntaxOptions: OptionSet {
   /// `(_: .*)` == `(?:.*)`
   public static var experimentalCaptures: Self { Self(1 << 5) }
 
+  /// The syntax kind of a multi-line literal. This will always be set when
+  /// parsing a multi-line literal. Note this does not imply extended syntax,
+  /// as that may be temporarily disabled while parsing.
+  public static var multilineLiteral: Self { Self(1 << 6) }
```
milseman, line 64: This also cannot be un-set and there's no syntax for setting it in the regex. I wonder if it should be a let on the parser or something like that instead of an option. Not sure, thoughts?

Sources/_RegexParser/Regex/Parse/Parse.swift
hamishknight: We talked about this offline and decided to ban `(?^)` outright as it's specific to PCRE and is defined to be equivalent to `(?-imnsx)`

Sources/_RegexParser/Regex/Parse/LexicalAnalysis.swift
hamishknight: Indeed, and I've renamed `multilineLiteral` to `multilineCompilerLiteral`

Sources/_RegexParser/Regex/Parse/SyntaxOptions.swift
hamishknight: Talked offline, and decided to keep this as-is for now

milseman (approved):
LGTM

Issue #483, opened 2022-06-15
[Benchmark] Excessive (and expensive) backtracking


milseman:
From [the forums](https://forums.swift.org/t/performance-of-the-new-regex-in-swift-5-7-degrades-quadratically-without-need/58122):

```swift
let pattern = "^ +A"
let regex1 = try! NSRegularExpression(pattern: pattern)
let regex2 = try! Regex(pattern)
let numberOfSpaces = 50000
let testString = String(repeating: " ", count: numberOfSpaces)
let date = Date()
print(regex1.firstMatch(in: testString, range: .init(location: 0, length: numberOfSpaces)) == nil)
print(-date.timeIntervalSinceNow)
print(try! regex2.firstMatch(in: testString) == nil)
print(-date.timeIntervalSinceNow)
```

milseman:
@rctcwyvrn, can you add this to the benchmark suite you're working on? Lots of opportunities exposed here.

We don't have a compiler IR yet, so we don't want to layer on excess complexity while the execution model is still settling down. But, there's _a ton_ of low-hanging fruit here.

---

`Processor.consume()` calculates the remaining size of the input slice. Switching to `index(offsetBy:limitedBy:)` removes this extra linear measurement, which is what was creating the quadratic factor. Changing this gave me a very easy 7x improvement for this specific benchmark (locally, via package build), which would only improve with input length.

@Azoy, can you try to spin up a PR incorporating [this patch](https://github.com/milseman/swift-experimental-string-processing/commit/298ec2fac9bd8a3af76c6c2131b6f408d047e5fd) and make sure we're test the boundaries correctly?


---

`firstMatch` is needlessly spinning up a new executor and processor for every single input position, only to fail right away on the anchor. We should integrate first-match with the processor so it can just reset its state.

[This patch](https://github.com/milseman/swift-experimental-string-processing/commit/33187351d37e7fcec52b28feb97f822bda4a36b1) gives an an additional 10% improvement on this after the 7x from above from re-using the same executor. It still creates a new processor every time though, avoiding that would give us another 30% beyond the above.

@rctcwyvrn Can you work with me on integrating it? Can you also work with me on crafting targeted benchmarks that demonstrates these prominently, as well as reusing the same processor?


---

After that, we're getting absolutely killed by ARC, which accounts for ~70% of the remaining time after the above improvements. 

[...]

PR #489, opened 2022-06-15
Reuse the executor in firstMatch

rctcwyvrn:
Reset the processor and it's registers instead of recreating them from scratch each iteration in `firstMatch`

Using this benchmark https://github.com/rctcwyvrn/swift-experimental-string-processing/blob/lily-benchmarker/Sources/RegexBenchmark/Suite/FirstMatch.swift we get a pretty big speedup, 60ms to 10ms on my machine

There isn't much benefit in the original test case from https://github.com/apple/swift-experimental-string-processing/issues/483 (BasicBacktrack in the suite) but maybe it'll be more pronounced once the other patch mentioned in that issue is implemented. There is also some slowdown in the other benchmarks and I'm not sure why. Since `wholeMatch` calls into `_firstMatch` we also see some speedup in the css benchmark

Before
``` 
Running
- ReluctantQuant 36.4ms
- ReluctantQuantWithTerminal 151ms
- EagarQuantWithTerminal 37.8ms
- BasicBacktrack 299ms
- BasicBacktrackNS 617µs
- BasicBacktrackFirstMatch 297ms
- BasicBacktrackNSFirstMatch 598µs
- cssRegex 174ms
- cssRegexNS 1.02ms
- FirstMatch 61.4ms
- FirstMatchNS 256µs
- AllMatches 295ms
- AllMatchesNS 819µs
```

After
```
Running
- ReluctantQuant 44.6ms
- ReluctantQuantWithTerminal 158ms
- EagarQuantWithTerminal 44.3ms
- BasicBacktrack 292ms
- BasicBacktrackNS 598µs
- BasicBacktrackFirstMatch 292ms
- BasicBacktrackNSFirstMatch 599µs
- cssRegex 155ms
- cssRegexNS 1.02ms
- FirstMatch 10.5ms
- FirstMatchNS 254µs
- AllMatches 296ms
- AllMatchesNS 813µs

### DIFF
--- a/Sources/_StringProcessing/Engine/Processor.swift
+++ b/Sources/_StringProcessing/Engine/Processor.swift
@@ -32,7 +32,7 @@ struct Processor<
   typealias Element = Input.Element
 
   let input: Input
-  var bounds: Range<Position>
+  let bounds: Range<Position>
   let matchMode: MatchMode
   var currentPosition: Position
 
@@ -95,19 +95,18 @@ extension Processor {
     assert(currentPosition <= end)
   }
 
-  mutating func reset(newBounds: Range<Position>)
+  mutating func reset(newPosition: Position)
   {
-    self.bounds = newBounds
-
     self.controller = Controller(pc: 0)
-    self.currentPosition = bounds.lowerBound
+    self.currentPosition = newPosition
 
     self.registers.reset(bounds.upperBound)
-    self.storedCaptures = self.storedCaptures.map {_ in .init()}
+    for idx in storedCaptures.indices {
+      storedCaptures[idx] = .init()
+    }
 
-    self.cycleCount = 0
-    self.savePoints = []
-    self.callStack = []
+    self.savePoints.removeAll(keepingCapacity: true)
+    self.callStack.removeAll(keepingCapacity: true)
     self.state = .inProgress
     self.failureReason = nil
   }
--- a/Sources/_StringProcessing/Engine/Registers.swift
+++ b/Sources/_StringProcessing/Engine/Registers.swift
@@ -162,17 +162,21 @@ extension Processor.Registers {
   mutating func reset(_ sentinel: Input.Index) {
     // note: Is there any issue with the program transform functions holding
     // state and not getting reset here? Do we care?
-    self.bools = Array(repeating: false, count: info.bools)
-    self.ints = Array(repeating: 0, count: info.ints)
-    self.floats = Array(repeating: 0, count: info.floats)
-    self.positions = Array(repeating: sentinel, count: info.positions)
-    self.values = Array(
-      repeating: SentinelValue(), count: info.values)
+    func clear<T>(_ xs: inout [T], _ v: T) {
+      for idx in xs.indices {
+        xs[idx] = v
+      }
+    }
 
-    self.instructionAddresses = Array(repeating: 0, count: info.instructionAddresses)
-    self.classStackAddresses = Array(repeating: 0, count: info.classStackAddresses)
-    self.positionStackAddresses = Array(repeating: 0, count: info.positionStackAddresses)
-    self.savePointAddresses = Array(repeating: 0, count: info.savePointAddresses)
+    clear(&self.bools, false)
+    clear(&self.ints, 0)
+    clear(&self.floats, 0)
+    clear(&self.positions, sentinel)
+    clear(&self.values, SentinelValue())
+    clear(&self.instructionAddresses, 0)
+    clear(&self.classStackAddresses, 0)
+    clear(&self.positionStackAddresses, 0)
+    clear(&self.savePointAddresses, 0)
   }
 }
 
--- a/Sources/_StringProcessing/Executor.swift
+++ b/Sources/_StringProcessing/Executor.swift
@@ -27,7 +27,7 @@ struct Executor {
   ) throws -> Regex<Output>.Match? {
     var cpu = engine.makeProcessor(
       input: input, bounds: inputRange, matchMode: mode)
-    return try consume(input, &cpu)
+    return try consume(input, &cpu, startingFrom: inputRange.lowerBound)
   }
   
   @available(SwiftStdlib 5.7, *)
@@ -43,7 +43,11 @@ struct Executor {
       input: input, bounds: inputRange, matchMode: mode)
 
     while true {
-      if let m: Regex<Output>.Match = try consume(input, &cpu) {
+      if let m: Regex<Output>.Match = try consume(
+        input,
+        &cpu,
+        startingFrom: low
+      ) {
         return m
       }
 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by rctcwyvrn** [comment]
Looking at `ReluctantQuant` the regression seems to be due to an extra `memmove` in `Processor.cycle()`. Not sure what's causing this extra move though

Before
```
3.21 s   70.1%	71.00 ms	 	             specialized Processor.cycle()
1.13 s   24.7%	32.00 ms	 	              closure #1 in Compiler.ByteCodeGen.emitAny()
512.00 ms   11.1%	378.00 ms	 	              specialized Processor.cycle()
426.00 ms    9.3%	426.00 ms	 	              swift_release
```

After
```
4.08 s   73.1%	86.00 ms	 	              specialized Processor.cycle()
1.15 s   20.6%	25.00 ms	 	               closure #1 in Compiler.ByteCodeGen.emitAny()
736.00 ms   13.1%	736.00 ms	 	               _platform_memmove
439.00 ms    7.8%	439.00 ms	 	               swift_release
420.00 ms    7.5%	244.00 ms	 	               specialized Processor.cycle()
```

File: Sources/_StringProcessing/Engine/Processor.swift [hunk 1]
```diff
@@ -32,7 +32,7 @@ struct Processor<
   typealias Element = Input.Element
 
   let input: Input
-  let bounds: Range<Position>
+  var bounds: Range<Position>
```
milseman, line 35: We have a bug or series of bugs that @natecook1000 is working on, but we probably won't want to be changing the bounds here.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -94,6 +94,23 @@ extension Processor {
...
+
+  mutating func reset(newBounds: Range<Position>)
+  {
+    self.bounds = newBounds
+
+    self.controller = Controller(pc: 0)
+    self.currentPosition = bounds.lowerBound
+
+    self.registers.reset(bounds.upperBound)
+    self.storedCaptures = self.storedCaptures.map {_ in .init()}
```
milseman, line 106: The perf win of reset comes from re-using the same arrays instead of allocating new ones.

```suggestion
    for idx in self.storedCaptures.indices {
      self.storedCaptures[idx] = .init()
    }
```

(@natecook1000 is there an algorithm for this?)

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -94,6 +94,23 @@ extension Processor {
...
+
+    self.cycleCount = 0
```
milseman, line 108:
```suggestion
```

I'd say leave it  alone for now, the engine can query it. But we will probably want to keep metrics soon, including counting save points and restores, etc.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -94,6 +94,23 @@ extension Processor {
...
+    self.savePoints = []
+    self.callStack = []
```
milseman, line 110:
```suggestion
    self.callStack.removeAll(keepingCapacity: true)
    self.savePoints.removeAll(keepingCapacity: true)
```

Sources/_StringProcessing/Engine/Registers.swift
```diff
@@ -156,6 +158,22 @@ extension Processor.Registers {
...
+    self.floats = Array(repeating: 0, count: info.floats)
+    self.positions = Array(repeating: sentinel, count: info.positions)
+    self.values = Array(
+      repeating: SentinelValue(), count: info.values)
+
+    self.instructionAddresses = Array(repeating: 0, count: info.instructionAddresses)
+    self.classStackAddresses = Array(repeating: 0, count: info.classStackAddresses)
+    self.positionStackAddresses = Array(repeating: 0, count: info.positionStackAddresses)
+    self.savePointAddresses = Array(repeating: 0, count: info.savePointAddresses)
+  }
```
milseman, line 176: Just clear these instead of allocating new arrays. Making new arrays is what makes creating a new processor slow.

Sources/_StringProcessing/Engine/Processor.swift
rctcwyvrn: So we're fine with leaving the lower bound at the beginning of the string? I guess it still makes sense as the bounds of the string and we just need to set `currentPosition` to be where we want to start from when we reset?

Sources/_StringProcessing/Engine/Processor.swift
natecook1000: No, we've had a mutating `forEach` pitched several times, but it's never been added to the stdlib or `swift-algorithms`.

Sources/_StringProcessing/Engine/Registers.swift
```diff
@@ -162,17 +162,21 @@ extension Processor.Registers {
...
-    self.ints = Array(repeating: 0, count: info.ints)
-    self.floats = Array(repeating: 0, count: info.floats)
-    self.positions = Array(repeating: sentinel, count: info.positions)
-    self.values = Array(
-      repeating: SentinelValue(), count: info.values)
+    func clear<T>(_ xs: inout [T], _ v: T) {
+      for idx in xs.indices {
+        xs[idx] = v
+      }
+    }
```
milseman, line 169: Or as an extension on `MutableCollection`. @natecook1000 does such an algorithm exist?

Sources/_StringProcessing/Executor.swift
```diff
@@ -43,7 +43,11 @@ struct Executor {
       input: input, bounds: inputRange, matchMode: mode)
 
     while true {
-      if let m: Regex<Output>.Match = try consume(input, &cpu) {
+      if let m: Regex<Output>.Match = try consume(
+        input,
+        &cpu,
+        startingFrom: low
```
milseman, line 49:
```suggestion
        input, &cpu, startingFrom: low
```

milseman:
The memmoves come from having to re-grow the save point stack, etc. Have you tried measuring now that we don't make new arrays?

Sources/_StringProcessing/Engine/Registers.swift
natecook1000: We could try these out here before adding them to swift-algorithms:

```swift
extension MutableCollection {
  mutating func setAll(to element: Element) {
    self.withEach { $0 = element }
  }
  
  mutating func withEach(_ body: (inout Element) throws -> Void) rethrows {
    var i = startIndex
    while i < endIndex {
      try body(&self[i])
      formIndex(after: &i)
    }
  }
}

// usage:
self.bools.setAll(to: false)
```

Sources/_StringProcessing/Engine/Processor.swift
milseman: Right, though it might temporarily break some `firstMatch` tests until some anchors get fixed. @natecook1000 can you work with Lily here?

Sources/_StringProcessing/Engine/Registers.swift
milseman: Also considering an underscore them so that no one accidentally ships it 😅

PR #490, opened 2022-06-15
Add tests for substring / anchor interaction

natecook1000:
`^` and `$` should match the start and end of the callee, even if that callee is a substring. Right now `^` and `$` match the start and end of the callee's base string, instead. In addition, `^` and `$` should only match the start and end of the callee when replacing a subrange, not the start and end of the subrange.


milseman (approved):
LGTM

natecook1000:
@swift-ci Please test

PR #491, opened 2022-06-15
Add regex benchmarker

rctcwyvrn:
Simple benchmarking setup reusing the timing code from `swift-collections-benchmark`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,7 +41,10 @@ let package = Package(
             targets: ["_RegexParser"]),
         .executable(
             name: "VariadicsGenerator",
-            targets: ["VariadicsGenerator"])
+            targets: ["VariadicsGenerator"]),
+        .executable(
+            name: "RegexBenchmark",
+            targets: ["RegexBenchmark"])
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.0.0"),
@@ -111,6 +114,17 @@ let package = Package(
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 "_RegexParser",
                 "_StringProcessing"
+            ]),
+        .executableTarget(
+            name: "RegexBenchmark",
+            dependencies: [
+                .product(name: "ArgumentParser", package: "swift-argument-parser"),
+                "_RegexParser",
+                "_StringProcessing",
+                "RegexBuilder"
+            ],
+            swiftSettings: [
+                .unsafeFlags(["-Xfrontend", "-disable-availability-checking"]),
             ]),
 
         // MARK: Exercises
--- a/Sources/RegexBenchmark/Benchmark.swift
+++ b/Sources/RegexBenchmark/Benchmark.swift
@@ -15,13 +15,13 @@ public struct Benchmark: RegexBenchmark {
   public enum MatchType {
     case whole
     case first
-    case enumerate
+    case allMatches
   }
   
   public func run() {
     switch ty {
     case .whole: blackHole(target.wholeMatch(of: regex))
-    case .enumerate: blackHole(target.matches(of: regex))
+    case .allMatches: blackHole(target.matches(of: regex))
     case .first: blackHole(target.firstMatch(of: regex))
     }
   }
@@ -54,14 +54,14 @@ public struct BenchmarkRunner {
   // Register instances of Benchmark and run them
   let suiteName: String
   var suite: [any RegexBenchmark]
-  let samples: Int = 40
+  let samples: Int = 20
   
   public init(suiteName: String) {
     self.suiteName = suiteName
     self.suite = []
   }
 
-  public mutating func register(new: some RegexBenchmark) {
+  public mutating func register(_ new: some RegexBenchmark) {
     suite.append(new)
   }
   
--- a/Sources/RegexBenchmark/Benchmark.swift
+++ b/Sources/RegexBenchmark/Benchmark.swift
@@ -54,7 +54,7 @@ public struct BenchmarkRunner {
   // Register instances of Benchmark and run them
   let suiteName: String
   var suite: [any RegexBenchmark]
-  let samples: Int = 20
+  let samples: Int = 40
   
   public init(suiteName: String) {
     self.suiteName = suiteName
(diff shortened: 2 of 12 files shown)


lorentey:
It may make sense to move the benchmark target out of the top-level package and into a hidden sub-package in a subdirectory, like with swift-collections.

This will let you be more flexible about bringing in extra dependencies for benchmarking without burdening the top-level package with them.

Granted, this isn't a package that is designed to be a dependency of anything else, so it may not matter that much!

Sources/RegexBenchmark/Benchmark.swift
```diff
@@ -0,0 +1,116 @@
...
+
+  public mutating func register(_ new: some RegexBenchmark) {
+    suite.append(new)
+  }
+  
+  func measure(benchmark: some RegexBenchmark) -> Time {
+    var times: [Time] = []
+    
+    // initial run to make sure the regex has been compiled
+    benchmark.run()
```
milseman, line 79: Future work: We'll want to know how much time is spent compiling vs not

Sources/RegexBenchmark/Benchmark.swift
```diff
@@ -0,0 +1,116 @@
...
+      benchmark.run()
+      let end = Tick.now
+      let time = end.elapsedTime(since: start)
+      times.append(time)
+    }
+    // todo: compute stdev and warn if it's too large
+    
+    // return median time
+    times.sort()
+    return times[samples/2]
```
milseman, line 93: Future work: provide the times as a type which the caller can ask for the median of.

Sources/RegexBenchmark/CLI.swift
```diff
@@ -0,0 +1,33 @@
+import ArgumentParser
+
+@main
+struct Runner: ParsableCommand {
+  @Argument(help: "Names of benchmarks to run")
+  var specificBenchmarks: [String] = []
+  
+  @Option(help: "Run only once for profiling purposes")
+  var profile = false
```
milseman, line 9: Hmm.... Sometimes it helps profiling to run many samples since you'll see what warm behavior looks like and the hot parts become more pronounced. What's the difference between this flag and a sample count of 1?

Sources/RegexBenchmark/CLI.swift
```diff
@@ -0,0 +1,33 @@
...
+  @Option(help: "How many samples to collect for each benchmark")
+  var samples = 20
+    
+  func makeRunner() -> BenchmarkRunner {
+    var benchmark = BenchmarkRunner("RegexBench", samples)
+    benchmark.addReluctantQuant()
+    benchmark.addBacktracking()
+    benchmark.addCSS()
+    benchmark.addFirstMatch()
+    return benchmark
```
milseman, line 20: Could we make it an array or some other way to make registration easier? Seems like we could then filter that array when creating the runner

Sources/RegexBenchmark/Suite/Backtracking.swift
```diff
@@ -0,0 +1,45 @@
...
+
+extension BenchmarkRunner {
+  mutating func addBacktracking() {
+    let r = "^ +A"
+    let s = String(repeating: " ", count: 10000)
+
+    let basicBacktrack = Benchmark(
+      name: "BasicBacktrack",
+      regex: try! Regex(r),
+      ty: .allMatches,
```
milseman, line 15: Go ahead and give the `ty` argument label a full name. That is ok as a argument name, but the label should clarify the use site.

Sources/RegexBenchmark/Suite/Backtracking.swift
```diff
@@ -0,0 +1,45 @@
...
+      target: s
+    )
+
+    let basicBacktrackNSFirstMatch = NSBenchmark(
+      name: "BasicBacktrackNSFirstMatch",
+      regex: try! NSRegularExpression(pattern: r),
+      ty: .first,
+      target: s
+    )
+
```
milseman, line 39: Hmm... we want whole-match benchmarks, match-from-front benchmarks, first-match, and all-matches (which is repeated first-match calls). We also want to make the NSRegularExpression equivalent of each, at least if there can be an equivalent.

Package.swift
```diff
@@ -112,6 +115,17 @@ let package = Package(
                 "_RegexParser",
                 "_StringProcessing"
             ]),
+        .executableTarget(
+            name: "RegexBenchmark",
+            dependencies: [
+                .product(name: "ArgumentParser", package: "swift-argument-parser"),
+                "_RegexParser",
```
rxwei, line 122: I don't think you need this dependency

Sources/RegexBenchmark/Suite/CssRegex.swift
```diff
@@ -0,0 +1,1672 @@
+import Foundation
+
+extension BenchmarkRunner {
+  mutating func addCSS() {
+    let r = "--([a-zA-Z0-9_-]+)\\s*:\\s*(.*?):"
+    
+    // sorry
+    let css = """
```
milseman, line 8: Move this string into its own file and give it a better internal name (or make it a static var on a CSS benchmark type). We could have an `Inputs` folder for these kinds of things

Sources/RegexBenchmark/Suite/ReluctantQuant.swift
```diff
@@ -0,0 +1,42 @@
+import _StringProcessing
+import RegexBuilder
+
+extension BenchmarkRunner {
+  mutating func addReluctantQuant() {
```
milseman, line 5: The regex can be `.*?` so that we can compare to NSRegularExpression

Sources/RegexBenchmark/Suite/ReluctantQuant.swift
```diff
@@ -0,0 +1,42 @@
...
+    let size = 500000
+    let s = String(repeating: "a", count: size)
```
milseman, line 7: Future work: These are very, very micro (nano?) and would be trivialized through some analysis and optimizations that wouldn't necessarily improve the state of real regexes or components of real regexes. I wonder if we should distinguish these from more realistic ones.

milseman (approved):
Let's get this in asap and we'll fix things in a follow up PR

PR #496, opened 2022-06-17
Disentangle disparate 'bounds' ideas in processor

natecook1000:
This starts the separation of two different ideas for boundaries in the base input:

- subjectBounds: These represent the actual subject in the input   string. For a `String` callee, this will cover the entire bounds, while for a `Substring` these will represent the bounds of the substring in the base.
- searchBounds: These represent the current search range in the subject. These bounds can be the same as `subjectBounds` or a subrange when searching for subsequent matches or replacing only in a subrange of a string.


natecook1000:
@swift-ci Please test

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -97,25 +97,25 @@ fileprivate extension Compiler.ByteCodeGen {
     switch kind {
     case .startOfSubject:
       builder.buildAssert { (input, pos, bounds) in
-        pos == input.startIndex
+        pos == bounds.lowerBound
```
milseman, line 100: Is the bounds here the search bounds or subject bounds?

milseman: Also, FWIW, I'm going to try to have @rctcwyvrn make a builtin assertion and/or character class instruction, so all of these closures will just be switch case bodies in the matching engine fairly soon.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -141,9 +142,10 @@ fileprivate extension Compiler.ByteCodeGen {
       }
 
     case .startOfLine:
+      // FIXME: Anchor.startOfLine must always use this first branch
```
milseman, line 145: What does this mean?

For the builtin instruction, I think we'd represent each distinct variation as a separate enum value, so compilation would just be picking the right operand to OR in.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -31,18 +31,36 @@ struct Processor<
...
+  ///
+  /// `subjectBounds` is equal to or a subrange of
+  /// `input.startIndex..<input.endIndex`.
+  let subjectBounds: Range<Position>
+  
+  /// The bounds within the subject for an individual search.
+  ///
+  /// `searchBounds` is equal to `subjectBounds` in some cases, but can be a
+  /// subrange when performing operations like searching for matches iteratively
+  /// or calling `str.replacing(_:with:subrange:)`.
```
milseman, line 52: Maybe a note about why it matters, i.e. the behavior of anchors

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -411,9 +426,9 @@ extension Processor {
 
     case .consumeBy:
       let reg = payload.consumer
-      guard currentPosition < bounds.upperBound,
+      guard currentPosition < subjectBounds.upperBound,
```
milseman, line 429: Subject or search bounds?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -31,18 +31,36 @@ struct Processor<
...
-  //
-  // FIXME: This also conflates search bounds too!
-  var bounds: Range<Position>
-
-  // The current position in the subject
+  
+  /// The current search position while processing.
+  ///
+  /// `currentPosition` must always be in the range `subjectBounds` or equal
+  /// to `subjectBounds.upperBound`.
```
milseman, line 63: Can you add the assertion that it is a subrange to `_checkInvariants()`?

Sources/_StringProcessing/Engine/Processor.swift
milseman: How do we, generally, know which one to choose?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -59,53 +77,50 @@ struct Processor<
...
     _checkInvariants()
   }
 
-
-  mutating func reset(searchBounds: Range<Position>) {
-    // FIXME: We currently conflate both subject bounds and search bounds
-    // This should just reset search bounds
-    self.bounds = searchBounds
-    self.currentPosition = self.bounds.lowerBound
+  mutating func reset(start: Position, end: Position) {
```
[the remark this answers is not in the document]
milseman: Do we need another notion of resetting for the purposes of an engine-baked all matches?

Tests/RegexTests/MatchTests.swift
```diff
@@ -730,11 +700,11 @@ extension RegexTests {
     firstMatchTest(#"[\Qa-c\E]+"#, input: "a-c", match: "a-c")
 
     firstMatchTest(#"["a-c"]+"#, input: "abc", match: "a",
-                   syntax: .experimental)
+                   syntax: .experimental, xfail: true)
     firstMatchTest(#"["abc"]+"#, input: "cba", match: "cba",
                    syntax: .experimental)
     firstMatchTest(#"["abc"]+"#, input: #""abc""#, match: "abc",
-                   syntax: .experimental)
+                   syntax: .experimental, xfail: true)
```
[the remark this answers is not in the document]
milseman: It's the same as the above, but we can also remove them. The quotes are just prettier `\Q...\E`

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -185,14 +185,26 @@ extension BidirectionalCollection {
 
 @available(SwiftStdlib 5.7, *)
 struct RegexMatchesCollection<Output> {
-  let input: Substring
+  let input: String
```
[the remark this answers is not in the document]
natecook1000: I found it a lot clearer to separate them out as early as possible — otherwise you can have internal methods calling into others with things like `input[start...]`, which adjusts the wrong boundary.

Sources/_StringProcessing/ByteCodeGen.swift
natecook1000: It's another regex syntax / DSL syntax discrepancy. Right now, `^` and `Anchor.startOfLine` both get written into the tree as `AST.AssertionKind.startOfLine`, but `^` actually can mean either "start of line" or "start of subject", depending on the current options settings. `Anchor.startOfLine` should always mean what it says, so we need to fix that.

Sources/_StringProcessing/ByteCodeGen.swift
milseman: Let's do the built in assertion change first (as in soon)

Sources/_StringProcessing/ByteCodeGen.swift
milseman: We will need a flattened enum post-option-application for the engine. That might also help us understand the model better.

Sources/_StringProcessing/Engine/Processor.swift
natecook1000: Generally, we want to be moving and matching with `searchBounds`; `subjectBounds` should really only be used to match anchors, not for general purpose index movements.

natecook1000: (i.e. this was incorrect)

Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift
```diff
@@ -74,6 +74,6 @@ extension BidirectionalCollection where SubSequence == Substring {
   @_disfavoredOverload
   @available(SwiftStdlib 5.7, *)
   public func contains(_ regex: some RegexComponent) -> Bool {
-    _contains(RegexConsumer(regex))
+    (try? regex.regex.firstMatch(in: self[...])) != nil
```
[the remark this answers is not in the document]
natecook1000: I've been trying to use the `Regex` APIs for all of these internal calls, so we don't have quite as serpentine a call stack. That said, `contains` should call `firstRange`, so that it gets any optimization we can do around ignoring capture groups.

Sources/_StringProcessing/Engine/Processor.swift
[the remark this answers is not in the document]
natecook1000: I modified `Processor.reset` again to only take the new starting position, since its `searchBounds` stay the same through a single call to `firstMatch`. I think you mentioned elsewhere that we eventually want to use a single processor for all the searches in `"foo".matches(of: regex)` — when we do that we'll need to change this again to allow `searchBounds` to be modified at the start of each successive search.

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Engine/Processor.swift
milseman: Any way we could use access control or some better coding convention here? E.g. if we always want search bounds except for anchors, can we isolate the code that refers to subject bounds inside the engine?

Sources/_StringProcessing/Algorithms/Algorithms/Contains.swift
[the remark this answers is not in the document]
milseman: Are you planning on doing this in this PR or is that blocked on something?

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
```diff
@@ -310,7 +325,11 @@ extension RegexMatchesCollection: Collection {
     
     guard
       let start = searchIndex(after: currentMatch),
-      let nextMatch = try? regex.firstMatch(in: input[start...])
+      start <= searchBounds.upperBound,
+      let nextMatch = try? regex._firstMatch(
+        input,
+        subjectBounds: subjectBounds,
+        searchBounds: start..<searchBounds.upperBound)
```
milseman, line 332: Elsewhere you mentioned adjusting current position instead of search bounds. Is that or does that need to be consistent here? When do we adjust one or the other?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -152,15 +154,16 @@ fileprivate extension Compiler.ByteCodeGen {
...
       } else {
-        builder.buildAssert { (input, pos, bounds) in
-          pos == input.startIndex
+        builder.buildAssert { (input, pos, subjectBounds) in
+          pos == subjectBounds.lowerBound
         }
       }
       
     case .endOfLine:
+      // FIXME: Anchor.endOfLine must always use this first branch
```
milseman, line 163: What does this comment mean? Would it be relevant to someone trying to migrate this code over into the engine itself? How should they migrate it? Any XFAIL tests here?

milseman (approved):
LGTM, some minor cleanups and testing requests.

Sources/_StringProcessing/Algorithms/Matching/Matches.swift
natecook1000: The `searchBounds` can be adjusted here, since it's a brand new search. The place we don't want to adjust `searchBounds` is within `_firstMatch`, since that's just moving forward the potential match's starting point.

Sources/_StringProcessing/ByteCodeGen.swift
natecook1000: I'll rewrite the comment and add an XFAIL.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -158,8 +183,8 @@ extension Processor {
   /// - Precondition: `bounds.contains(index) || index == bounds.upperBound`
   /// - Precondition: `index >= currentPosition`
   mutating func resume(at index: Input.Index) {
-    assert(index >= bounds.lowerBound)
-    assert(index <= bounds.upperBound)
+    assert(index >= searchBounds.lowerBound)
+    assert(index <= searchBounds.upperBound)
```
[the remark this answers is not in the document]
natecook1000: The end is a valid index to resume at, since that's where an end of subject anchor would need to match. An instruction that needs to test for a character at the current index would just need to fail (not crash) at that point. We already have the right behavior in place for these.

natecook1000:
@swift-ci Please test

PR #255, opened 2022-04-06
Remove _Unicode and script data

Azoy:

PR #500, opened 2022-06-18
Fix duplicate symbols static-linking failure by renaming symbols

kateinoigakukun:
`_swift_stdlib_XX` symbols defined in `_CUnicode` module are also defined in runtime library as hidden visibility.
It works in linking as a shared library, but it doesn't when statically linking with lld, which resolves `lib_StringProcessing.a`'s `_swift_stdlib_getScript` and loads `UnicodeScalarProps.c.o`, then loads `libswiftCore.a`'s `UnicodeScalarProps.cpp.o` for other _swift_stdlib_XX symbols that are not defined in UnicodeScalarProps.c.o.

This patch tries to fix the failure by renaming those symbols in _CUnicode module.

XFAILing test has been added by https://github.com/apple/swift/pull/59556, and will be un-XFAILed by https://github.com/apple/swift/pull/59562

### DIFF
--- a/Sources/_CUnicode/UnicodeData.c
+++ b/Sources/_CUnicode/UnicodeData.c
@@ -46,10 +46,10 @@ static uint32_t hash(uint32_t scalar, uint32_t level, uint32_t seed) {
 
 // This implementation is based on the minimal perfect hashing strategy found
 // here: https://arxiv.org/pdf/1702.03154.pdf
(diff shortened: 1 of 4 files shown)


milseman:
@Azoy can you look at this?

Azoy (approved):
This looks good as a workaround until https://github.com/apple/swift-experimental-string-processing/pull/255 is landed. Thanks!

Azoy:
@swift-ci please test

kateinoigakukun:
Thanks!

PR #501, opened 2022-06-20
[benchmark] Simplify and add more benchmarks

milseman:
Add grapheme break benchmarks and adds a little more infrastructure


milseman:
@swift-ci please test

Sources/RegexBenchmark/Suite/CssRegex.swift
```diff
@@ -3,22 +3,12 @@ import _StringProcessing
...
   mutating func addCSS() {
-    let r = "--([a-zA-Z0-9_-]+)\\s*:\\s*(.*?):"
-    
-    let cssRegex = Benchmark(
-      name: "cssRegex",
-      regex: try! Regex(r),
-      ty: .allMatches,
-      target: Inputs.swiftOrgCSS
-    )
+    let r = #"--([a-zA-Z0-9_-]+)\s*:\s*(.*?):"#
```
rctcwyvrn, line 6: first and all are running at the same time because the regex isn't matching anything, there should be a semicolon at the end not a colon 

`let r = #"--([a-zA-Z0-9_-]+)\s*:\s*(.*?);"#`

PR #503, opened 2022-06-20
Make unary builder return `Regex` type consistently

rxwei:
Currently, unary regex component builders simply forward the component's base type. However, this is inconsistent with non-unary builder results. The current behavior may lead to surprising results when the user marks a property with `@RegexComponentBuilder`.

This patch makes `RegexComponentBuilder.buildPartialBlock<R>(first: R)` return a `Regex<R.RegexOutput>` rather than `R` itself.

---

Before:

```swift
// error: cannot convert value of type 'OneOrMore<Substring>' to specified type 'Regex<Substring>'
@RegexComponentBuilder
var r: Regex<Substring> {
  OneOrMore("a")
  // Adding other components below will make the error go away.
}

struct MyCustomRegex: RegexComponent {
  // error: cannot convert value of type 'OneOrMore<Substring>' to specified type 'Regex<Substring>'
  @RegexComponentBuilder
  var regex: Regex<Substring> {
    OneOrMore("a")
  }
}
```

After: No errors.

### DIFF
--- a/Sources/RegexBuilder/Builder.swift
+++ b/Sources/RegexBuilder/Builder.swift
@@ -18,8 +18,10 @@ public enum RegexComponentBuilder {
     .init(node: .empty)
   }
 
-  public static func buildPartialBlock<R: RegexComponent>(first: R ) -> R {
-    first
+  public static func buildPartialBlock<R: RegexComponent>(
+    first component: R
+  ) -> Regex<R.RegexOutput> {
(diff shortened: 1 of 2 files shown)


rxwei:
@swift-ci please test

milseman (approved):
LGTM

rxwei:
@swift-ci please test

PR #504, opened 2022-06-20
Allow `Regex<AnyRegexOutput>` to be used in the DSL.

rxwei:
Apply a workaround to allow `Regex<AnyRegexOutput>` to be used in the DSL. This workaround emits each nested `Regex<AnyRegexOutput>` as a custom matcher so that it's essentially treated as a separate compilation unit.

A proper fix for this is to introduce scoped type erasure in the matching engine so that all type erasure (including the top-level one) goes through this model. I left some stubs in (`beginTypeErase`, `endTypeErase`) which I'll implement in a follow-up.

Since this implementation is using `Executor.match(...) -> Regex.Match` in the regex compiler, we need to add availability annotations to the `Executor` and `Compiler` types.

Resolves rdar://94320030.


milseman:
> Apply a workaround to allow Regex<AnyRegexOutput> to be used in the DSL. This workaround emits each nested Regex<AnyRegexOutput> as a custom matcher so that it's essentially treated as a separate compilation unit.

What does this do to anchors? Do we need to fix up the search bounds?

Sources/_StringProcessing/Engine/Instruction.swift
```diff
@@ -278,6 +278,10 @@ extension Instruction {
     ///
     case backreference
 
+    case beginTypeErase
+
+    case endTypeErase
```
milseman, line 283: What do these instructions do? What's their semantics?

milseman (commented):
What's the intended long-term design? I.e. what are the erasure instructions doing?

Sources/_StringProcessing/Engine/Instruction.swift
rxwei: Add comments. The processor will maintain a stack of capture lists. `beginTypeErase` will push a new list onto the capture stack, so that all captures get added to that list. `endTypeErase(_: ValReg)` will convert all elements of the current capture list to an `AnyRegexOutput`, and moves it to the given value register, and pop the stack.  It's similar to how `matcher` works.

Tests/RegexBuilderTests/RegexDSLTests.swift
```diff
@@ -1071,6 +1071,63 @@ class RegexDSLTests: XCTestCase {
       }
     }
   }
+
+  func testTypeErasedRegexInDSL() throws {
+    do {
+      let input = "johnappleseed: 12."
+      let numberRegex = try! Regex(#"(\d+)\.?"#)
```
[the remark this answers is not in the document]
rxwei: Added one.  The idea is to fix up the search bounds so that `^` refers to the start of the input.

It happens to work today because anchors currently use the base string's bounds, not the input slice's bounds.

rxwei:
> What's the intended long-term design? I.e. what are the erasure instructions doing?

https://github.com/apple/swift-experimental-string-processing/pull/504#discussion_r903132251

Hope this clarifies the intended design. Let me know if you have any feedback.

rxwei:
@swift-ci please test

Sources/_StringProcessing/Engine/Instruction.swift
milseman: I'm not fully parsing this. Could you write some XFAIL tests that illustrate the work that remains to be done after this PR?

rxwei: There's no XFAIL-able tests. This PR fixes the bug, but it's not as efficient as handling this in the bytecode directly. I'd be happy to chat in person also.

milseman:
@swift-ci please test

rxwei:
There's a small bug in there and the fix should be relative simple. I'll take a look tomorrow.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -765,6 +774,28 @@ fileprivate extension Compiler.ByteCodeGen {
...
+      // output type erasure in the matching engine (`beginTypeErase`,
+      // `endTypeErase`).
+      //
+      // Long-term design:
+      //   beginTypeErase
+      //   <code for child>
+      //   endTypeErase
+      let program = try Compiler(tree: DSLTree(child)).emit()
+      let executor = Executor(program: program)
+      return emitMatcher { input, startIndex, range in
```
milseman, line 790: Matchers don't get the subject bounds, so I don't see how this would support anchors that refer to the subject bounds. We _could_ have a different matcher interface or else add subject bounds as an additional parameter for the internal code (probably not API though).

Sources/_StringProcessing/Regex/Core.swift
```diff
@@ -56,7 +56,13 @@ public struct Regex<Output>: RegexComponent {
   }
 
   public var regex: Regex<Output> {
-    self
+    if Output.self == AnyRegexOutput.self {
+      if case .typeErase = root {
+        return self
+      }
+      return .init(node: .typeErase(root))
```
milseman, line 63: Can we do this on creation instead, or somewhere else?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -583,16 +589,28 @@ extension DSLTree.Node {
...
+  @available(SwiftStdlib 5.7, *)
   var captureList: CaptureList {
     var list = CaptureList()
+    // FIXME: This is peering through any top-level `.typeErase`. Once type
+    // erasure was handled in the engine, this can be simplified to using `root`
+    // directly.
+    var root = root
+    while case let .typeErase(child) = root {
+      root = child
+    }
```
milseman, line 613: Why is this scan in here? It seems like you'd want to capture an `ARO`, is that not possible?

Tests/RegexBuilderTests/RegexDSLTests.swift
```diff
@@ -1071,6 +1071,78 @@ class RegexDSLTests: XCTestCase {
...
+      let regex = Regex {
+        Capture {
+          OneOrMore(.word)
+        }
+        ZeroOrMore(.whitespace)
+        ":"
+        ZeroOrMore(.whitespace)
+        Capture { numberRegex }
+      }
+      XCTAssertNil(input.wholeMatch(of: regex))
```
milseman, line 1123: Positive match tests with anchors?

Tests/RegexBuilderTests/RegexDSLTests.swift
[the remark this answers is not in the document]
milseman: Right, there is a bug here that another bug is masking, but when that other bug is fixed, how do we fix this bug?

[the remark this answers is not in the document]
milseman: You can illustrate the difference with a substring input, where the subject bounds are the substring's bounds and the search bounds are contained within.

Sources/_StringProcessing/Engine/Instruction.swift
milseman: How would `begin/endTypeErase` work in the context of backtracking. For that matter, how does this PR support any backtracking into `ARO` at all, i.e. isn't this forcing it to be atomic?

rxwei: I think backtracking into ARO will be supported, since `savePoints` currently stores the capture list. It can be extended to store the capture stack.

PR #509, opened 2022-06-21
Add new benchmarks and benchmarker functionality (try 2)

rctcwyvrn:
Fixes since the last PR

- ~~Use `ISO8601FormatStyle` to fix the linux build failure~~ It turns out that doesn't exist on corelibs-foundation either. It turns out the entire `format` function is missing so work around it with some `#if`
- Remove diff filtering for comparison results (just make sure it's more than a picosecond to remove any diffs from floating point rounding)
- Add a `createBenchmark.py` util script to create benchmarks + register them. After looking around at swift-collection-benchmark and the main swift repo's benchmarks, this seems to be the best way of doing things


rctcwyvrn:
@swift-ci test

rctcwyvrn:
@swift-ci test

Utils/createBenchmark.py
```diff
@@ -0,0 +1,61 @@
+# python3 createBenchmark.py MyRegexBenchmark
```
milseman, line 1: If the benchmarks just added themselves to an array, would that simplify the registration process?

milseman (approved):
LGTM

rctcwyvrn:
@swift-ci test

PR #511, opened 2022-06-22
Use a bitset for ascii-only character classes

rctcwyvrn:
In the cases where a custom character set is made up of ascii characters, scalars, and ranges, use a 128 bit set to represent the character class instead of using the mess of closures we were using before

Branched from https://github.com/apple/swift-experimental-string-processing/pull/509

Results show big improvements on the custom character class microbenchmarks, and smaller improvements in the real world benchmarks where the closure overhead represented less of the runtime, notably the email regexes which contain character classes that have many non-range items

```
=== Improvements ====================================================
- basicCCC				11.2ms	33.5ms	-22.3ms	-67.0%
- basicRangeCCC				12ms	13.5ms	-1.58ms	-12.0%
- invertedCCC				30.6ms	48.9ms	-18.2ms	-37.0%
- caseInsensitiveCCC			13.1ms	124ms	-111ms	-89.0%

- GraphemeBreakNoCapFirst		85.5µs	122µs	-37µs	-30.0%
- emailLookaheadNoMatchesFirst		62.5ms	88.3ms	-25.8ms	-29.0%
- emailRFCAll				51.4ms	54.3ms	-2.89ms	-5.0%
- cssAll				5.03ms	5.33ms	-297µs	-6.0%
- emailLookaheadAll			86.1ms	103ms	-16.6ms	-16.0%
- GraphemeBreakNoCapAll			9.49ms	10.1ms	-621µs	-6.0%
- emailLookaheadNoMatchesAll		62.5ms	88.3ms	-25.7ms	-29.0%
- emailLookaheadFirst			19.2µs	22.7µs	-3.5µs	-15.0%
- emailRFCFirst				8.33µs	9.75µs	-1.42µs	-15.0%
```


Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -51,7 +51,43 @@ extension DSLTree.Node {
...
+      return c.asciiValue!
+    case let .scalar(s):
+      return UInt8(ascii: s)
+    case let .unconverted(atom):
+      return atom.toAsciiValue()
+    default:
+      fatalError("Should have been checked by isAscii first")
+    }
+  }
+
```
milseman, line 90: How about `var asciiValue: UInt8?` for both of these?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -51,7 +51,43 @@ extension DSLTree.Node {
...
+      return atom.isAscii()
+    default:
+      return false
+    }
+  }
+  
+  func toAsciiValue() -> UInt8 {
+    switch self {
+    case let .char(c):
+      return c.asciiValue!
```
milseman, line 81: We should exclude CR-LF here.

>   /// A character with the value "\r\n" (CR-LF) is normalized to "\n" (LF) and
>  /// has an `asciiValue` property equal to 10.

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -235,6 +293,42 @@ extension AST.Atom {
...
+    case let .atom(a):
+      return a.isAscii()
+    case let .range(low, high):
+      return low.isAscii() && high.isAscii()
+    // The remaining cases have nested character classes with possibly different
+    // inversion so leave them out of this optimization
+    default:
+      return false
+    }
+  }
```
milseman, line 307: similarly `var asciiBitset: ASCIIBitset?`

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -226,6 +226,20 @@ extension Processor {
...
     return true
   }
+  
+  // If we have a bitset we know that the CharacterClass only matches against
+  // ascii characters, so check if the current input element is ascii then
+  // check if it is set in the bitset
+  mutating func matchBitset(
+    _ bitset: DSLTree.CustomCharacterClass.AsciiBitset
+  ) -> Bool {
+    guard let cur = load(), bitset.matches(char: cur) else {
```
milseman, line 236: Future work: we can implement this on the UTF-8 view, but we'd have to handle grapheme breaking ourselves.

BTW, @natecook1000 what is the model for semantic mode processing around here?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -164,6 +164,96 @@ extension DSLTree {
       indirect case subtraction(CustomCharacterClass, CustomCharacterClass)
       indirect case symmetricDifference(CustomCharacterClass, CustomCharacterClass)
     }
+    
+    public struct AsciiBitset {
+      let isInverted: Bool
+      let isCaseInsensitive: Bool
+      var a: UInt64 = 0
+      var b: UInt64 = 0
```
milseman, line 172: This is ok, another option is something like `var rawValue: (UInt64, UInt64)` and add the OptionSet conformance.

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -164,6 +164,96 @@ extension DSLTree {
...
+        setBit(val)
+        if isCaseInsensitive {
+          let c = Character(Unicode.Scalar.init(val))
+          let otherCase: String
+          if c.isUppercase {
+            otherCase = c.lowercased()
+          } else {
+            otherCase = c.uppercased()
+          }
+          setBit(otherCase.first!.asciiValue!)
```
milseman, line 215: I suppose this works, but ideally we'd have fewer assumptions baked in. @Azoy would just asking for the scalar's mapping work too?

milseman: Alternatively, since it's pure ASCII, we could just compute and apply the mapping.

milseman (commented):
Looking good so far; some early feedback.

Sources/_StringProcessing/Regex/DSLTree.swift
Azoy: If we're only dealing with ASCII here, we should probably just compute the uppercase/lowercase since there's relatively few scalars in ASCII that actually have uppercase/lowercase (is it just the letters?)

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -164,6 +164,96 @@ extension DSLTree {
       indirect case subtraction(CustomCharacterClass, CustomCharacterClass)
       indirect case symmetricDifference(CustomCharacterClass, CustomCharacterClass)
     }
+    
+    public struct AsciiBitset {
+      let isInverted: Bool
+      let isCaseInsensitive: Bool
```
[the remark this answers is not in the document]
rctcwyvrn: What do you mean?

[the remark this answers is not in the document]
natecook1000: I don't think we can fold `isInverted` into the value, since that needs to match every non-ASCII character in addition to the ASCII characters that aren't included.

Sources/_StringProcessing/Regex/DSLTree.swift
rctcwyvrn: `(UInt64, UInt64)` doesn't have `FixedWidthInteger` conformance so I think we'd end up having to implement all of the set operations ourselves. I just left it as is

Sources/_StringProcessing/Regex/DSLTree.swift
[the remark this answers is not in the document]
rctcwyvrn: Oops, I was already folding in the value for isCaseInsensitive, I just forgot to remove the stored property. And yea as Nate said we can't fold isInverted into the value

Tests/RegexTests/MatchTests.swift
```diff
@@ -571,6 +571,12 @@ extension RegexTests {
     // Character class subtraction
     firstMatchTest("[a-d--a-c]", input: "123abcdxyz", match: "d")
 
+    // Inverted character class
+    matchTest(#"[^a]"#,
+              ("💿", true),
+              ("A", true),
+              ("a", false))
+
```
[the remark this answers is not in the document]
rctcwyvrn: I added some, does that cover the cases you were thinking of?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -167,63 +167,55 @@ extension DSLTree {
...
+      internal mutating func add(_ val: UInt8, _ isCaseInsensitive: Bool) {
         setBit(val)
         if isCaseInsensitive {
-          let c = Character(Unicode.Scalar.init(val))
-          let otherCase: String
-          if c.isUppercase {
-            otherCase = c.lowercased()
-          } else {
-            otherCase = c.uppercased()
+          if val >= 64 && val <= 90 {
```
milseman, line 202: Can also write it as a switch over `val` with ranges. Something like

```
switch val {
  case 64...90: setBit(val + 32)
  case 97...122: setBit(val - 32)
  default: break
}
```

Tests/RegexTests/MatchTests.swift
```diff
@@ -574,9 +574,27 @@ extension RegexTests {
...
 
+    matchTest("[a]",
+      ("a\u{301}", false))
+
+    // CR-LF special case: \r\n is a single character with ascii value equal
+    // to \n, so make sure the ascii bitset optimization handles this correctly
+    matchTest("[\r\n]",
+      ("\r\n", true),
+      ("\n", false),
+      ("\r", false))
```
milseman, line 589: @natecook1000 does this depend on matching semantics mode? That is, if we see `[\r\n]` is that either CR or LF, or is it exactly CR-LF, or does it depend on mode?

Tests/RegexTests/MatchTests.swift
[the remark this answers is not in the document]
milseman: Can you add a test against the input of `"a\u{301}"`, that is we need to guarantee the grapheme cluster boundary after the "a".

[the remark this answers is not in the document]
rctcwyvrn: Oh I did add one, the diff it's showing is outdated for some reason 

```
matchTest(#"[^a]"#,
              ("💿", true),
              ("a\u{301}", true),
              ("A", true),
              ("a", false))

matchTest("[a]",
      ("a\u{301}", false))
```

Sources/_StringProcessing/Engine/Processor.swift
milseman: So IIUC, this optimization _only_ applies to grapheme-semantic mode right now, which is an unfortunate limitation. Lily, can you make sure to write a test for this somehow? We may need to revise our compilation testing approach.

milseman (approved):
Overall I'm pro this change. I have some concerns and want to make sure we're not doing the wrong thing in scalar semantic mode.

Sources/_StringProcessing/Engine/Processor.swift
rctcwyvrn: I think it would make sense to do optimizations for scalar mode (new instructions, new processor functions) as a future PR, for now I just made it not generate the bitset when in scalar mode.

I also added some support to `Compiler` and `CompileTests` to check for the existence of certain opcodes under different semantic levels

rctcwyvrn:
@swift-ci test

milseman (approved):
Cleanup for future noted, otherwise LGTM

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -644,7 +650,8 @@ fileprivate extension Compiler.ByteCodeGen {
     _ ccc: DSLTree.CustomCharacterClass
   ) throws {
     if let asciiBitset = ccc.asAsciiBitset(options),
-        options.semanticLevel == .graphemeCluster {
+        options.semanticLevel == .graphemeCluster,
+        !compileOptions.contains(.unoptimized) {
```
milseman, line 654: Maybe add a 

```
fileprivate var optimizationsEnabled: Bool { !compileOptions.contains(.disableOptimizations) }
```

Sources/_StringProcessing/Regex/Core.swift
```diff
@@ -132,3 +135,10 @@ extension Regex {
     self.program = Program(tree: .init(node))
   }
 }
+
+@available(SwiftStdlib 5.7, *)
+extension Regex {
+  internal mutating func _setCompilerOptionsForTesting(_ opts: Compiler.CompileOptions) {
```
milseman, line 141: Hmm, my concern is that this might not interact well with how we cache compiled programs. How would a caller compile and run it both ways?

Tests/RegexTests/MatchTests.swift
```diff
@@ -14,22 +14,32 @@ import XCTest
...
   }
   let caps = result.output.slices(from: input)
+  
+  if validate {
+    var unoptRegex = try Regex(regexStr, syntax: syntax)
+    unoptRegex._setCompilerOptionsForTesting(.unoptimized)
+    guard let unoptResult = try unoptRegex.firstMatch(in: input) else {
+      throw MatchError("match not found for unoptimized \(regexStr) in \(input)")
+    }
+    XCTAssertEqual(String(input[result.range]), String(input[unoptResult.range]))
```
milseman, line 41: Can you add an optimization bug (locally) to make sure this gets triggered? (or vice versa)

Tests/RegexTests/MatchTests.swift
```diff
@@ -14,22 +14,32 @@ import XCTest
...
+  var message: String
+  init(_ message: String) {
+    self.message = message
+  }
 }
 
 func _firstMatch(
   _ regexStr: String,
   input: String,
+  validate: Bool,
```
milseman, line 26: Probably validateOptimizations or something.

milseman (commented):
Again, looking good. You can feel free to merge or keep working off the same PR

Tests/RegexTests/MatchTests.swift
rctcwyvrn: Yep and it was triggered as expected

rctcwyvrn:
@swift-ci test

PR #512, opened 2022-06-22
Make the output of benchmark compares nicer

rctcwyvrn:
- Space out the names properly instead of relying on tabs
- Add a decimal point to the percentage
- Filter out NS benchmarks from the comparison (since they should never improve/regress, but they still show up due to our machines being inconsistent)
- Sort comparisons by amount of improvement/regression (by s, not % beceause we have lots of variance + low runtime benchmarks)


rctcwyvrn:
@swift-ci test

PR #513, opened 2022-06-23
Make the variadic overloads AEIC

Azoy:
This marks all of the variadic builder overloads as `@_alwaysEmitIntoClient` so that these aren't permanently part of our ABI. However, in order to do so we _have_ to expose DSLTree and friends as a public interface. This removes all of the `@_spi(RegexBuilder)` attributes and makes everything public but underscored.


Azoy:
@swift-ci please test

rxwei (commented):
I'm pretty concerned about making the tree itself be ABI. Have you considered declaring only a set of `@usableFromInline` `Regex` factory functions instead?  It's particularly risky to expose `DSLTree.Node` as ABI because we might want future optimizations to the representation, such as changing it to a struct.

I was imagining something along the lines of:

```swift
enum _RegexFactoryABI {
    @usableFromInline
    struct _Node {
        let node: DSLTree.Node // So that we never expose the tree itself
    }

    @usableFromInline
    func concatenation<Output>(_: [_Node], output: Output.Type) -> Regex<Output>

    @usableFromInline
    func alternation<Output>(_: [_Node], output: Output.Type) -> Regex<Output>

    ...
}
```

milseman:
I support adding an underscored function (underscore can be on the namespace) to creating corresponding DSL tree nodes and removing the `public` and `spi` from anything we possibly can after that.

milseman:
(assuming these creation methods let us de-`public` DSLTree nodes)

PR #514, opened 2022-06-24
Make variadic builder overloads AEIC (take 2)

Azoy:
This marks all of the variadic builder overloads as @_alwaysEmitIntoClient so that these aren't permanently part of our ABI.

### DIFF
--- a/Sources/RegexBuilder/DSL.swift
+++ b/Sources/RegexBuilder/DSL.swift
@@ -140,6 +140,7 @@ public struct One<Output>: RegexComponent {
 public struct OneOrMore<Output>: _BuiltinRegexComponent {
   public var regex: Regex<Output>
 
+  @usableFromInline
   internal init(_ regex: Regex<Output>) {
     self.regex = regex
   }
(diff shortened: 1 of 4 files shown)


Azoy:
@swift-ci please test

Azoy:
@swift-ci please test

rxwei (approved):

milseman (approved):
This seems like a better route than first take.

If we instead had some of these in `_StringProcessing`, would we be able to outright remove `DSLTree` as SPI? That would be an even cleaner not-technically-stable ABI.

rxwei:
> If we instead had some of these in _StringProcessing, would we be able to outright remove DSLTree as SPI? That would be an even cleaner not-technically-stable ABI.

That would require the `_RegexFactory` to be public, whereas it can be internal in the RegexBuilder module today. I don't think the SPI is concerning since the two modules are always co-evolved and built together.

rxwei:
IMO it's better to have SPI than make things public. Today, I don't think _StringProcessing exports any `public` underscored symbols.

milseman:
I don't think we're talking about the same thing. I'm not saying anything should be non-SPI public that isn't API. Why is it better to make DSLTree SPI-public instead of making _RegexFactory SPI-public?

rxwei:
Oh sorry I misunderstood. My impression was one couldn't use SPIs in AEIC functions? I may be wrong.

Azoy:
Ah that's right, if _RegexFactory was public SPI in string processing, we couldn't use it in AEIC.

rxwei:
In that case, it is indeed a SPI/public tradeoff. IMO not exposing new public decls is better.

PR #517, opened 2022-06-28
Reject .NET subtraction and quote range operands

hamishknight:
Tighten up validation of character class range operands such that we reject quotes and custom character classes. This includes rejecting syntax that would be a subtraction in .NET. We throw a custom error that suggests using `--` instead.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/Parse.swift
+++ b/Sources/_RegexParser/Regex/Parse/Parse.swift
@@ -557,7 +557,7 @@ extension Parser {
   mutating func parsePotentialCCRange(
     into members: inout [CustomCC.Member]
   ) throws {
-    guard case .atom(let lhs)? = members.last else { return }
+    guard let lhs = members.last, lhs.isSemantic else { return }
 
     // Try and see if we can parse a character class range. Each time we parse
     // a component of the range, we append to `members` in case it ends up not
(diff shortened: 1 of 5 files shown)


Sources/_RegexParser/Regex/Parse/Parse.swift
```diff
@@ -550,48 +552,88 @@ extension Parser {
...
+        return a
+      case .custom:
+        // Not supported. While .NET allows `x-[...]` to spell subtraction, we
+        // require `x--[...]`. We also ban `[...]-x` for consistency.
+        if isLHS {
+          throw Source.LocatedError(
+            ParseError.invalidCharacterClassRangeOperand, m.location)
+        } else {
+          throw Source.LocatedError(
+            ParseError.unsupportedDotNetSubtraction, m.location)
```
hamishknight, line 595: @milseman We don't strictly have to ban `[[a]-b]` here (we could instead treat `-` as literal), but it seems reasonable to me at least to be consistent with the banning of `[a-[b]]`. Any thoughts?

Tests/RegexTests/ParseTests.swift
```diff
@@ -517,10 +517,36 @@ extension RegexTests {
...
+
+    parseTest("(?x)[a-c - - -]", concat(
+      changeMatchingOptions(matchingOptions(adding: .extended)),
+      charClass(range_m("a", "c"), range_m("-", "-"))
+    ))
 
     parseTest("[-a-]", charClass("-", "a", "-"))
     parseTest("[[a]-]", charClass(charClass("a"), "-"))
-    parseTest("[[a]-b]", charClass(charClass("a"), "-", "b"))
+    parseTest("[-[a]]", charClass("-", charClass("a")))
```
hamishknight, line 539: @milseman This could be an invalid .NET subtraction, but for consistency with PCRE and ICU, I thought we should probably allow it as literal. Any thoughts?

Sources/_RegexParser/Regex/Parse/Parse.swift
milseman: I'd prefer a ban right now than lots of magical reasoning (from the developer's perspective) around whether `-` is literal or not.

milseman (approved):
LGTM

PR #518, opened 2022-06-28
[5.7] Reject .NET subtraction and quote range operands

hamishknight:
*5.7 cherry-pick of https://github.com/apple/swift-experimental-string-processing/pull/517*

Tighten up validation of character class range operands such that we reject quotes and custom character classes. This includes rejecting syntax that would be a subtraction in .NET. We throw a custom error that suggests using `--` instead.


hamishknight:
@swift-ci please test

PR #520, opened 2022-06-28
Make variadic builder overloads AEIC (take 3)

Azoy:
This moves `_RegexFactory` into s-p as a public underscored facility and de-publicizes some of the types in DSLTree, most notably Node. There are a few types that are still public spi for regexbuilder to work however.


rxwei (commented):
This is exposing new public decls. I’m not sure why this is better than take 2.

Sources/_StringProcessing/Utility/RegexFactory.swift
```diff
@@ -0,0 +1,193 @@
...
+  ) -> Regex<Output> {
+    .init(node: .customCharacterClass(ccc))
+  }
+  
+  @available(SwiftStdlib 5.7, *)
+  public static func zeroOrOne<Output>(
+    _ component: some RegexComponent,
+    _ behavior: RegexRepetitionBehavior? = nil
+  ) -> Regex<Output> {
+    let kind: DSLTree.QuantificationKind = behavior.map { .explicit($0.dslTreeKind) } ?? .default
```
milseman, line 81: Minor code cleanup: what about an initializer that takes an optional behavior? Then this can just be `.init(behavior)` for these call sites.

milseman (approved):
This is looking much better. Let's continue to think if we can get a little more `@_spi` on or closer to `_RegexFactory`, but this is better than putting Node into our ABI.

Azoy:
@swift-ci please test

milseman (approved):
LGTM

PR #515, opened 2022-06-24
Unicode and DNA regexes + more benchmarker improvements

rctcwyvrn:
- Benchmarks for unicode regexes and matching around unicode data
- Fix the names 
- Add suite filtering by patterns instead of names
- Exclude `firstMatch` benchmarks by default because they were running too quickly and not providing information the `matches` benchmarks weren't
- Add flag for excluding the NSRegex comparison benchmarks

PR #525, opened 2022-06-30
Optimize matching to match on scalar values when possible

rctcwyvrn:
- Emit `matchScalar` instructions when emitting ascii quoted literals, ascii characters, and any scalar, as well as when in unicode scalars mode. `matchScalar` stores the scalar value inline instead of in a register like the character for `match`, making it faster and less susceptible to ARC
- Add `matchScalarBitset` to use the optimization from https://github.com/apple/swift-experimental-string-processing/pull/511 when in unicode scalars mode

Future work: We currently on match on ASCII because we may be given non normalized strings. Two options to deal with this are to A. normalize inputs B. compile two different programs, one optimized for if the input string is normalized and the other not making any assumptions

Large wins in the DNA benchmarks from https://github.com/apple/swift-experimental-string-processing/pull/515 since they have long quoted literal sequences in their patterns
```
- DnaEndsMatchAll                         200ms	288ms	-88.8ms		-30.8%
- DnaMatchAll                             112ms	158ms	-45.9ms		-29.0%
- DnaEndsMatchFirst                       38.2ms	55.3ms	-17.1ms		-30.9%
- LiteralSearchAll                        70.6ms	85.4ms	-14.8ms		-17.3%
- EmailRFCNoMatchesAll                    98.3ms	108ms	-9.45ms		-8.8%
- EmailRFCAll                             42.2ms	51.2ms	-8.98ms		-17.5%
- DnaMatchFirst                           19.5ms	27.9ms	-8.38ms		-30.0%
- EmailLookaheadAll                       74.9ms	82.5ms	-7.63ms		-9.2%
- EmailLookaheadNoMatchesAll              53.8ms	60.5ms	-6.74ms		-11.1%
- ReluctantQuantWithTerminalWhole         8.6ms	10.5ms	-1.87ms		-17.9%
- HtmlAll                                 11.5ms	13.4ms	-1.85ms		-13.8%
- HangulSyllableAll                       6.69ms	8.25ms	-1.56ms		-18.9%
- TaggedEmojisAll                         16.6ms	18.1ms	-1.5ms		-8.3%
- InvertedCCC                             27.8ms	29.1ms	-1.35ms		-4.6%
- CaseInsensitiveCCC                      11.1ms	12ms	-933µs		-7.8%
- HangulSyllableFirst                     3.1ms	3.99ms	-891µs		-22.3%
- CssAll                                  4.05ms	4.85ms	-794µs		-16.4%
- BasicRangeCCC                           10.4ms	11.2ms	-785µs		-7.0%
- BasicCCC                                9.68ms	10.4ms	-742µs		-7.1%
- NotFoundAll                             6.41ms	7.06ms	-654µs		-9.3%
```


Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -74,6 +74,77 @@ fileprivate extension Compiler.ByteCodeGen {
...
+          while let ch = iterator.next() {
+            guard currentIndex < bounds.upperBound,
+                  ch.lowercased() == input[currentIndex].lowercased()
+            else { return nil }
+            input.formIndex(after: &currentIndex)
+          }
+          return currentIndex
+        }
+      } else {
+        if optimizationsEnabled && s.allSatisfy({char in char.isASCII}) {
```
milseman, line 96: Could this be a guard-else instead, to simplify the control flow a bit?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -74,6 +74,77 @@ fileprivate extension Compiler.ByteCodeGen {
...
+              builder.buildMatchScalar(scalar, boundaryCheck: false)
+            }
+          }
+          let lastChar = s.last!
+          for scalar in lastChar.unicodeScalars {
+            // Only boundary check if we are the last scalar in the last character
+            // to make sure that there isn't a combining scalar after the quoted literal
+            let boundaryCheck = scalar == lastChar.unicodeScalars.last!
+            builder.buildMatchScalar(scalar, boundaryCheck: boundaryCheck)
+          }
```
milseman, line 109: I think it would be simpler to say something like

```
let lastIdx = s.unicodeScalars.indicies.last! // BTW, make sure we have a non-empty invariant
for idx in s.unicodeScalars.indices {
  ...
  ... if isGraphemeSemantic && idx == lastIdx {
    ...
  } else { 
    ...
  }
}
```

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -74,6 +74,77 @@ fileprivate extension Compiler.ByteCodeGen {
     }
   }
 
+  mutating func emitQuotedLiteral(_ s: String) {
+    if options.semanticLevel == .graphemeCluster {
+      if options.isCaseInsensitive {
+        // future work: if all ascii, emit matchBitset instructions with
+        // case insensitive bitsets
```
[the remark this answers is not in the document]
milseman: I.e. we'll want to get all of this off of using consumer interfaces at some point, but doesn't have to be this PR.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -216,6 +287,11 @@ fileprivate extension Compiler.ByteCodeGen {
       builder.buildConsume(by: consumeScalar {
         $0.properties.lowercaseMapping == s.properties.lowercaseMapping
       })
+      return
+    }
+    
+    if optimizationsEnabled { // should we just do this unconditionally?
```
milseman, line 293: Let's do this unconditionally, and consider getting case insensitive into the instruction (whether separate opcode or bit in opcode, same thing effectively either way). Can you double check that we have good testing here that the grapheme boundary isn't checked? That is, knowing and testing whatever code paths get to `emitScalar`.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -241,9 +317,17 @@ fileprivate extension Compiler.ByteCodeGen {
           ? input.index(after: bounds.lowerBound)
           : nil
       }
-    } else {
-      builder.buildMatch(c)
     }
+    
+    if optimizationsEnabled && c.isASCII {
+      for scalar in c.unicodeScalars {
+        let boundaryCheck = scalar == c.unicodeScalars.last!
```
milseman, line 324: I think you need to check indices. You could have the same combining acute accent appear multiple times. 

Actually, could you write a test of exactly that while we have the bug here?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -241,9 +317,17 @@ fileprivate extension Compiler.ByteCodeGen {
...
+        builder.buildMatchScalar(scalar, boundaryCheck: boundaryCheck)
```
milseman, line 325: Another idea for the builder is to have a convenience function that adds a grapheme boundary check at the current position. It could either update the most recent instruction if that instruction is capable of supporting it or insert the anchor otherwise.

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -11,6 +11,13 @@
 
 @_implementationOnly import _RegexParser
 
+extension Character {
+  var singleScalarAsciiValue: UInt8? {
```
milseman, line 15: Weakly prefer underscoring this, since it could otherwise be confusable with API. I realize we've not been super consistent on this.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -227,6 +227,30 @@ extension Processor {
...
+       (!boundaryCheck || input.isOnGraphemeClusterBoundary(idx)) {
+      return idx
+    }
+    return nil
+  }
+  
+  mutating func matchScalar(_ s: Unicode.Scalar, boundaryCheck: Bool) -> Bool {
+    guard let curScalar = loadScalar(),
+          curScalar == s,
+          let idx = nextScalarIndex(offsetBy: 1, boundaryCheck: boundaryCheck)
```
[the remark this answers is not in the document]
milseman: Or you could consider adding that now if you like, something like the below and we can micro-optimize it later:

```
extension String {
  func _consumeScalar(in bounds: Range<String.Index>) -> (Unicode.Scalar, Index)? {
    assert(indices.contains(bounds))
    ... 
  }
}
```

Sources/_StringProcessing/ByteCodeGen.swift
rctcwyvrn: If we have a combining accent then c would not be ascii so I don't think the old code has a bug, the only multi scalar ascii character is CR-LF. In any case index checking is cleaner and doesn't have to worry about edge cases like that so I'll rewrite it to do that

rctcwyvrn:
While implementing case insensitive matching instructions I opted for adding new instructions for each new case, however after doing that I benchmarked it and noticed pretty significant regressions 

```
Comparing against benchmark result file before_case_insens_instrs.json
- HangulSyllableFirst                     3.22ms	3.02ms	195µs		6.4%
- CssAll                                  4.34ms	4.05ms	289µs		7.1%
- HangulSyllableAll                       7.09ms	6.69ms	394µs		5.9%
- EagarQuantWithTerminalWhole             7ms	6.5ms	505µs		7.8%
- NotFoundAll                             6.95ms	6.42ms	535µs		8.3%
- GraphemeBreakNoCapAll                   9.8ms	9.22ms	578µs		6.3%
- BasicRangeCCC                           11.2ms	10.4ms	830µs		8.0%
- TaggedEmojisAll                         17.3ms	16.5ms	833µs		5.0%
- BasicCCC                                10.5ms	9.69ms	852µs		8.8%
- ReluctantQuantWithTerminalWhole         9.47ms	8.6ms	869µs		10.1%
- HtmlAll                                 12.4ms	11.5ms	873µs		7.6%
- IntersectionCCC                         16.2ms	15.3ms	901µs		5.9%
- SubtractionCCC                          15.7ms	14.8ms	936µs		6.3%
- CaseInsensitiveCCC                      12.1ms	11.1ms	1.03ms		9.3%
- symDiffCCC                              39.5ms	38.4ms	1.04ms		2.7%
- AnchoredNotFoundAll                     16ms	14.9ms	1.1ms		7.4%
- ReluctantQuantWhole                     15.2ms	13.9ms	1.27ms		9.1%
- InvertedCCC                             28.7ms	27.3ms	1.36ms		5.0%
- DnaMatchFirst                           21.4ms	19.6ms	1.79ms		9.2%
- EmailRFCAll                             45.5ms	42.2ms	3.28ms		7.8%
- DnaEndsMatchFirst                       41.9ms	38.1ms	3.8ms		10.0%
- EmailRFCNoMatchesAll                    102ms	98.3ms	3.99ms		4.1%
- LiteralSearchNotFoundAll                67.7ms	63.1ms	4.57ms		7.2%
- EmailLookaheadNoMatchesAll              58.8ms	53.8ms	4.94ms		9.2%
- LiteralSearchAll                        75.7ms	70.7ms	5.04ms		7.1%
- EmojiRegexAll                           68.1ms	61.7ms	6.38ms		10.3%
- EmailLookaheadAll                       81.3ms	74.8ms	6.47ms		8.7%
- DnaMatchAll                             123ms	112ms	10.6ms		9.4%
- DnaEndsMatchAll                         222ms	200ms	21.9ms		11.0%
```

After adjusting the scalar matching instructions to use one instruction with a more complex payload instead of 3 instructions the regression was much less significant

```
Comparing against benchmark result file before_case_insens_instrs.json
- CssAll                                  4.08ms	4.05ms	29.7µs		0.7%
- EagarQuantWithTerminalWhole             6.56ms	6.5ms	61.7µs		0.9%
- NotFoundAll                             6.5ms	6.42ms	77.9µs		1.2%
- GraphemeBreakNoCapAll                   9.31ms	9.22ms	94.8µs		1.0%
- HtmlAll                                 11.6ms	11.5ms	101µs		0.9%
- ReluctantQuantWithTerminalWhole         8.75ms	8.6ms	150µs		1.7%
- InvertedCCC                             27.5ms	27.3ms	153µs		0.6%
- ReluctantQuantWhole                     14.1ms	13.9ms	164µs		1.2%
- DnaMatchFirst                           19.7ms	19.6ms	168µs		0.9%
- TaggedEmojisAll                         16.7ms	16.5ms	172µs		1.0%
- IntersectionCCC                         15.5ms	15.3ms	181µs		1.2%
- SubtractionCCC                          15ms	14.8ms	188µs		1.3%
- DnaEndsMatchFirst                       38.4ms	38.1ms	217µs		0.6%
- LiteralSearchAll                        71.4ms	70.7ms	729µs		1.0%
- EmailRFCAll                             43ms	42.2ms	787µs		1.9%
- DnaMatchAll                             114ms	112ms	1.16ms		1.0%
- symDiffCCC                              39.9ms	38.4ms	1.49ms		3.9%
- DnaEndsMatchAll                         202ms	200ms	2.46ms		1.2%
- EmailRFCNoMatchesAll                    102ms	98.3ms	3.96ms		4.0%
- AnchoredNotFoundAll                     14.7ms	14.9ms	-143µs		-1.0%
- BasicRangeCCC                           10.3ms	10.4ms	-121µs		-1.2%
- BasicCCC                                9.58ms	9.69ms	-116µs		-1.2%
- CaseInsensitiveCCC                      11ms	11.1ms	-99.6µs		-0.9%
```

In the future we could probably continue to squeeze performance out by creating specialized instructions only in the hottest paths but for now I'll implement everything by adding bits into the payload

Got pretty large improvements from removing 6 instructions (matchScalar unchecked, case insensitive, case insensitive unchecked, match case insensitive, match seq, match scalar bitset). I'll have to rework how `CompileTests` work though, since it'll have to look in the payloads to differentiate the instructions now
```
Comparing against benchmark result file matchSeq_and_many_case_insens_instrs.json
- DnaEndsMatchAll                         205ms	220ms	-15.4ms		-7.0%
- DnaMatchAll                             116ms	123ms	-7.73ms		-6.3%
- EmailLookaheadAll                       75.1ms	81.4ms	-6.26ms		-7.7%
- EmailRFCAll                             42.4ms	48.3ms	-5.91ms		-12.2%
- EmojiRegexAll                           62.4ms	68ms	-5.58ms		-8.2%
- EmailRFCNoMatchesAll                    98.8ms	104ms	-5.49ms		-5.3%
```

rctcwyvrn:
Currently `CompileTests` are broken, I'll fix them tomorrow

rctcwyvrn:
@swift-ci test

milseman:
@swift-ci please test

milseman (approved):

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -62,7 +62,11 @@ fileprivate extension Compiler.ByteCodeGen {
       emitCharacter(c)
 
     case let .scalar(s):
-      emitScalar(s)
+      if options.semanticLevel == .graphemeCluster {
+        emitCharacter(Character(s))
+      } else {
+        emitMatchScalar(s)
+      }
```
milseman, line 69: This should fail tests, and if it's not we may need to write the tests and XFAIL them. E.g. `/a\u{301}/`, IIUC, as written would try to make a proper Character from the combining scalar.

rctcwyvrn:
@swift-ci test

PR #541, opened 2022-07-05
Throw error when Output is wrong type

Azoy:
When casting a string representation to a regex, instead of throwing an actual error if the parsed output type does not match the input output type, we let the cast succeed and crash at runtime. Let's throw an error instead!


hamishknight (approved):
Thanks!

PR #545, opened 2022-07-05
Flatten optional nesting for regex literal captures

### LINKED ISSUE #399: Flatten optionality of regex captures
We currently nest optionals for regex literal captures, we should flatten them out to match the regex literal proposal.

### LINKED ISSUE #543: Failed captures produced `.some(nil)`
For e.g:

```
      let r = Regex {
        Optionally {
          Optionally {
            Capture {
              "a"
            }
          }
        }
      }
```

matching against the empty input yields a capture of `.some(nil)`. We ought to produce `nil`.

hamishknight:
When computing the `CaptureList` for AST nodes, including converted AST -> DSL nodes, only permit at most one level of optionality. This means that regex literal captures are now either `Substring` or `Substring?`.

Optional nesting is however still performed in the DSL (due to result builder limitations). If a regex literal is nested in the DSL, it may only add at most one extra level of optionality to the current nesting level.

Additionally, change the value of a failed nested optional capture from `.some(nil)` to `nil`.

Resolves #399
Resolves #543


hamishknight:
@swift-ci please test

rxwei (approved):
I haven't looked at the implementation very closely but the new behavior makes sense to me.

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

PR #546, opened 2022-07-05
Fix crash when emitting an atom of any

Azoy:
When emitting an atom of any, we were unconditionally failing and crashing the process. I think this is an oversight because this case can occur pretty often with `CharacterClass.any` who creates one of these.

Resolves: rdar://96280236


Azoy:
@swift-ci please test

hamishknight (commented):
Thanks! Looks reasonable to me

PR #547, opened 2022-07-05
Implement instructions for matching builtin character classes and assertions

rctcwyvrn:
- Adds `matchBuiltin` to match builtins instead of a consumeFn calling `_CharacterClassModel.matches` in most cases
- Removes `AssertionFunction` and changes `assertBy` to match in `Processor` instead

There are some gains (10% in microbenchmarks on builtin character classes) but the main purpose is to set up for future work

Future work:
- The layout of the payloads for `matchBuiltin` and `assertBy` will be optimized (hopefully soon) along with the rest of the instruction layout. Changes made in this branch and the scalar optimization branch showed that this is a strong next step
- I wasn't able to fully nuke `_CharacterClassModel.matches` but I think we're getting very close to being able to rip out most of it and the consumer interface for `CustomCharacterClass`


rctcwyvrn:
WIP because I need to figure out these weird performance behaviours

Weird thing number 1: The first commit in this branch (copying over the AsciiBitset changes from the matchScalar PR) is resulting in significant speedups

```
- emailLookaheadNoMatchesAll              58ms	63ms	-5.05ms		-8.0%
- emailLookaheadNoMatchesFirst            58ms	63ms	-4.95ms		-7.9%
- emailLookaheadAll                       81.9ms	86.8ms	-4.95ms		-5.7%
- basicCCC                                9.63ms	10.7ms	-1.08ms		-10.1%
- caseInsensitiveCCC                      11.3ms	12.3ms	-1.05ms		-8.6%
- invertedCCC                             28.5ms	29.4ms	-946µs		-3.2%
- emailRFCAll                             50.2ms	51.1ms	-942µs		-1.8%
- basicRangeCCC                           10.6ms	11.4ms	-782µs		-6.9%
```

I sanity checked this by trying to minimize the change (not copying it over to a new file, not adding the new scalar api) and got the same result. 

```
diff --git a/Sources/_StringProcessing/ConsumerInterface.swift b/Sources/_StringProcessing/ConsumerInterface.swift
index af46b53..9c41002 100644
--- a/Sources/_StringProcessing/ConsumerInterface.swift
+++ b/Sources/_StringProcessing/ConsumerInterface.swift
@@ -11,6 +11,13 @@
 
+extension Character {
+   var _singleScalarAsciiValue: UInt8? {
+     guard self != "\r\n" else { return nil }
+     return asciiValue
+   }
+ }
+
```
```
before
      internal func matches(char: Character) -> Bool {
        let ret: Bool
        if let val = char.asciiValue {
          if val < 64 {
            ret = (a >> val) & 1 == 1
          } else {
            ret =  (b >> (val - 64)) & 1 == 1
          }
        } else {
          ret = false
        }

        if isInverted {
          return !ret
        }

        return ret
      }
after
      private func matches(_ val: UInt8) -> Bool {
        if val < 64 {
          return (a >> val) & 1 == 1
        } else {
          return (b >> (val - 64)) & 1 == 1
        }
      }

       internal func matches(char: Character) -> Bool {
         let matched: Bool
         if let val = char._singleScalarAsciiValue {
           matched = matches(val)
         } else {
           matched = false
         }

         if isInverted {
           return !matched
         }
         return matched
       }
```

Some quick profiling seems to indicate that the performance increase came the call to `_singleScalarAsciiValue` being faster than the call to `asciiValue`, even though `_singleScalarAsciiValue` also uses `asciiValue`. My best guess is that it has to do with `Character.asciiValue.getter` being inlined before the change but not after, though I honestly don't know why that would make a difference

rctcwyvrn:
Weird performance thing number 2: The optimization is causing regressions in regexes that do not produce the new instruction and very little improvement in the ones that do

Comparing with main with the bitset changes, we get major regressions in all of the email regexes and smaller ones in almost every other result
```
Comparing against benchmark result file sanity_check_after_bitset.json
- htmlFirst                               26.9µs	26.7µs	209ns		0.8%
- emailLookaheadFirst                     18.5µs	18.2µs	250ns		1.4%
- emailRFCFirst                           8.38µs	7.92µs	458ns		5.8%
- cssFirst                                93µs	91.9µs	1.17µs		1.3%
- HangulSyllableFirst                     4.06ms	4.01ms	47.8µs		1.2%
- cssAll                                  4.87ms	4.82ms	47.9µs		1.0%
- caseInsensitiveCCC                      11.5ms	11.4ms	107µs		0.9%
- ReluctantQuantWithTerminalWhole         10.6ms	10.4ms	113µs		1.1%
- basicRangeCCC                           10.7ms	10.5ms	126µs		1.2%
- HangulSyllableAll                       8.41ms	8.24ms	167µs		2.0%
- htmlAll                                 13.5ms	13.3ms	205µs		1.5%
- subtractionCCC                          15.3ms	15.1ms	278µs		1.8%
- intersectionCCC                         15.8ms	15.5ms	278µs		1.8%
- notFoundAll                             15.1ms	14.8ms	285µs		1.9%
- notFoundFirst                           15.1ms	14.8ms	298µs		2.0%
- symDiffCCC                              39.2ms	38.3ms	885µs		2.3%
- invertedCCC                             29.5ms	28.6ms	937µs		3.3%
- emailLookaheadNoMatchesFirst            59.5ms	58.3ms	1.12ms		1.9%
- emailLookaheadNoMatchesAll              59.3ms	58.2ms	1.13ms		1.9%
- emailRFCAll                             51.5ms	50ms	1.53ms		3.1%
- emailLookaheadAll                       83.5ms	81.8ms	1.72ms		2.1%
- emailRFCNoMatchesAll                    110ms	107ms	3.03ms		2.8%
- emailRFCNoMatchesFirst                  110ms	107ms	3.06ms		2.8%
- GraphemeBreakNoCapAll                   9.2ms	9.45ms	-257µs		-2.7%
- EagarQuantWithTerminalWhole             6.67ms	6.76ms	-83.9µs		-1.2%
- GraphemeBreakNoCapFirst                 79.2µs	82.8µs	-3.54µs		-4.3%
```

- Checking the compilation does confirm that neither email regex is generating the new instruction
- The current benchmark suite doesn't have many benchmarks that mostly use builtin character classes, most mostly use custom classes and only use the builtins for whitespace. Need to add new microbenchmarks and real world benchmarks that use builtins.

rctcwyvrn:
Removing the remnants of the bitset fast path optimization I was testing seems to have reverted most of the regressions

```
Comparing against benchmark result file sanity_check_after_bitset.json
- emailRFCFirst                           8.17µs	7.92µs	249ns		3.1%
- invertedCCC                             28.8ms	28.6ms	167µs		0.6%
- emailLookaheadAll                       82.3ms	81.8ms	446µs		0.5%
- emailRFCAll                             50.5ms	50ms	528µs		1.1%
- GraphemeBreakNoCapAll                   9.02ms	9.45ms	-430µs		-4.6%
- caseInsensitiveCCC                      11ms	11.4ms	-375µs		-3.3%
- basicCCC                                9.54ms	9.9ms	-359µs		-3.6%
- ReluctantQuantWhole                     14ms	14.2ms	-238µs		-1.7%
- EagarQuantWithTerminalWhole             6.54ms	6.76ms	-216µs		-3.2%
- ReluctantQuantWithTerminalWhole         10.3ms	10.4ms	-184µs		-1.8%
- htmlAll                                 13.1ms	13.3ms	-137µs		-1.0%
- intersectionCCC                         15.4ms	15.5ms	-121µs		-0.8%
- basicRangeCCC                           10.5ms	10.5ms	-77.5µs		-0.7%
- cssAll                                  4.74ms	4.82ms	-74µs		-1.5%
- GraphemeBreakNoCapFirst                 78.5µs	82.8µs	-4.25µs		-5.1%
- cssFirst                                90.8µs	91.9µs	-1.13µs		-1.2%
- htmlFirst                               26.4µs	26.7µs	-291ns		-1.1%
```

But why? Why did an extra computed property on `_CharacterClassModel` and an extra computation when emitting a `matchBuiltin` affect all these other benchmarks???? Why are the email regexes still regressing?? (albeit only a little bit now) 

Need to make 100% sure that compile time is not affecting the benchmark times

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -243,6 +249,197 @@ extension Processor {
...
+    case .digit:
+      matched = c.properties.numericType != nil && (c.isASCII || !isStrictAscii)
+    case .hexDigit:
+      matched = Character(c).isHexDigit && (c.isASCII || !isStrictAscii)
+    case .horizontalWhitespace:
+      matched = c.isHorizontalWhitespace && (c.isASCII || !isStrictAscii)
+    case .verticalWhitespace:
+      matched = c.isNewline && (c.isASCII || !isStrictAscii)
+    case .newlineSequence:
+      matched = c.isNewline && (c.isASCII || !isStrictAscii)
```
rctcwyvrn, line 322: This was carried over from the old code, but why does this match a full `\r\n` sequence when in unicode scalars mode? Should we change that?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -243,6 +249,197 @@ extension Processor {
...
+          case .graphemeCluster:
+            return input[currentPosition].isNewline
+          case .unicodeScalar:
+            return input.unicodeScalars[currentPosition].isNewline
+          }
+        } else {
+          return currentPosition == subjectBounds.upperBound
+        }
+  
+      case .wordBoundary:
```
rctcwyvrn, line 425: There seem to be no test cases that use the `usesSimpleUnicodeBoundaries` option, should we be testing that?

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/InstPayload.swift
```diff
@@ -227,11 +226,64 @@ extension Instruction.Payload {
...
+    case .firstMatchingPositionInSubject: kind = 3
+    case .notTextSegment: kind = 4
+    case .notWordBoundary: kind = 5
+    case .resetStartOfMatch: kind = 6
+    case .startOfLine: kind = 7
+    case .startOfSubject: kind = 8
+    case .textSegment: kind = 9
+    case .wordBoundary: kind = 10
+    }
+    self.init(rawValue: kind + optionsBits)
```
milseman, line 259: I think it would clarify things to encapsulate this logic into a struct. That is the struct would store the `UInt64` and its methods would get/set the bits and kinds. Then the instruction's payload is just the struct's raw value (which we assert doesn't have the opcode bits set at all)

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -339,6 +340,107 @@ extension Processor {
...
+    _ semanticLevel: MatchingOptions.SemanticLevel
+  ) throws -> Bool {
+    // Future work: Optimize layout and dispatch
+    
+    // FIXME: Depends on API model we have... We may want to
+    // think through some of these with API interactions in mind
+    //
+    // This might break how we use `bounds` for both slicing
+    // and things like `firstIndex`, that is `firstIndex` may
+    // need to supply both a slice bounds and a per-search bounds.
```
milseman, line 353: I think this stuff got fixed when @natecook1000 added both subject bounds and search bounds.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -339,6 +340,107 @@ extension Processor {
...
+    switch kind {
+    case .startOfSubject: return currentPosition == subjectBounds.lowerBound
+
+    case .endOfSubjectBeforeNewline:
+      if currentPosition == subjectBounds.upperBound { return true }
+      switch semanticLevel {
+      case .graphemeCluster:
+        return input.index(after: currentPosition) == subjectBounds.upperBound
+         && input[currentPosition].isNewline
```
milseman, line 362: CC @Azoy , these are more examples of a consume operation that we want for more efficient access to String. Something like `func nom() -> (current: Character, next: Index)`

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -339,6 +340,107 @@ extension Processor {
...
+      case .unicodeScalar:
+        return input.unicodeScalars.index(after: currentPosition) == subjectBounds.upperBound
+         && input.unicodeScalars[currentPosition].isNewline
+      }
+
+    case .endOfSubject: return currentPosition == subjectBounds.upperBound
+
+    case .resetStartOfMatch:
+      // FIXME: Figure out how to communicate this out
+      throw Unsupported(#"\K (reset/keep assertion)"#)
```
milseman, line 372: Can you make sure this error is detected during compilation, not just evaluation?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -339,6 +340,107 @@ extension Processor {
...
+
+    case .firstMatchingPositionInSubject:
+      // TODO: We can probably build a nice model with API here
+
+      // FIXME: This needs to be based on `searchBounds`,
+      // not the `subjectBounds` given as an argument here
+      // (Note: the above fixme was in reference to the old assert function API.
+      //   Now that we're in processor, we have access to searchBounds)
+      return false
```
milseman, line 381: Huh? What should we do here?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -339,6 +340,107 @@ extension Processor {
...
+
+    case .textSegment: return input.isOnGraphemeClusterBoundary(currentPosition)
+
+    case .notTextSegment: return !input.isOnGraphemeClusterBoundary(currentPosition)
+
+    case .startOfLine:
+      // FIXME: Anchor.startOfLine must always use this first branch
+      // The behavior of `^` should depend on `anchorsMatchNewlines`, but
+      // the DSL-based `.startOfLine` anchor should always match the start
+      // of a line. Right now we don't distinguish between those anchors.
```
milseman, line 391: I think this should ideally be resolved at compilation time, since options are compilation options and not run-time options

Sources/_StringProcessing/Engine/Processor.swift
milseman: We definitely should be.

@Azoy?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -508,15 +610,18 @@ extension Processor {
...
-          &wordIndexMaxIndex,
-          input,
-          currentPosition,
-          subjectBounds
+        guard try regexAssert(
+          by: kind,
+          anchorsMatchNewlines,
+          usesSimpleUnicodeBoundaries,
+          usesASCIIWord,
+          semanticLevel
```
milseman, line 629: When we have the struct that just wraps the `UInt64` and has API for these queries, there's no need to pack/unpack it when evaluating. We can just branch directly on the bits.

Sources/_StringProcessing/Unicode/WordBreaking.swift
```diff
@@ -12,6 +12,39 @@
...
+    }
+    let priorIdx = input.index(before: currentPosition)
+    if currentPosition == subjectBounds.upperBound {
+      return matchesWord(at: priorIdx)
+    }
+    
+    let prior = matchesWord(at: priorIdx)
+    let current = matchesWord(at: currentPosition)
+    return prior != current
+  }
```
milseman, line 45: @Azoy can you review this code?

Sources/_StringProcessing/Engine/Processor.swift
rctcwyvrn: I think what the fixme is saying is that the DSLTree doesn't have a way to represent them differently right now, so we have no way of differentiating between them

Sources/_StringProcessing/Unicode/WordBreaking.swift
rctcwyvrn: For reference this was mostly copied from `_CharacterClassModel.isBoundary`

rctcwyvrn:
I may have gotten a little carried away while updating this branch to match main (internalization of `_CharacterClassModel` and changes to assertions) and I ended up implementing a new DSLTree node for character classes

It seems to all be working but there's a slight hiccup in inverting character classes that I'll work out on Monday

Sources/_StringProcessing/Engine/Processor.swift
rctcwyvrn: Fixed in https://github.com/apple/swift-experimental-string-processing/pull/580

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/Processor.swift
hamishknight: It's what is specified in [the Unicode proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0363-unicode-for-string-processing.md):

> Additionally, `\R` and `CharacterClass.newline` provide a way to include the CR+LF pair, even when matching with Unicode scalar semantics.

I don't know the exact rationale behind it, but I would assume it's done in an effort to be compatible with other regex engines that match scalars by default but with `\R` still matching `\r\n`.

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/Processor.swift
Azoy: We have a couple: `testSimpleWordBoundaries` and https://github.com/apple/swift-experimental-string-processing/blob/96fb215facf425226ca0645d5cab6e84a3befece/Tests/RegexBuilderTests/RegexDSLTests.swift#L540 but the latter was xfail, I have to fix the test because the test is wrong.

Sources/_StringProcessing/Unicode/WordBreaking.swift
Azoy: Looks good to me!

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+  ) -> Bool {
+    guard let c = load() else {
+      signalFailure()
+      return false
+    }
+
+    var matched: Bool
+    var next = input.index(after: currentPosition)
+    switch cc {
+    case .any, .anyGrapheme: matched = true
```
milseman, line 17: Can you test this against `"\n"`? I don't recall if the `any` here is `dot` or really any character.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+    case .anyScalar:
+      matched = true
+      next = input.unicodeScalars.index(after: currentPosition)
+    case .digit:
+      matched = c.isNumber && (c.isASCII || !isStrictAscii)
+    case .horizontalWhitespace:
+      matched = c.unicodeScalars.first?.isHorizontalWhitespace == true
```
milseman, line 24: Could you write this as an underscored property on `Character`, so that it follows the character properties model (even if it's not public)?

Sources/_StringProcessing/Engine/MEBuiltins.swift
milseman: Are these affected by strict ASCII? If so, can we hoist that check? If not, can we make a local

```
var asciiCheck: Bool { c.isASCII || !isStrictAscii }
```

to clean up the switch cases?

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+    _ isInverted: Bool,
+    _ isStrictAscii: Bool
+  ) -> Bool {
+    guard let c = load() else {
+      signalFailure()
+      return false
+    }
+
+    var matched: Bool
+    var next = input.index(after: currentPosition)
```
milseman, line 15: It seems like we can reduce the complexity a little bit by handling `.anyScalar` (and possibly others) early, and that would have fewer live variable across the switch.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+      signalFailure()
+      return false
+    }
+  }
+  
+  func isAtStartOfLine(_ payload: AssertionPayload) -> Bool {
+    if currentPosition == subjectBounds.lowerBound { return true }
+    switch payload.semanticLevel {
+    case .graphemeCluster:
+      return input[input.index(before: currentPosition)].isNewline
```
milseman, line 96: CC @Azoy do we have a test case for when this starts with a non-grapheme-cluster aligned lower bound?

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+    case .unicodeScalar:
+      return input.unicodeScalars[input.unicodeScalars.index(before: currentPosition)].isNewline
+    }
+  }
+  
+  func isAtEndOfLine(_ payload: AssertionPayload) -> Bool {
+    if currentPosition == subjectBounds.upperBound { return true }
```
milseman, line 103: ... similarly non-aligned upper bounds.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+    // Future work: Optimize layout and dispatch
+    switch payload.kind {
+    case .startOfSubject: return currentPosition == subjectBounds.lowerBound
+
+    case .endOfSubjectBeforeNewline:
+      if currentPosition == subjectBounds.upperBound { return true }
+      switch payload.semanticLevel {
+      case .graphemeCluster:
+        return input.index(after: currentPosition) == subjectBounds.upperBound
+         && input[currentPosition].isNewline
```
milseman, line 122: Similarly, alignment?

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+      case .unicodeScalar:
+        return input.unicodeScalars.index(after: currentPosition) == subjectBounds.upperBound
+         && input.unicodeScalars[currentPosition].isNewline
+      }
+
+    case .endOfSubject: return currentPosition == subjectBounds.upperBound
+
+    case .resetStartOfMatch:
+      fatalError("Unreachable, we should have thrown an error during compilation")
```
milseman, line 131: Future: we'll want to version support so that we can reason about back deployment

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+       _ anchorsMatchNewlines: Bool,
+       _ usesSimpleUnicodeBoundaries: Bool,
+       _ usesASCIIWord: Bool,
+       _ semanticLevel: MatchingOptions.SemanticLevel
+  ) {
+    // 4 bits of options
+    let anchorBit: UInt64 = anchorsMatchNewlines ? (1 << 55) : 0
+    let boundaryBit: UInt64 = usesSimpleUnicodeBoundaries ? (1 << 54) : 0
+    let strictBit: UInt64 = usesASCIIWord ? (1 << 53) : 0
+    let semanticLevelBit: UInt64 = semanticLevel == .unicodeScalar ? (1 << 52) : 0
```
milseman, line 199: These bits should be static computed vars.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
...
+    let optionsBits: UInt64 = anchorBit + boundaryBit + strictBit + semanticLevelBit
+
+    // 4 bits for the assertion kind
+    // Future work: Optimize this layout
+    let kind = assertion.rawValue
+    self.init(rawValue: kind + optionsBits)
```
milseman, line 205: Can we `OR` them together (and assert the `AND` is 0)?

milseman (commented):
In progress review, but plenty to communicate. Looks great! Lots of little tweaks and clarifications but this is awesome stuff!

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -162,6 +162,8 @@ extension DSLTree.Atom {
     case .assertion:
       // TODO: We could handle, should this be total?
       return nil
+    case .characterClass(let cc):
+      return cc.generateConsumer(opts)
```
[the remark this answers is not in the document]
rctcwyvrn: The main reason why all these consumer functions + `_CharacterClassModel.matches` still exist is because we have to emit any non-ascii CustomCharacterClass as one big consumer so it needs to know how to create a consumer for each possible member type (which includes Atom). 

Do you think we could emit it as an ordered choice over its members instead? I think the main thing to add would be instructions for handling the custom character class set operations, and if we had that then we could rip out all of the non character property related stuff in ConsumerInterface

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -182,6 +184,16 @@ extension DSLTree.Atom {
   }
 }
 
+extension DSLTree.Atom.CharacterClass {
+  func generateConsumer(_ opts: MatchingOptions) -> MEProgram.ConsumeFunction {
+    return { input, bounds in
+      // FIXME: should we worry about out of bounds?
```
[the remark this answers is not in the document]
rctcwyvrn: I had copied that fixme from the original code, looking more carefully we do have a guard in `Processor` so we're good. I'll remove the comment 

```
case .consumeBy:
      let reg = payload.consumer
      guard currentPosition < searchBounds.upperBound,
            let nextIndex = registers[reg](
              input, currentPosition..<searchBounds.upperBound)
      else {
        signalFailure()
        return
      }
```

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -1,13 +1,221 @@
-
+@_implementationOnly import _RegexParser // For AssertionKind
 
 extension Processor {
+  mutating func matchBuiltin(
+    _ cc: _CharacterClassModel.Representation,
+    _ isInverted: Bool,
+    _ isStrictAscii: Bool
+  ) -> Bool {
+    guard let c = load() else {
```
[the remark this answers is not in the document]
rctcwyvrn: We call the right matchBuiltin function based on the `isScalar` property in the payload 

```
case .matchBuiltin:
      let payload = payload.characterClassPayload
      if payload.isScalar {
        if matchBuiltinScalar(payload.cc, payload.isInverted, payload.isStrict) {
          controller.step()
        }
      } else {
        if matchBuiltin(payload.cc, payload.isInverted, payload.isStrict) {
          controller.step()
        }
      }
```

Sources/_StringProcessing/Engine/MEBuiltins.swift
Azoy: I'm unsure if we do or not. How would we go about adding a test for this?

rctcwyvrn: On a related note: what should we do about the `.anyScalar` class in grapheme mode? If we forbid that in grapheme mode then we would always be grapheme-cluster aligned unless a user custom consumer returns a non-aligned index, and we could always forbid that as well

rctcwyvrn: As for the test I don't think we could ever have a non-grapheme aligned `subjectBounds` or `searchBounds` since we always handle creating and stepping the bounds

```
    var low = searchBounds.lowerBound
    let high = searchBounds.upperBound
    while true {
      if let m: Regex<Output>.Match = try _match(
        input, from: low, using: &cpu
      ) {
        return m
      }
      if low >= high { return nil }
      if graphemeSemantic {
        input.formIndex(after: &low)
      } else {
        input.unicodeScalars.formIndex(after: &low)
      }
      cpu.reset(currentPosition: low)
    }
```

Sources/_StringProcessing/ConsumerInterface.swift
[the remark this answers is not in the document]
milseman: I'd like to avoid emitting backtracking traffic for custom character classes if we can. However, since there's a compilation boundary around each member anyways, that might not account for much. Let's keep this for now and try to come back to it again

Sources/_StringProcessing/Engine/MEBuiltins.swift
milseman: That's right, there should be a grapheme boundary assertion whenever stepping up into grapheme semantic mode.

Sources/_StringProcessing/ConsumerInterface.swift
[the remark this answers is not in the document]
rctcwyvrn: Another thing to consider is that if we keep non-ascii custom character classes as one big consumer we're unable to access any optimized matching for any of its members.

Also just no longer having all the matching logic exactly duplicated across `ConsumerInterface` and `Processor` seems like something pretty important for future extensibility/preventing bugs

rctcwyvrn:
@swift-ci test

PR #554, opened 2022-07-06
Fix `_testDSLCaptures`

hamishknight:
Previously we would ignore the case where the match fails, but the test expects the match to succeed.


hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@Azoy Looks like a couple of tests fail with this change, can you take a look? Filed https://github.com/apple/swift-experimental-string-processing/issues/555

PR #556, opened 2022-07-06
[5.7] Fix `_testDSLCaptures`

hamishknight:
*5.7 cherry-pick of https://github.com/apple/swift-experimental-string-processing/pull/554*

Previously we would ignore the case where the match fails, but the test expects the match to succeed.


hamishknight:
@swift-ci please test

stephentyrone (approved):

PR #559, opened 2022-07-07
Validate optimizations when a match fails

hamishknight:
This allows us to catch the case where a match occurs without optimizations, but doesn't occur with optimizations. Additionally fix the `xfail` param such that it can't be used on tests that actually match expectations.

### DIFF
--- a/Tests/RegexTests/MatchTests.swift
+++ b/Tests/RegexTests/MatchTests.swift
@@ -25,23 +25,33 @@ func _firstMatch(
   input: String,
   validateOptimizations: Bool,
   syntax: SyntaxOptions = .traditional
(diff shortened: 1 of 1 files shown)


hamishknight:
@swift-ci please test

rctcwyvrn (approved):
I ran into this same issue while working on https://github.com/apple/swift-experimental-string-processing/pull/525 and ended up implementing a similar change https://github.com/apple/swift-experimental-string-processing/pull/525/files#diff-9420a6ee27325155caaa6f4ae0692789b51ae1066736319212df0fcc52c67dbf

Yours covers more cases and fixes the xfail thing so I'm fine with taking your changes over mine

Issue #558, opened 2022-07-06
Null alternation causes infinite loop


rctcwyvrn:
The first regex runs instantly as expected, but the second one appears to run forever. The first has `@` in the alternation while the second has it outside as an anchor. 

Running them on regex101 shows that their engines run the two regexes in a similar number of cycles and match the same results.

```
  func testInfiniteLoop() {
    print("Starting regex")
    // let regex = try! Regex(#"(?:\d|\w|\.|-|_|%|\+|@)+"#)
    let regex = try! Regex(#"(?:\d|\w|\.|-|_|%|\+|)+@(?:\d|\w|\.|-|_|%|\+|)+"#)
    print(try! regex.firstMatch(in: "lily.test.email@test.com"))
    print("Finished regex")
  }
```

rctcwyvrn:
There was a typo in my regex, the last alternation in `(?:\d|\w|\.|-|_|%|\+|)` is empty and that seems to be the issue 

This regex runs correctly `#"(?:\d|\w|\.|-|_|%|\+)+@(?:\d|\w|\.|-|_|%|\+)+"#`

This regex loops forever `#"(\w|)+"#`

rctcwyvrn:
The loop is caused by `DSLTree.Node.empty` emitting no instructions, resulting in the alternation instruction jumping to the instruction after the alternation instead of the second alternative.

More precisely, what we currently do is

For each alternation except for the last one
1. Emit a save point to the next alternation 
2. Emit instructions for the current alternation
3. Branch to end
4. Loop 

Then for the last alternation just emit it's instructions (if it fails, the entire alternation fails, if it succeeds just fall through to the next block)

However if we have an empty node as the last block it emits no instructions, so the previous alternation's save point doesn't save to the final alternation but rather the first instruction AFTER the alternations. In the example regex this ends up being a `splitSaving` instruction emitted by the quantification instruction, which is intended to attempt to eagerly match as much as possible.

So what happens is that the first alternation (`\w`) fails, it tries to jump to the last alternation (`.empty`), but instead it jumps to the `splitSaving` instruction which is normally only reached if matching was successful, so it tries to match again and loops forever.

The solution is simple: Filter out any `.empty` nodes from the alternation before we emit them

rctcwyvrn:
There are many other edge cases that result in the same issue as this empty alternative
- `"(?:\w|(?i-i:))+"` last alternative is a non-capturing group that only changes matching options
[...]

Issue #542, opened 2022-07-05
`(a*)*` spins the matching engine forever


hamishknight:
Attempting to match the input `"a"` against the pattern `(a*)*` appears to spin the matching engine forever.

rdar://96461197

hamishknight:
cc @milseman

milseman:
This is a common/natural fall out of simple backtracking. We'll want to eventually have loop analysis over an IR to avoid it, but there might be a shorter term fix we can do.

PR #560, opened 2022-07-07
Break out of quantification loop if there is no forward progress

rctcwyvrn:
If we have an inner node of a quantification that doesn't progress our position in the input, we simply loop forever https://github.com/apple/swift-experimental-string-processing/issues/558

The cases where this can happen range from nefarious (purposefully adding groups around instructions that cannot normally be quantified) to cases that are fairly easy to accidentally hit by making a typo (like the original case in the issue where there was an empty alternation case).

The fix is to add a check to quantification to ensure that our position has progressed, however this comes at a performance cost so this PR also adds an optimization to skip these checks if we can ensure that the inner node will have forward progress

Resolves https://github.com/apple/swift-experimental-string-processing/issues/558 and resolves https://github.com/apple/swift-experimental-string-processing/issues/542
rdar://96461197

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -652,23 +652,25 @@ fileprivate extension Compiler.ByteCodeGen {
     //   branch min-trip-count
     builder.label(loopBody)
 
-    var positionReg: PositionRegister? = nil
     // if we aren't sure if the child node will have forward progress and
-    // we have an unbounded quantification,
-    if (!optimizationsEnabled || !child.guaranteesForwardProgress),
-       extraTrips == nil
-    {
-      positionReg = builder.makePositionRegister()
-      builder.buildMoveCurrentPosition(into: positionReg!)
+    // we have an unbounded quantification
+    let startPosition: PositionRegister?
+    let emitPositionChecking =
+      (!optimizationsEnabled || !child.guaranteesForwardProgress) &&
+      extraTrips == nil
+
+    if emitPositionChecking {
+      startPosition = builder.makePositionRegister()
+      builder.buildMoveCurrentPosition(into: startPosition!)
+    } else {
+      startPosition = nil
     }
     try emitNode(child)
-    if (!optimizationsEnabled || !child.guaranteesForwardProgress),
-       extraTrips == nil
-    {
+    if emitPositionChecking {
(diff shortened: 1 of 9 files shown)


rctcwyvrn:
Additional edge cases to consider
-  `(?:\d|(?i)){3}a` a non-progressing adjustment to matching options. The adjusted matching options stay within that scope so it shouldn't affect the outside, we don't need to match it a specific number of times
- A non-progressing custom consumer/matcher that contains state and expects to be called a certain number of times. Is this part of the API that we provide? Should we only break out of the quantification if it's both unlimited and not progressing?

hamishknight:
Nice! Does this also resolve https://github.com/apple/swift-experimental-string-processing/issues/542?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -646,7 +651,26 @@ fileprivate extension Compiler.ByteCodeGen {
     //   <subexpression>
     //   branch min-trip-count
     builder.label(loopBody)
+
+    var positionReg: PositionRegister? = nil
```
milseman, line 655: Might be cleaner to say something like

```swift
  let startPosition: PositionRegister?
  if ... {
    startPosition = builder.makePositionRegister()
  } else {
    startPosition = nil
  }

  if let startPos = startPosition {
    ...
  }
  try emitNode(child)
  if let startPos = startPosition {
    ...
  }
```

Sources/_StringProcessing/Engine/Registers.swift
```diff
@@ -120,11 +130,15 @@ extension Processor.Registers {
 
     self.values = Array(
       repeating: SentinelValue(), count: info.values)
+    self.positions = Array(
+      repeating: Processor.Registers.sentinelIndex,
+      count: info.positions)
```
milseman, line 135: Future: Why the start index instead of the provided sentinel? Should or could we construct an invalid index and at least assert that we're never loading that?

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/Registers.swift
rctcwyvrn: `SentinelValue` only works on the value registers since they are `[Any]`. For now we only emit position registers in this one case, if/when we have more complicated cases in the future we'll want some kinda validation like that

PR #561, opened 2022-07-08
Store source locations of builder expressions for regex compilation errors

rxwei:
Store source locations using `buildExpression(_:)` and print locations upon regex compilation errors. The regex builder DSL proposal already includes a `buildExpression(_:)` that stores source locations in a `RegexComponentBuilder.Component<R>` type.

Resolves rdar://96119443.

----

Example:

```swift
let a = Reference(Substring.self)
let regex = Regex {
  "abc"
  a
  OneOrMore(a)
}
_ = "abc".firstMatch(of: regex)
```

```console
// Fatal error: ... Found a reference used before it captured any match. Used at:
// - .../Tests/RegexBuilderTests/RegexDSLTests.swift@testBackreference():1126:10
// - .../Tests/RegexBuilderTests/RegexDSLTests.swift@testBackreference():1127:10
```

### DIFF
--- a/Tests/RegexBuilderTests/RegexDSLTests.swift
+++ b/Tests/RegexBuilderTests/RegexDSLTests.swift
@@ -1118,6 +1118,22 @@ class RegexDSLTests: XCTestCase {
         }
       }
     }
+
+    do {
+       let a = Reference(Substring.self)
+       let _ = Regex {
+         "abc"
+         a
+         OneOrMore(a)
+       }
+       // Run the following line to trigger the runtime compilation error:
+       // _ = "abc".firstMatch(of: regex)
+       //
+       // Fatal error: ... Found a reference used before
+       // it captured any match. Used at:
+       // - .../Tests/RegexBuilderTests/RegexDSLTests.swift@testBackreference():1126:10
+       // - .../Tests/RegexBuilderTests/RegexDSLTests.swift@testBackreference():1127:10
+     }
   }
 
   struct SemanticVersion: Equatable {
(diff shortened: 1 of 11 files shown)


rxwei:
@swift-ci please test

rxwei:
@swift-ci please test

rxwei:
I also made these builder methods AEIC as they might (and are easy to) change depending on how the proposal review goes.

Tests/RegexBuilderTests/RegexDSLTests.swift
```diff
@@ -1118,6 +1118,22 @@ class RegexDSLTests: XCTestCase {
...
+         OneOrMore(a)
+       }
+       // Run the following line to trigger the runtime compilation error:
+       // _ = "abc".firstMatch(of: regex)
+       //
+       // Fatal error: ... Found a reference used before
+       // it captured any match. Used at:
+       // - .../Tests/RegexBuilderTests/RegexDSLTests.swift@testBackreference():1126:10
+       // - .../Tests/RegexBuilderTests/RegexDSLTests.swift@testBackreference():1127:10
+     }
```
hamishknight, line 1136: Seems unfortunate we don't have a way to test this, would it be worth creating a wrapper similar to `_compileRegex` to test it?

hamishknight (approved):

rxwei:
It's probably not a good idea to store source locations in the tree regardless of the client code's optimization mode.

Perhaps there should be a variant of buildExpression that will be called only when the client code is built in debug mode.

PR #565, opened 2022-07-08
Fix scalar matching in grapheme semantic mode

### LINKED ISSUE #563: UnicodeScalar as a RegexComponent uses scalar semantics even under grapheme semantic mode
The following produces a match:

```swift
let r = Regex {
  "a" as UnicodeScalar
}

print(try! r1.firstMatch(in: "a\u{301}"))
```

however under grapheme semantic mode, it shouldn't. This also affects the `CharacterClass.anyOf` API, e.g:

```swift
let r = Regex {
  CharacterClass.anyOf(["a" as UnicodeScalar, "👍"])
}
print(try! r.firstMatch(in: "a\u{301}"))
```

also incorrectly produces a match (the use of ` "👍"` is to defeat an optimization that would otherwise have the correct semantics).

rdar://96662651

### LINKED ISSUE #564: Preserve scalar syntax in DSL conversion
When attempting to convert `/\u{B4}/` to the DSL, you get:

```swift
let r = Regex {
  "´"
}
```

we ought to try and preserve the users original syntax and use `\u{B4}` instead.

hamishknight:
Previously we would allow consuming a single scalar of a grapheme in grapheme semantic mode when matching a DSL `.scalar`. Change this behavior such that it is treated the same as a character with a single scalar. That is:

- In grapheme mode it must match an entire grapheme.
- In scalar semantic mode, it may match a single scalar.

Additionally, start converting AST `.scalar` atoms into `.scalar` DSL atoms. This is both for consistency with the DSL, and allows us to preserve the `\u{...}` syntax in more cases when a DSL conversion is performed.

Resolves #563
Resolves #564
rdar://96662651

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -62,7 +62,10 @@ fileprivate extension Compiler.ByteCodeGen {
       try emitCharacter(c)
 
     case let .scalar(s):
-      try emitScalar(s)
+      // A scalar always matches the same as a single scalar character. This
+      // means it must match a whole grapheme in grapheme semantic mode, but
+      // can match a single scalar in scalar semantic mode.
+      try emitCharacter(Character(s))
 
     case let .assertion(kind):
       try emitAssertion(kind.ast)
@@ -244,8 +247,12 @@ fileprivate extension Compiler.ByteCodeGen {
       }
     }
   }
-  
-  mutating func emitScalar(_ s: UnicodeScalar) throws {
+
+  /// Emit a consume of a single scalar value. This must only be used in scalar
+  /// semantic mode.
+  mutating func emitConsumeScalar(_ s: UnicodeScalar) throws {
+    assert(options.semanticLevel == .unicodeScalar, "Wrong semantic level")
+
     // TODO: Native instruction buildMatchScalar(s)
     if options.isCaseInsensitive {
       // TODO: e.g. buildCaseInsensitiveMatchScalar(s)
@@ -263,7 +270,7 @@ fileprivate extension Compiler.ByteCodeGen {
     // Unicode scalar matches the specific scalars that comprise a character
     if options.semanticLevel == .unicodeScalar {
       for scalar in c.unicodeScalars {
-        try emitScalar(scalar)
+        try emitConsumeScalar(scalar)
       }
       return
     }
--- a/Sources/_StringProcessing/ConsumerInterface.swift
+++ b/Sources/_StringProcessing/ConsumerInterface.swift
@@ -53,6 +53,45 @@ extension DSLTree._AST.Atom {
   }
 }
 
+extension Character {
+  func generateConsumer(
+    _ opts: MatchingOptions
+  ) throws -> MEProgram.ConsumeFunction? {
+    let isCaseInsensitive = opts.isCaseInsensitive
+    switch opts.semanticLevel {
+    case .graphemeCluster:
+      return { input, bounds in
+        let low = bounds.lowerBound
+        if isCaseInsensitive && isCased {
+          return input[low].lowercased() == lowercased()
+            ? input.index(after: low)
+            : nil
+        } else {
+          return input[low] == self
+            ? input.index(after: low)
+            : nil
+        }
+      }
+    case .unicodeScalar:
+      // TODO: This should only be reachable from character class emission, can
+      // we guarantee that? Otherwise we'd want a different matching behavior.
+      let consumers = unicodeScalars.map { s in consumeScalar {
+        isCaseInsensitive
+          ? $0.properties.lowercaseMapping == s.properties.lowercaseMapping
+          : $0 == s
+      }}
+      return { input, bounds in
+        for fn in consumers {
+          if let idx = fn(input, bounds) {
+            return idx
+          }
+        }
+        return nil
+      }
+    }
+  }
+}
+
 extension DSLTree.Atom {
   var singleScalarASCIIValue: UInt8? {
     switch self {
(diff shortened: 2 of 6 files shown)


rctcwyvrn:
I'm still not convinced that we should have this scalar api while in grapheme semantics at all, converting it to a matching a character under the hood seems unintuitive from a user's perspective. What does a scalar RegexComponent even mean then? If it's given a single scalar character it'll match it as a character and if it's given something that isn't a character (like `\u{301}`) it'll just never match anything? Why even have this option in grapheme semantic mode then? Seems like strange behaviour to me

Would it be possible for us to instead throw an error if we have a `.scalar` while in grapheme semantic mode?

hamishknight:
> converting it to a matching a character under the hood seems unintuitive from a user's perspective

I don't disagree with this, though I should note that it's equivalent to using a `/\u{...}/` regex literal in place of the unicode scalar. IMO the current behavior of potentially splitting graphemes is also somewhat surprising (especially in a character class where some elements could match only a scalar, others could match a character), and isn't something I'd expect in grapheme mode.

> Why even have this option in grapheme semantic mode then? Seems like strange behaviour to me

Yeah, I'm starting to think we should maybe consider retracting these scalar APIs on RegexBuilder until we work out these details (there's only 2 of them AFAIK, the `RegexComponent` conformance + `CharacterClass.anyOf` overload). I think we'd still want this change to the matching engine tho, as `/\u{...}/` should still match a grapheme IMO.

> Would it be possible for us to instead throw an error if we have a .scalar while in grapheme semantic mode?

Yeah, that would be a possible option, though it would surface as a runtime crash on the first call to a matching function.

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -62,7 +62,10 @@ fileprivate extension Compiler.ByteCodeGen {
       try emitCharacter(c)
 
     case let .scalar(s):
-      try emitScalar(s)
+      // A scalar always matches the same as a single scalar character. This
+      // means it must match a whole grapheme in grapheme semantic mode, but
+      // can match a single scalar in scalar semantic mode.
+      try emitCharacter(Character(s))
```
milseman, line 68: Shouldn't this be conditional on whether we're in grapheme semantics or scalar semantics?

Sources/_StringProcessing/ConsumerInterface.swift
```diff
@@ -72,44 +111,15 @@ extension DSLTree.Atom {
...
     case let .scalar(s):
-      return consumeScalar {
-        isCaseInsensitive
-          ? $0.properties.lowercaseMapping == s.properties.lowercaseMapping
-          : $0 == s
-      }
+      // A scalar always matches the same as a single scalar character. This
+      // means it must match a whole grapheme in grapheme semantic mode, but
+      // can match a single scalar in scalar semantic mode.
+      return try Character(s).generateConsumer(opts)
```
milseman, line 122: This seems pretty convoluted. Why do we not just emit a scalar consuming instruction in scalar semantic mode and  set the grapheme boundary bit? CC @rctcwyvrn to see if we need to get some of her instructions in first.

Tests/RegexBuilderTests/RegexDSLTests.swift
```diff
@@ -1120,6 +1120,45 @@ class RegexDSLTests: XCTestCase {
...
+    let r3 = Regex {
+      "👨" as UnicodeScalar
+      "\u{200D}" as UnicodeScalar
+      "👨" as UnicodeScalar
+      "\u{200D}" as UnicodeScalar
+      "👧" as UnicodeScalar
+      "\u{200D}" as UnicodeScalar
+      "👦" as UnicodeScalar
+    }
+    XCTAssertNil(try r3.firstMatch(in: "👨‍👨‍👧‍👦"))
```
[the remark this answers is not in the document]
milseman: The scalars should get globbed together into a stream of scalars, over which grapheme breaking can occur. This is similar to scalar escaped inside string literals. For the DSL, I'd treat them like a stream of scalars ala `/\u{...}\u{...}\u{...}`.

milseman (approved):
LGTM, but I think we can clean up the code gen path a little.

Sources/_StringProcessing/ByteCodeGen.swift
hamishknight: `emitCharacter` will call `emitConsumeScalar` for scalar mode

Sources/_StringProcessing/ConsumerInterface.swift
hamishknight: Can we do that within a consumer matching function? I know we eventually want to migrate off consumer functions, but for now this will generate a `consumeScalar` consumer in scalar semantic mode

rctcwyvrn: This path should only be reached for non-ascii custom character classes so I don't think we can emit an instruction instead of a consume fn in that case. Maybe we could if we made custom character classes emit in the same way we do alternations?

Sources/_StringProcessing/Regex/ASTConversion.swift
```diff
@@ -216,7 +216,7 @@ extension AST.Atom {
 
     switch self.kind {
     case let .char(c):                    return .char(c)
-    case let .scalar(s):                  return .char(Character(s.value))
+    case let .scalar(s):                  return .scalar(s.value)
```
rctcwyvrn, line 219: Do we want the behaviour of regex to be different depending on if we input a unicode value as its scalar code or as a literal character? ie `"ợ"` vs `"\u{1ee3}"`? With this one would be interpreted as a `.char` and the other as a `.scalar` in the DSLTree. I think the intent for the original code here was to make those two inputs equivalent

hamishknight: I think they should be semantically equivalent (i.e they go through the exact same code paths in byte code emission), but this distinction allows us to have more accurate printing of the DSL tree when performing a literal -> DSL conversion

rctcwyvrn: If they behave the same then why would we need to print them differently? Printing them differently would imply that they behave differently I think.

Also yea, seeing as we're moving towards making a `.scalar` behave exactly the same as `.char`, maybe it would make sense to remove it outright

hamishknight: > If they behave the same then why would we need to print them differently? Printing them differently would imply that they behave differently I think.

For example, if the user wrote `/e\u{301}/`, ideally the transformed DSL tree would be:

```swift
Regex {
  "e\u{301}"
}
```

instead of:

```swift
Regex {
  "é"
}
```

Semantically they're the same, but IMO it's much nicer to preserve the original syntax the user wrote.

milseman:
> Should a single scalar composed value match the decomposed version while in grapheme mode?

Yes, in grapheme cluster semantics we do Unicode Canonical Equivalence.

rctcwyvrn:
Ok so https://github.com/apple/swift-experimental-string-processing/pull/525, is ready to merge and up to date with this branch. Should we go ahead and merge that one?

@hamishknight were you planning on doing additional work on combining scalars together?

hamishknight:
@rctcwyvrn I'm happy with landing both, I can work on the scalar combination logic in a follow-up

milseman:
Let's merge this whenever you're ready @rctcwyvrn, and we can coordinate how to land @hamishknight's work on top and cherry-pick things together

hamishknight:
@swift-ci please test

rctcwyvrn:
Changes merged in https://github.com/apple/swift-experimental-string-processing/commit/77520471139d4f9fe780b0814ef4d8787807c461

PR #577, opened 2022-07-14
[Optimization] Specialized quantification instruction

rctcwyvrn:
The idea is to have an optimized quantification instruction that fuses match + save + loop into one instruction
- Fused instructions are usually good optimizations
- This allows us to compact the save point representation of a quantified non-capturing single grapheme match into a range of indices to restore to
- By moving the hottest paths out of `Processor.cycle()` this will hopefully help with the instruction layout work that is soon to come. Some instructions like `splitSaving` or `condBranchIfZeroElseDecrement` might become much less common if most quantifications end up just being one `quantify` instruction

The plan: Compile the most common quantification cases like `.*`, `\w*`, `[a-z0-9]*` and `a*` into this fused instruction

Results:
```
Comparing against benchmark result file speedy_builtins.json
- NumbersAll                              10.5ms	9.14ms	1.4ms		15.3%
- EmailLookaheadAll                       37.3ms	86.2ms	-48.8ms		-56.7%
- EmailLookaheadNoMatchesAll              39ms	60.9ms	-21.9ms		-36.0%
- EmailLookaheadList                      9.2ms	23.6ms	-14.4ms		-61.0%
- CompilerMessagesAll                     114ms	125ms	-11.5ms		-9.2%
- InvertedCCC                             22.5ms	28.7ms	-6.19ms		-21.6%
- EagarQuantWithTerminalWhole             2.4ms	7.58ms	-5.19ms		-68.4%
- IPv6Address                             4.06ms	7.74ms	-3.68ms		-47.6%
- LinesAll                                3.09ms	6.57ms	-3.47ms		-52.9%
- GraphemeBreakNoCapAll                   6.74ms	10.1ms	-3.4ms		-33.5%
- WordsAll                                22.4ms	24.3ms	-1.94ms		-8.0%
- DiceRollsInTextAll                      60.6ms	62.2ms	-1.55ms		-2.5%
- EmailRFCAll                             49.8ms	51ms	-1.26ms		-2.5%
- CaseInsensitiveCCC                      11.2ms	12.4ms	-1.14ms		-9.2%
- symDiffCCC                              40.1ms	41.1ms	-951µs		-2.3%
- AnchoredNotFoundWhole                   8.66ms	9.44ms	-789µs		-8.4%
- BasicRangeCCC                           10.6ms	11.3ms	-723µs		-6.4%
- EmojiRegexAll                           70.7ms	71.4ms	-677µs		-0.9%
- DiceNotation                            6.93ms	7.55ms	-622µs		-8.2%
- BasicCCC                                10.1ms	10.7ms	-531µs		-5.0%
- MACAddress                              3.11ms	3.54ms	-431µs		-12.2%
- SubtractionCCC                          15.5ms	15.9ms	-397µs		-2.5%
- CssAll                                  3.87ms	4.19ms	-324µs		-7.7%
```

- Large improvements in quantifications that consume more characters (`LinesAll`, `EmailLookahead`, `EagerQuantWithTerminalWhole`) as expected
- Only small improvements from the compact save point representation (`AnchoredNotFoundWhole`) due to being killed by ARC in `signalFailure()`
- Regression in quantifications that consume few characters/usually match nothing (`NumbersAll`) due to the increased overhead

Generally good improvements but if we could speed up `signalFailure()` some of these benchmarks would be much faster, there were many cases of the quantification part of the regex getting faster but being much slower in `signalFailure()` due to a combination of ARC and having to call `index(before:)`

Note: based on top of #547


Sources/_StringProcessing/Engine/MEQuantify.swift
```diff
@@ -0,0 +1,128 @@
+extension Processor {
+  func _doQuantifyMatch(_ payload: QuantifyPayload) -> Input.Index? {
+    // Future work: Do we want to rework our Processor.Cycle() switch loop
+    // to do something like this for all of the matching instructions?
+    // ie: A bunch of _doMatchThing functions that return Input.Index?
+    // which we then signalFailure if nil or currentPosition = next otherwise
+    // This would have the benefit of potentially allowing us to not duplicate
+    // code between the normal matching instructions and this loop here
+    var next: Input.Index?
```
rctcwyvrn, line 9: Future work: Do we want to rework our Processor.Cycle() switch loop to do something like this for all of the matching instructions?
ie: A bunch of _doMatchThing functions that return `Input.Index?` which we then `signalFailure `if nil or `currentPosition = next` otherwise
This would have the benefit of potentially allowing us to not duplicate code between the normal matching instructions and this switch here

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -921,4 +880,51 @@ extension DSLTree.Node {
     default: return false
     }
   }
+
+  /// If the given node can be wrapped in a .quantify instruction
+  /// Currently this is conservative to reduce the coupling in ByteCodeGen between the normal case and
+  /// the quantified cases
+  ///
+  /// Essentially we trade off implementation complexity for runtime speed by adding more true cases to this
+  func shouldDoFastQuant(_ opts: MatchingOptions) -> Bool {
```
rctcwyvrn, line 889: // Future work: Should we allow ConsumeFunctions into .quantify?
      // This would open up non-ascii custom character classes as well as the
      // possibility of wrapping weirder cases into consume functions
      // (allowing us to .quantify anything we want, but increasing our reliance on ConsumerInterface)

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/MEQuantify.swift
```diff
@@ -0,0 +1,122 @@
...
+  
+  /// Specialized quantify instruction interpreter for ?
+  mutating func runZeroOrOneQuantify(_ payload: QuantifyPayload) -> Bool {
+    assert(payload.minTrips == 0
+           && payload.extraTrips == 1)
+    let next = _doQuantifyMatch(payload)
+    guard let idx = next else {
+      return true // matched zero times
+    }
+    if payload.quantKind != .possessive {
```
milseman, line 114: Do these work for reluctant? Possessive also never backtracks, so I wonder if (future work) we should consider it entirely separately.

milseman (approved):
Overall LGTM, but can you make sure reluctant quantifications is well tested in the new builtins?

Sources/_StringProcessing/Engine/MEQuantify.swift
rctcwyvrn: Reluctant quantification isn't allowed into `.quantify` because it never loops inside it like the other two so it didn't make sense to add into `.quantify`

In the future we could have a specialized reluctant quantifier instruction that takes in a both the quantification and the match instruction after it, which would handle cases like `.*?;`

Sources/_StringProcessing/Engine/InstPayload.swift
```diff
@@ -341,3 +355,191 @@ extension Instruction.Payload {
...
+  // b38-b35 - Payload type (one of 4 types, stored on 3 bits)
+  // b35-b27 - minTrips (8 bit int)
+  // b27-b18 - extraTrips (8 bit value, one bit for nil)
+  // b18-b16 - Quantification type (one of three types)
+  // b16-b0  - Payload value (depends on payload type)
+  static let quantKindShift: UInt64   = 16
+  static let extraTripsShift: UInt64  = 18
+  static let minTripsShift: UInt64    = 27
+  static let typeShift: UInt64        = 35
+  static let maxStorableTrips: UInt64 = (1 << 8) - 1
```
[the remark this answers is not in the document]
rctcwyvrn: What do you mean by metatype?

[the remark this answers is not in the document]
milseman: `let`s need to be stored somewhere, while computed vars don't. Computed var is Swift's closest `#define` equivalent.

Sources/_StringProcessing/Engine/MEQuantify.swift
milseman: Can you add some asserts then?

Sources/_StringProcessing/ByteCodeGen.swift
rctcwyvrn: Ah right, in that case we'd need a separate `runQuantify` for consumers that would emit save points the normal way

Sources/_StringProcessing/Engine/MEQuantify.swift
rctcwyvrn: Oops, I had some before but they got lost in the other changes

rctcwyvrn: Future work: It should be possible to optimize the pattern of reluctant quantifier + anchor character in a similar style by having a payload of two int registers that store the respective payloads for the quantification and the anchor.

The reason why we can't do this peephole optimization now is that we're stuck with the tree representation so there isn't a good way of determining if the reluctant quantifier is followed by an anchor

milseman (approved):

rctcwyvrn:
@swift-ci test

PR #578, opened 2022-07-14
Internalize `_CharacterClassModel`

hamishknight:
Rip out unused `_CharacterClassModel` API and make it an internal type by removing the last use of `makeDSLTreeCharacterClass`.

### DIFF
--- a/Sources/_StringProcessing/Regex/DSLTree.swift
+++ b/Sources/_StringProcessing/Regex/DSLTree.swift
@@ -740,6 +740,32 @@ extension DSLTree {
     @_spi(RegexBuilder)
     public struct Atom {
       internal var ast: AST.Atom
+
+      // FIXME: The below APIs should be removed once the DSL tree has been
+      // migrated to use proper DSL atoms for them.
+
+      public static var _anyGrapheme: Self {
+        .init(ast: .init(.escaped(.graphemeCluster), .fake))
+      }
+      public static var _whitespace: Self {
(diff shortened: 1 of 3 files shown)


hamishknight:
@swift-ci please test

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -740,6 +740,32 @@ extension DSLTree {
     @_spi(RegexBuilder)
     public struct Atom {
       internal var ast: AST.Atom
+
+      // FIXME: The below APIs should be removed once the DSL tree has been
+      // migrated to use proper DSL atoms for them.
```
milseman, line 745: What does this entail?

milseman (approved):
LGTM, thanks

Sources/_StringProcessing/Regex/DSLTree.swift
hamishknight: Adding them as dedicated cases to `DSLTree.Atom` (or splitting out a new `BuiltinCharacterClass` type perhaps), and then getting the consumer logic working with them. It's probably not too effort, but I'm going to leave it as a follow up for now. I can file an issue to track it.

hamishknight: Filed https://github.com/apple/swift-experimental-string-processing/issues/579

PR #581, opened 2022-07-15
Benchmarker improvements + more benchmarks

rctcwyvrn:
Taking the stuff not waiting for legal approval from https://github.com/apple/swift-experimental-string-processing/pull/515 and adding some more stuff

Benchmarker improvements
- Fixing some benchmark names
- Add suite filtering by patterns instead of names
- Exclude firstMatch benchmarks by default because they were running too quickly and not providing information the matches benchmarks weren't
- Add flag for excluding the NSRegex comparison benchmarks
- Make `save` and `compare` take file paths like a normal CLI instead of the weirdness it was doing before
- Add some basic statistics to the runner (standard deviation computation, t-test to determine notable improvements/regressions)
- Add a benchmark type for running a whole match regex on each item in a list of strings

New benchmarks
- Emojis
- Dice notation
- Swiftc error messages
- ipv4, ipv6, Mac addresses

(If you saw the bit I had here about the interesting list benchmark results, ignore it. It was just a bug :C)


rctcwyvrn:
@swift-ci test

milseman (approved):
LGTM!

milseman:
@swift-ci please test

Sources/RegexBenchmark/BenchmarkRunner.swift
```diff
@@ -37,36 +34,43 @@ public struct BenchmarkRunner {
...
+      }
+      
+      if result.stdev > Stats.maxAllowedStdev {
+        print("Warning: Standard deviation > \(Time(Stats.maxAllowedStdev)) for \(b.name)")
+        print("N = \(samples), median: \(result.median), stdev: \(Time(result.stdev))")
+        print("Rerunning \(b.name)")
+        result = measure(benchmark: b, samples: result.samples*2)
+        print("N = \(result.samples), median: \(result.median), stdev: \(Time(result.stdev))")
+        if result.stdev > Stats.maxAllowedStdev {
+          fatalError("Benchmark \(b.name) is too variant")
```
rxwei, line 60: Should this print the error instead of trapping the program?

rctcwyvrn: Being too variant generally indicates that the system is running stuff in the background and making the benchmarker variant, so in those cases I thought it would probably best to just stop everything and make the user restart it

PR #576, opened 2022-07-13
Move options from RegexComponent to Regex

Azoy:

PR #575, opened 2022-07-13
Rename various APIs

Azoy:
These APIs did not match the proposed spelling, so this changes their names to match what is currently being proposed.

PR #582, opened 2022-07-15
[5.7] Fix API naming mismatches from proposals 

Azoy:
**Explanation**:
In the regex builder proposal, we proposed adding regex options directly on `Regex` vs. on `RegexComponent`. In addition to changing that, the Unicode proposal for string processing proposed using the names `simple` and `default` for Unicode word breaking level 1 and level 2 respectfully.

**Scope**:
This only affects client who are using the current, incorrect, API names as well as using options on `RegexComponent`s.

**Main Branch PR**:
https://github.com/apple/swift-experimental-string-processing/pull/575 and https://github.com/apple/swift-experimental-string-processing/pull/576

**Resolves**:
rdar://97024624

**Risk**:
Low

**Reviewed By**:
@milseman 

**Testing**:
Updated our current tests to use the new API names/altered them somewhat to reflect the options change on `Regex`.

### DIFF
--- a/Sources/RegexBuilder/Anchor.swift
+++ b/Sources/RegexBuilder/Anchor.swift
@@ -104,6 +104,12 @@ extension Anchor {
   ///
   /// This anchor is equivalent to `^` in regex syntax when the `m` option
   /// has been enabled or `anchorsMatchLineEndings(true)` has been called.
+  ///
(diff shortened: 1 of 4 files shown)


milseman (approved):
LGTM

PR #590, opened 2022-07-26
Emit custom character classes like an alternation

rctcwyvrn:
I thought this would be relatively easy to implement and I was curious what the performance benefits were so I implemented it quickly

A number of regressions in benchmarks without any code that hits any of the changes (most of these don't even have a custom character class). Most likely due to the instruction I added? 

As expected `EmailRFCNoMatches` is much faster now that it's non-ascii CCC has it's ascii members collected into a bitset

The big benefit is being able to remove all the code in `ConsumerInterface` that did matching on things we already have instructions for (characters, scalars, case insensitive, etc)

Based on https://github.com/apple/swift-experimental-string-processing/pull/547
```
Comparing against benchmark result file before.json
- EmailLookaheadAll                       88.6ms	85.8ms	2.84ms		3.3%
- DiceRollsInTextAll                      68.7ms	66ms	2.66ms		4.0%
- EmailLookaheadNoMatchesAll              63.2ms	61.1ms	2.09ms		3.4%
- BasicBuiltinCharacterClassAll           16.2ms	14.9ms	1.32ms		8.9%
- EmailLookaheadList                      24.5ms	23.4ms	1.04ms		4.4%
- EmailBuiltinCharacterClassAll           26.2ms	25.1ms	1.02ms		4.1%
- NumbersAll                              11.3ms	10.3ms	986µs		9.6%
- InvertedCCC                             29.6ms	28.9ms	668µs		2.3%
- WordsAll                                26.3ms	25.9ms	411µs		1.6%
- AnchoredNotFoundWhole                   9.73ms	9.41ms	328µs		3.5%
- EmailRFCNoMatchesAll                    62.8ms	106ms	-43.1ms		-40.7%
- symDiffCCC                              19.4ms	40.7ms	-21.4ms		-52.5%
- IntersectionCCC                         12ms	16.2ms	-4.2ms		-26.0%
- SubtractionCCC                          11.7ms	15.7ms	-3.92ms		-25.0%
- EmailRFCAll                             48.5ms	50.6ms	-2.14ms		-4.2%
- EagarQuantWithTerminalWhole             7.75ms	8.41ms	-663µs		-7.9%
```


Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -913,12 +1017,34 @@ extension DSLTree.Node {
...
       return atLeast ?? 0 > 0 && child.guaranteesForwardProgress
     default: return false
     }
   }
 }
+
+extension DSLTree.CustomCharacterClass {
+  /// We allow trivia into CustomCharacterClass, which could result in a CCC that matches nothing
+  /// ie (?x)[ ]
+  var guaranteesForwardProgress: Bool {
```
hamishknight, line 1033: I believe we should reject `(?x)[ ]` at parse time (we should reject a character class with no semantic members). That being said, it is possible to have a custom character class that matches nothing, e.g `[a--a]` (though as I understand it, that would still guarantee forward progress).

natecook1000:
Superseded by #660.

PR #593, opened 2022-08-01
Rename builder parameter name

Azoy:
This renames the component parameter name to `componentBuilder` on overloads taking a RegexBuilder parameter.

Resolves: rdar://97532766


Azoy:
@swift-ci please test

PR #595, opened 2022-08-03
More benchmarker features

rctcwyvrn:
Adds 
- Loading results from a file
- Measuring compile times (via a spi function in `Regex`)
- Comparing compile times and against NSRegularExpression
- Saving comparison results as csv
- Charts (based on #585)


rxwei (approved):
📊

Sources/RegexBenchmark/BenchmarkChart.swift
```diff
@@ -0,0 +1,81 @@
...
+    sampleCount: Int
+  ) -> some ChartContent {
+    // Normalized runtime
+    BarMark(
+      x: .value("Name", name),
+      y: .value("Normalized runtime", new / old))
+    .foregroundStyle(LinearGradient(
+      colors: [.accentColor, new - old <= 0 ? .green : .yellow],
+      startPoint: .bottom,
+      endPoint: .top))
```
rxwei, line 61: You could inline this into the `ForEach { ... }`. I pulled this out to `chartBody` because of "expression too complex". That doesn't happen with just one bar mark.

rctcwyvrn:
@swift-ci test

rctcwyvrn:
Oh huh what version is the CI on?

```
/Users/ec2-user/jenkins/workspace/pr-swift-experimental-string-processing-macos/branch-main/swift-experimental-string-processing/Sources/RegexBenchmark/BenchmarkChart.swift:14:8: error: no such module 'Charts'
import Charts
       ^
```

rxwei:
I don't think the CI runs on macOS 13. Maybe `#if canImport(Charts)` around the import and around `Chart` usage.

Sources/_StringProcessing/Regex/Core.swift
```diff
@@ -135,6 +135,25 @@ extension Regex {
...
+    do {
+      let _ = try Compiler(
+        tree: program.tree,
+        compileOptions: program.compileOptions).emit()
+      return true
+    } catch {
+      return false
+    }
+  }
+}
```
milseman, line 155: This kind of approach looks ok to me. I wonder if we can generalize the interface a little bit to give us some future flexibility without needing to add more SPI over time. For example, we could return a (resilient) result type which we can add metrics to in the future (number of instructions, complexity or star-height, etc). We could also have a (resilient) compilation options parameter, such as the proposed force-recompilation behavior which we could extend with more options like control over optimizations or even targeted engine version (e.g. back deployment)

rctcwyvrn:
- Added setting compiler options to the SPI interface
- Added metrics and tracing enabled as options for the compiler 
- Using the new additions, added CLI options for turning on metrics/tracing for benchmarker regexes
- Added parse time measurements to the benchmarker

rctcwyvrn:
@swift-ci test

rctcwyvrn:
@milseman Could you check that `Regex._forceAction` in `Core.swift` matches what you had in mind? Also that it's ABI stable the way it's written right now?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -89,6 +89,8 @@ struct Processor {
   // MARK: Metrics, debugging, etc.
   var cycleCount = 0
   var isTracingEnabled: Bool
+  let shouldMeasureMetrics: Bool
```
milseman, line 92: While well-predicted, do we see any impact from checking this bool inside the processor? Worst case we could make it conditionally compile and the benchmark can enable it.

milseman (approved):
LGTM after making sure to underscore the SPI

Sources/_StringProcessing/Engine/Processor.swift
rctcwyvrn: image

Ah... the answer is quite a lot. Removing both the tracing and metric conditions speeds things up by 10-20% across the board. I'll see if I can figure out the conditional compilation

rctcwyvrn:
image

Made the charts nice and pretty <:

rctcwyvrn:
@swift-ci test

Sources/_StringProcessing/Engine/Processor.swift
rctcwyvrn: After looking around it doesn't seem like swiftPM supports the feature of "define this compilation variable only when being built as a dependency for target X". The only conditions it can enable/disable compilation variables on are platforms and debug/release. Trying to add multiple targets with flags enabled/disabled for `_StringProcessing` ends up with "overlapping source code" 

https://forums.swift.org/t/swiftpm-and-conditional-compilation/36874
https://forums.swift.org/t/compilation-conditions-and-swift-packages/34627/4

rctcwyvrn:
@swift-ci test

rctcwyvrn:
@swift-ci test

rctcwyvrn:
augh I forgot to switch it back to squash... oops

PR #596, opened 2022-08-03
[Optimization] Add instructions for consuming non-newlines and advancing in scalar view

rctcwyvrn:
A simple new instruction to replace the consumers that are used for emitting `.any` and `.anyNonNewline` instructions in scalar and grapheme mode

Makes matching a default options `.` about ~3-5% faster
```
- ReluctantQuantWithTerminalWhole         7.8ms	8.2ms	-402µs		-4.9%
- ReluctantQuantWhole                     11.7ms	12ms	-345µs		-2.9%
```

Doesn't affect a lot of benchmarks in our suite because we mostly have quantified `.`s like `.*` and that ends up in the `.quantify` optimization
image

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -217,27 +217,16 @@ fileprivate extension Compiler.ByteCodeGen {
     case .graphemeCluster:
       builder.buildAdvance(1)
     case .unicodeScalar:
-      // TODO: builder.buildAdvanceUnicodeScalar(1)
-      builder.buildConsume { input, bounds in
-        input.unicodeScalars.index(after: bounds.lowerBound)
-      }
+      builder.buildAdvanceUnicodeScalar(1)
     }
   }
 
   mutating func emitAnyNonNewline() {
     switch options.semanticLevel {
     case .graphemeCluster:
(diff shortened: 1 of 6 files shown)


milseman:
Is this substantially different than a consume builtin where anyNonNewline is a kind of builtin? I.e. is the dedicated opcode just a particular encoding of the operation?

rctcwyvrn:
We could write it like this 
```  
mutating func emitAnyNonNewline() {
    let model = _CharacterClassModel.init(
      cc: .verticalWhitespace,
      options: options,
      isInverted: true)
    builder.buildMatchBuiltin(model: model)
  }
```

But since this falls into the switch statement with all the other builtin cases it ends up being much slower 

```
- ReluctantQuantWhole                     14ms	12ms	2.01ms		16.7%
- ReluctantQuantWithTerminalWhole         10.1ms	8.2ms	1.88ms		22.9%
```

milseman:
Right, so the choice of a different instruction is reorganizing the branching story. There are lots of ways to organize this, e.g. we could branch early for this case or we could even switch over the combination of op-code and builtin (and for non-builtin-taking operations, that would be zero-ed out).

Anyways, I'm fine taking this because it locks in the kind of performance story we will want at the end.

rctcwyvrn:
@swift-ci test

PR #597, opened 2022-08-04
Fix matching of backreferences in scalar mode

rctcwyvrn:
`.backreference` always matched in grapheme semantic mode (using `matchSeq()` which calls `match`), making this test fail

```
    // Backreferences in scalar mode
    // In scalar mode the backreference should not match
    firstMatchTest(#"(.+)\1"#, input: "ée\u{301}", match: "ée\u{301}")
    firstMatchTest(#"(.+)\1"#, input: "ée\u{301}", match: nil, semanticLevel: .unicodeScalar)
```

Fixed this by adding a bit to the payload for back references to check if it was emitted in unicode scalar mode


rctcwyvrn:
@swift-ci test

milseman (approved):
LGTM

PR #598, opened 2022-08-11
Backreferences do not guarantee forward progress

rctcwyvrn:
The regex `(b?)\1*` spins the matching engine forever. This is the same bug as the one in rdar://96461197 but we had missed this case.

rdar://98517792


rctcwyvrn:
@swift-ci test

milseman (approved):
LGTM

PR #600, opened 2022-08-15
Don't assume quoted literals are non-empty during bytecodegen

rctcwyvrn:
When given an empty quoted literal (ie `\Q\E`) the regex engine crashes while attempting an optimization which assumes that the input is non-empty

rdar://98691039


rctcwyvrn:
@swift-ci test

PR #602, opened 2022-08-16
Deprecate `Regex.init(quoting:)`

hamishknight:
This was renamed to `init(verbatim:)`, but we never removed the old overload. Deprecate it, and suggest using `init(verbatim:)` instead.

rdar://98737122


hamishknight:
@swift-ci please test

stephentyrone (approved):

PR #603, opened 2022-08-16
[5.7] Deprecate `Regex.init(quoting:)`

hamishknight:
*5.7 cherry-pick of https://github.com/apple/swift-experimental-string-processing/pull/602*

This was renamed to `init(verbatim:)`, but we never removed the old overload. Deprecate it, and suggest using `init(verbatim:)` instead.

rdar://98737122


hamishknight:
@swift-ci please test

stephentyrone (approved):

PR #606, opened 2022-08-24
Update CMake file list

Azoy:
Just updating the cmake file list for our various libraries with new/removed files.

Hopefully resolves: rdar://96611602


Azoy:
@swift-ci please test

PR #610, opened 2022-10-06
Atomically load the lowered program

### LINKED ISSUE #609: Regex: Inconsistent atomic access
In `Regex/Core.swift`, we have the following code:

```swift
    var loweredProgram: MEProgram {
      if let loweredObject = _loweredProgramStorage as? ProgramBox {
        return loweredObject.value
      }
      let lowered = try! Compiler(tree: tree, compileOptions: compileOptions).emit()
      _stdlib_atomicInitializeARCRef(object: &_loweredProgramStorage, desired: ProgramBox(lowered))
      return lowered
    }
```

This allows regular non-atomic loads of the `_loweredProgramStorage` variable to occur concurrently with atomic updates (cmpxchg) of it. This is explicitly undefined behavior per the Law of Exclusivity:

> Two accesses to the same variable aren't allowed to overlap unless both accesses are reads or both accesses are atomic. -- [SE-0282](https://github.com/apple/swift-evolution/blob/main/proposals/0282-atomics.md)

We _must_ avoid such accesses by converting the nonatomic load into an atomic load instruction:

```swift
    if let ref = _stdlib_atomicLoadARCRef(object: &_loweredProgramStorage) {
      return unsafeDowncast(ref, to: ProgramBox.self)
    }
```

On a secondary note, in case the current thread loses a concurrent update race, this property ought to return the winner program box, not the loser.

natecook1000:
Since we're atomically initializing the compiled program in `Regex.Program`, we need to pair that with an atomic load.

Resolves #609.

### DIFF
--- a/Sources/_StringProcessing/Regex/Core.swift
+++ b/Sources/_StringProcessing/Regex/Core.swift
@@ -95,12 +95,29 @@ extension Regex {
     
     /// The program for execution with the matching engine.
     var loweredProgram: MEProgram {
-      if let loweredObject = _loweredProgramStorage as? ProgramBox {
-        return loweredObject.value
+      /// Atomically loads the compiled program if it has already been stored.
+      func loadProgram() -> MEProgram? {
+        guard let loweredObject = _stdlib_atomicLoadARCRef(object: &_loweredProgramStorage)
+          else { return nil }
+        return unsafeDowncast(loweredObject, to: ProgramBox.self).value
       }
-      let lowered = try! Compiler(tree: tree, compileOptions: compileOptions).emit()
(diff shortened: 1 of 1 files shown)


natecook1000:
@swift-ci Please test

lorentey (approved):
Looks great! Thanks Nate 👍 


(There is a general feeling of uneasiness about using the implicit inout-to-pointer conversion to acquire a stable address for `_loweredProgramStorage` -- I believe it technically counts as mutating access covering the entire call, so the Law of Exclusivity might still be violated. However, I don't think we can currently do anything practical to resolve that*, and in practice it will just get us the stable address in this context.)

(\* Unless we dynamically allocate memory for `_loweredProgramStorage` separately.)

PR #611, opened 2022-10-20
Improve StringProcessing and RegexBuilder documentation

natecook1000:
I think this provides at least basic docs for all the public symbols. Would love help identifying specific places that we need more/better documentation.

I do think we need a document that explains how regex syntax gets turned into a capture type — e.g. `/(a+)(b)?/` -> `(Substring, Substring, Substring?)`.


rxwei (approved):

Azoy (approved):
This is great, thanks Nate!

Sources/RegexBuilder/Anchor.swift
```diff
@@ -147,7 +147,7 @@ extension Anchor {
   ///
   /// Word boundaries are identified using the Unicode default word boundary
   /// algorithm by default. To specify a different word boundary algorithm,
-  /// see the `RegexComponent.wordBoundaryKind(_:)` method.
+  /// see the `RegexComponent/wordBoundaryKind(_:)` method.
```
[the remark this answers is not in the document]
natecook1000: It doesn't seem like linking to symbols in different modules is supported by DocC yet, so I think this should just be in plain code voice.

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
```diff
@@ -11,7 +11,10 @@
 
 @_implementationOnly import _RegexParser
 
-/// A type-erased regex output.
+/// The type-erased, dynamic output of a regular expression match.
+///
+/// When you find a match using regular expression that has `AnyRegexOutput`
+/// as its output type, you can find information about matches by iterating
```
natecook1000, line 17: Looks like I need to finish this comment.

Sources/RegexBuilder/DSL.swift
```diff
@@ -186,13 +218,33 @@ public struct ChoiceOf<Output>: _BuiltinRegexComponent {
...
+  ///   components, each of which can be exclusively matched.
   public init(@AlternationBuilder _ builder: () -> Self) {
     self = builder()
   }
 }
 
 // MARK: - Capture
 
+/// A regex component that saves the matched substring, or a transformed result,
+/// for access in a regular expression match.
```
natecook1000, line 247: Need an example here that demonstrates capturing and disambiguates from `TryCapture`.

Sources/RegexBuilder/DSL.swift
```diff
@@ -205,6 +257,8 @@ public struct Capture<Output>: _BuiltinRegexComponent {
   // Note: Public initializers are currently gyb'd. See Variadics.swift.
 }
 
+/// A regex component that attempts to transform a matched substring, saving
+/// the result if successful and backtracking if the transformation fails.
```
natecook1000, line 261: Same note as for `Capture`.

Sources/RegexBuilder/DSL.swift
```diff
@@ -219,10 +273,10 @@ public struct TryCapture<Output>: _BuiltinRegexComponent {
 
 // MARK: - Groups
 
-/// An atomic group.
+/// A regex component that represents an atomic group.
 ///
-/// This group opens a local backtracking scope which, upon successful exit,
-/// discards any remaining backtracking points from within the scope.
+/// An atomic group opens a local backtracking scope which, upon successful
+/// exit, discards any remaining backtracking points from within the scope.
```
natecook1000, line 279: Need an example here.

Sources/RegexBuilder/DSL.swift
```diff
@@ -235,11 +289,81 @@ public struct Local<Output>: _BuiltinRegexComponent {
 
 // MARK: - Backreference
 
+/// A reference to a captured portion of a regular expresion.
```
xAlien95, line 292:
```suggestion
/// A reference to a captured portion of a regular expression.
```

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
```diff
@@ -20,21 +23,22 @@ public struct AnyRegexOutput {
 
 @available(SwiftStdlib 5.7, *)
 extension AnyRegexOutput {
-  /// Creates a type-erased regex output from an existing match.
+  /// Creates a dynamic regular expression match output from an existing match.
```
milseman, line 26: Are we settling on "dynamic" for this? I thought we were leaning more towards "type-erased" at one point

Sources/_StringProcessing/Documentation.docc/README.md
```diff
@@ -0,0 +1,41 @@
+# ``_StringProcessing/Regex``
```
milseman, line 1: What does this file do?

milseman (commented):
Overall this looks really good, thank you. Were you able to write test cases for the examples given in the docs?

Sources/_StringProcessing/Documentation.docc/README.md
natecook1000: This is for organizing how the different symbols defined in the `Regex` type are organized on the `Regex` documentation page. By default, symbols are grouped in sections by kind: "Initializers", "Instance Properties", etc.

It looks like I didn't change this file name — I'll update that.

natecook1000:
> Were you able to write test cases for the examples given in the docs?

Great point! I have these in a playground locally, so I'll make sure I move them into a new test file.

Sources/_StringProcessing/Regex/AnyRegexOutput.swift
natecook1000: Oh, right — great question. I made this change because while "type-erased" is the way we look at `AnyRegexOutput` as implementors, for `Regex` users the experience of using a `Match<AnyRegexOutput>` is more like it's a dynamic type that they can query. Typically type erasure results in difficulty accessing the underlying values, which isn't as much the case here due to the different `AnyRegexOutput` API.

natecook1000:
@milseman Added tests for all the samples (wow do I wish these were automatic), plus longer docs for `Capture` and `TryCapture` — could you take a look at the framing and detail for these two? Also would love any ideas for demonstrating `Local`…

natecook1000:
@swift-ci Please test

natecook1000:
(Note: The macOS test failure is due to testing with a < 5.7 Swift.)

PR #614, opened 2022-11-18
Fix abstract for `Regex.dotMatchesNewlines(_:)`

amartini51:
The old version looks like it was accidentally duplicated from `anchorsMatchLineEndings(_:)` just below it.

Fixes rdar://102365369


amartini51:
@swift-ci Please test.

natecook1000 (approved):

Issue #613, opened 2022-11-18
matches(of:) and firstMatch(of:) behave differently when multiline is used with negative lookahead


fwgreen:
```Swift
import Foundation

let regex = /(?m)^\b.*(^(?!SERIAL|TT).+)/

let string = """
CONT NEG TESTS - ALL PINS

SERIAL NUMBER: 15      MODULE: 0          Mon Jan  1 03:41:42 2007
TT   STMT PIN MEAS VALUE      FORCING         LESS THAN  GREATER THAN
---- ---- --- --------------- --------------- ---------- ------------
"""

for match in string.matches(of: regex) {
    print(match.output.1)
}

print("")

print(string.firstMatch(of: regex)?.output.1 ?? "")
```

Expected output:
```
CONT NEG TESTS - ALL PINS

CONT NEG TESTS - ALL PINS
```
Actual output:
```
CONT NEG TESTS - ALL PINS
---- ---- --- --------------- --------------- ---------- ------------

CONT NEG TESTS - ALL PINS
```
While removing multiline fixes it, this is only one capture group of a larger regex; and it works as expected, with multiline, in both Java (on my local machine) and the PCRE2 flavor on [Regex101.com](https://regex101.com/r/AV6ASe/1).

natecook1000:
This is caused by the default word-boundary algorithm used by `Regex`, which recognizes [Unicode "default" word boundaries](https://www.unicode.org/reports/tr29/#Word_Boundaries) instead of the "simple" word boundaries used by other regular expression engines. In particular, the default word-boundary algorithm _always_ sees the position at the start and end of a string, and before and after line breaks, as a word boundary, whereas the simple algorithm sees them as word boundaries only when followed (or preceded, as relevant) by a "word character" (i.e. a character that is matched by `\w`).

To get the behavior you're expecting, you can switch to use simple boundaries by calling `.wordBoundaryKind(.simple)` on your regex:

```
let regex = /(?m)^\b.*(^(?!SERIAL|TT).+)/
    .wordBoundaryKind(.simple)
```

PR #616, opened 2022-12-01
Add tests for line start/end word boundary diffs

natecook1000:
The `default` and `simple` word boundaries have different behaviors at the start and end of strings/lines. These tests validate that we have the correct behavior implemented. Related to issue #613.


natecook1000:
@swift-ci Please test

PR #617, opened 2022-12-01
Remove `RegexConsumer` and fix its dependencies

natecook1000:
This eliminates the `RegexConsumer` type and rewrites its users to call through to other, existing functionality on Regex or in the other Algorithms implementations. `RegexConsumer` doesn't take account of the dual subranges required for matching, so it can produce results that are inconsistent with `matches(of:)` and `ranges(of:)`, which were rewritten earlier.

For example, this is the incorrect behavior:

```swift
let regex = /.*\Ax/
let str = " xyz"
print(str.ranges(of: regex))        // []
print(str.firstRange(of: regex))    // range of "x"
```

`regex` shouldn't match, as correctly shown by `ranges(of:)`.

rdar://102841216


itingliu (approved):

natecook1000:
@swift-ci Please test

PR #621, opened 2022-12-15
Set availability for inverted character class test

natecook1000:
This feature depends on running with a Swift 5.7 stdlib, and fails when that isn't available.

### DIFF
--- a/Tests/DocumentationTests/RegexBuilderTests.swift
+++ b/Tests/DocumentationTests/RegexBuilderTests.swift
@@ -19,8 +19,13 @@ import RegexBuilder
 
 class RegexBuilderTests: XCTestCase {}
 
+@available(SwiftStdlib 5.7, *)
 extension RegexBuilderTests {
   func testCharacterClass_inverted() throws {
+    // `CharacterClass` depends on some standard library SPI that is only
+    // available in >= macOS 13. The warning for the next line is spurious.
+    guard #available(macOS 13, *) else { return }
+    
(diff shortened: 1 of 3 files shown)


natecook1000:

natecook1000:
@swift-ci Please test

Tests/DocumentationTests/RegexBuilderTests.swift
```diff
@@ -20,6 +20,7 @@ import RegexBuilder
 class RegexBuilderTests: XCTestCase {}
 
 extension RegexBuilderTests {
+  @available(macOS 13, *)
```
jshier, line 23: I've never had availability like this limit where tests run. I always have to do a runtime check inside the method itself, at least on macOS / Xcode. I'd guess the Obj-C dynamic lookup doesn't see the availability, so it tries to run the test anyway. Is there something special about these tests?

Also, why put all methods in an extension?

natecook1000:
@swift-ci Please test

Tests/DocumentationTests/RegexBuilderTests.swift
natecook1000: Yeah, I thought this would work, but no… 😞

Re: extension, I think I'm cargo culting the test classes in the other test targets.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #626, opened 2023-01-30
Fix output type mismatch with RegexBuilder

### LINKED ISSUE #625: Runtime crash when accessing DSL-composed Regex Match with named capture groups
**Description**
Composing a `Regex` with the DSL using `RegexLiteral`s with named capture groups works, and matching works, but accessing the matches crashes at runtime. 

Using the same patterns in `RegexLiteral` & DSL without the names works fine.

_forum.swift.org: [runtime-crash-regexbuilding-with-regex-literals-containing-named-capture-groups/62539](https://forums.swift.org/t/runtime-crash-regexbuilding-with-regex-literals-containing-named-capture-groups/62539)_

**Steps to reproduce**
Define one or more patterns with named capture groups, and use the DSL to compose a `Regex`.  Match (`whole`, `prefix`, etc.) to a string.  The the pattern matches correctly, but accessing the `Output` causes a runtime crash.

```
enum FrameComponentRegex {
    static let number = /frame \#(?<frame_number>\d+):/
    static let address = /(?<frame_address>\dx\w{16})/
    static let module = /(?<frame_module>\w+\.{0,1}\w+)/
    static let function = /(?<frame_function>.+)/
}

let frame = Regex {
    FrameComponentRegex.number
    One(.whitespace)
    FrameComponentRegex.address
    One(.whitespace)
    FrameComponentRegex.module
    One("`")
    FrameComponentRegex.function
}

let frameString = """
frame apple/swift#10: 0x00007fff2011383a libdispatch.dylib`_dispatch_client_callout + 8
"""

if let match = frameString.wholeMatch(of: frame) {
    print(match) 
/**
Match(anyRegexOutput: _StringProcessing.AnyRegexOutput(
  input: "frame apple/swift#10: 0x00007fff2011383a libdispatch.dylib`_dispatch_client_callout + 8", 
  _elements: [
    _StringProcessing.AnyRegexOutput.ElementRepresentation(optionalDepth: 0, content: Optional((range: Range(Swift.String.Index(_rawBits: 15)..<Swift.String.Index(_rawBits: 4980743)), value: nil)), name: nil, referenceID: nil), 
    _StringProcessing.AnyRegexOutput.ElementRepresentation(optionalDepth: 0, content: Optional((range: Range(Swift.String.Index(_rawBits: 458757)..<Swift.String.Index(_rawBits: 590087)), value: nil)), name: Optional("frame_number"), referenceID: nil), 
    _StringProcessing.AnyRegexOutput.ElementRepresentation(optionalDepth: 0, content: Optional((range: Range(Swift.String.Index(_rawBits: 721159)..<Swift.String.Index(_rawBits: 1900807)), value: nil)), name: Optional("frame_address"), referenceID: nil), 
    _StringProcessing.AnyRegexOutput.ElementRepresentation(optionalDepth: 0, content: Optional((range: Range(Swift.String.Index(_rawBits: 1966343)..<Swift.String.Index(_rawBits: 3080455)), value: nil)), name: Optional("frame_module"), referenceID: nil), 
    _StringProcessing.AnyRegexOutput.ElementRepresentation(optionalDepth: 0, content: Optional((range: Range(Swift.String.Index(_rawBits: 3145733)..<Swift.String.Index(_rawBits: 4980743)), value: nil)), name: Optional("frame_function"), referenceID: nil)
  ]), 
  range: Range(Swift.String.Index(_rawBits: 15)..<Swift.String.Index(_rawBits: 4980743)))
**/
  
let _ = match.output /** Runtime crash
Could not cast value of type 
'(Swift.Substring, Swift.Substring, Swift.Substring, Swift.Substring, Swift.Substring)' (0x7ff848f83d00) 
to 'Swift.Substring' (0x7ff848af00c8).
**/

let (_, number, address, module, function) = match.output /** Compiler exception
Type of expression is ambiguous without more context 
**/
```

**Expected behavior**
It should be possible to access the captures via tuple.  It would be *nice* to also have subscript-based access, but that's a separate issue.

```
enum FrameComponentRegex_Unnamed {
    static let number = /frame \#(\d+):/
    static let address = /(\dx\w{16})/
    static let module = /(\w+\.{0,1}\w+)/
    static let function = /(.+)/
}
if let match = frameString.wholeMatch(of: frame) {
    let (_, number, address, module, function) = match.output
    print("""
    number:     \(number)
    address:    \(address)
    module:     \(module)
    function:   \(function)
    """)
}
/**
number:     10
address:    0x00007fff2011383a
module:     libdispatch.dylib
function:   _dispatch_client_callout + 8
**/
```

**Environment**
- Swift compiler version info `swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)`
- Xcode version info `Version 14.1 (14B47b)`
- Deployment target: `macOS 13.0`

natecook1000:
Some regex literals (and presumably other `Regex` instances) lose their output type information when used in a RegexBuilder closure due to the way the concatenating builder calls are overloaded. In particular, any output type with labeled tuples or where the sum of tuple components in the accumulated and new output types is greater than 10 will be ignored.

Regex internals don't make this distinction, however, so there ends up being a mismatch between what a `Regex.Match` instance tries to produce and the output type of the outermost regex. For example, this code results in a crash, because `regex` is a `Regex<Substring>` but the match tries to produce a `(Substring, number: Substring)`:

```swift
    let regex = Regex {
        ZeroOrMore(.whitespace)
        /:(?<number>\d+):/
        ZeroOrMore(.whitespace)
    }
    let match = try regex.wholeMatch(in: " :21: ")
    print(match!.output)
```

To fix this, we add a new `ignoreCapturesInTypedOutput` DSLTree node to mark situations where the output type is discarded. This status is propagated through the capture list into the match's storage, which lets us produce the correct output type. Note that we can't just drop the capture groups when building the compiled program because (1) different parts of the regex might reference the capture group and (2) all capture groups are available if a developer converts the output to `AnyRegexOutput`.

```swift
    let anyOutput = AnyRegexOutput(match)
    // anyOutput[1] == "21"
    // anyOutput["number"] == Optional("21")
```

Fixes #625. rdar://104823356


natecook1000:
@swift-ci Please test

Sources/RegexBuilder/Variadics.swift
```diff
@@ -571,7 +571,11 @@ extension RegexComponentBuilder {
     accumulated: R0, next: R1
   ) -> Regex<Substring> where R0.RegexOutput == W0  {
     let factory = makeFactory()
-    return factory.accumulate(accumulated, next)
+    if #available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *) {
```
rxwei, line 574: Does `if #available(SwiftStdlib 5.8, *)` work?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -650,7 +680,7 @@ extension DSLTree.Node {
   /// output but forwarding its only child's output.
   var isOutputForwarding: Bool {
     switch self {
-    case .nonCapturingGroup:
+    case .nonCapturingGroup, .ignoreCapturesInTypedOutput:
```
rxwei, line 683: I wonder `wholeMatchType` would ever return the wrong type with this change.

It's probably not output-forwarding because it doesn't always have the same output as the child.

rxwei (approved):

Sources/RegexBuilder/Variadics.swift
natecook1000: No, that isn't allowed in an always-emit function.

natecook1000:
@swift-ci Please test

Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -308,7 +308,11 @@ struct VariadicsGenerator: ParsableCommand {
     output("""
         {
           let factory = makeFactory()
-          return factory.accumulate(accumulated, next)
+          if #available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *) {
+            return factory.accumulate(accumulated, factory.ignoreCapturesInTypedOutput(next))
+          } else {
+            return factory.accumulate(accumulated, next)
+          }
```
milseman, line 315: How about a helper function that will reduce spew and, more importantly, give a name and place to commend why we're doing this and how it differs across versions.

E.g.

```swift
// comment...
private func dropCaptures(_ next: ...) -> ... {
  // ... comment about old and new behavior
  if #available(...) {
   return  ...
  }
  return ...
}

...

return factory.accumulate(accumulated, dropCaptures(next))
```

(or some better name than `dropCaptures`)

Sources/_StringProcessing/Utility/RegexFactory.swift
```diff
@@ -20,6 +20,16 @@ public struct _RegexFactory {
...
+
+  @available(SwiftStdlib 5.8, *)
+  public func ignoreCapturesInTypedOutput(
+    _ child: some RegexComponent
+  ) -> Regex<Substring> {
+    // Don't wrap `child` again if it's a leaf node.
+    child.regex.root.hasChildNodes
+      ? .init(node: .ignoreCapturesInTypedOutput(child.regex.root))
+      : .init(node: child.regex.root)
+  }
```
milseman, line 32: Why is this has-child instead of has-capture? Do we need to wrap this around everything, or is this just part of how capture concatenation works?

milseman (commented):
Overall LGTM after a quick review

Sources/_StringProcessing/Utility/RegexFactory.swift
natecook1000: This branch isn't technically necessary, but it reduces the number of times we apply this wrapper node. I didn't want to do a `hasCapture` check, since that would have to search the whole tree. (But maybe it would be worth it to produce a smaller tree.)

Sources/VariadicsGenerator/VariadicsGenerator.swift
natecook1000: That's a good call, I only ever wrote this once, but it is codegen'd an awful lot.

natecook1000:
@swift-ci Please test

natecook1000:

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Utility/RegexFactory.swift
milseman: Hmm... one of the downsides of using enums for trees is that it's really hard to just add a bit to track `hasCapture` information. 

This looks fine to me for now, the extra nodes in the tree should be transient, though we might want to also track the time spent allocating each node.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #627, opened 2023-01-30
Correct misspelling in DSL renderer

natecook1000:
vertial -> vertical

rdar://104602317

PR #628, opened 2023-02-01
Add type annotations in RegexBuilder tests

natecook1000:
These changes work around a change to the way result builders are compiled that removes the ability for result builder closure outputs to affect the overload resolution elsewhere in an expression.

Workarounds for rdar://104881395 and rdar://104645543

This should fix the CI errors visible in #626 and #627.


natecook1000:
@swift-ci Please test

PR #630, opened 2023-02-07
[Do not merge] Reduced test case for strange test crashes

natecook1000:
This is a very odd bug that is showing up in CI and on the command line, but not in Xcode with a similar toolchain selected. The `GoodBadFunctionTests.swift` file has a reduced test case that demonstrates the problem: With this code included, I see several _other_ tests that crash, even in different test modules.

The problem disappears if you comment out the call to `badFunction`, but then re-appears if you make `badFunction` internal instead of fileprivate. I've been testing this with
`swift-DEVELOPMENT-SNAPSHOT-2023-01-30-a.xctoolchain`.

### DIFF
--- a/Tests/RegexBuilderTests/GoodBadFunctionTests.swift
+++ b/Tests/RegexBuilderTests/GoodBadFunctionTests.swift
@@ -1,0 +1,18 @@
+import XCTest
+import _StringProcessing
+@testable import RegexBuilder
+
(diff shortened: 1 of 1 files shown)


natecook1000:
@swift-ci Please test

shahmishal:
@swift-ci test Windows

PR #636, opened 2023-02-15
Use `some RegexComponent` in variadic inits

natecook1000:
This updates VariadicsGenerator to print declaration with `some RegexComponent<...>` instead of the full generic signature, which will hopefully give the type checker more detailed information to work with about `RegexComponentBuilder` closures.

### DIFF
--- a/Sources/RegexBuilder/DSL.swift
+++ b/Sources/RegexBuilder/DSL.swift
@@ -15,9 +15,9 @@
 @available(SwiftStdlib 5.7, *)
 extension Regex {
   /// Creates a regular expression using a RegexBuilder closure.
-  public init<Content: RegexComponent>(
-    @RegexComponentBuilder _ content: () -> Content
-  ) where Content.RegexOutput == Output {
+  public init(
+    @RegexComponentBuilder _ content: () -> some RegexComponent<Output>
+  ) {
     self = content().regex
   }
 }
@@ -91,9 +91,9 @@ public struct One<Output>: RegexComponent {
   public var regex: Regex<Output>
 
   /// Creates a regex component that matches the given component exactly once.
-  public init<Component: RegexComponent>(
-    _ component: Component
-  ) where Component.RegexOutput == Output {
+  public init(
+    _ component: some RegexComponent<Output>
+  ) {
     self.regex = component.regex
   }
 }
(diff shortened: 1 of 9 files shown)


xedin:
Please test with following pull request:
https://github.com/apple/swift/pull/63677

@swift-ci please test

Sources/RegexBuilder/DSL.swift
```diff
@@ -15,9 +15,9 @@
 @available(SwiftStdlib 5.7, *)
 extension Regex {
   /// Creates a regular expression using a RegexBuilder closure.
-  public init<Content: RegexComponent>(
```
rxwei, line 18: Would this break ABI? Isn't `some RegexComponent<Output>` in the closure return type going to be an opaque return type, not a generic parameter on the parent declaration?

natecook1000: This particular change is just a syntactic transformation of the original. You can get into trouble with multiple generic parameters, since the order is important to the ABI, but luckily all of those are `@_alwaysEmitIntoClient`. (This one should probably be that as well, tbh.)

xedin:
Please test with following pull request:
https://github.com/apple/swift/pull/63677

@swift-ci please test

xedin:
Please test with following pull request:
https://github.com/apple/swift/pull/63677

@swift-ci please test

xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:
@swift-ci please test

xedin:
I guess we have to wait some more...

natecook1000:
Yeah, I don't think the compiler change has landed in the nightly yet, so probably not until later Friday.

natecook1000:
@swift-ci Please test

natecook1000:
Alright, the compiler changes are in the `swift-DEVELOPMENT-SNAPSHOT-2023-02-16-a` toolchain — just need that in CI for this to be ready.

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

PR #650, opened 2023-04-04
Remove the unsupported `anyScalar` case

### LINKED ISSUE #647: Remove the `anyScalar` CustomCharacterClass
`Regex` doesn't support the `\O` "any scalar" meta character, but since it's included in the CCC representation, we have untestable branches in various places with dubious implementations. We should remove it throughout.

natecook1000:
We decided not to support the `anyScalar` character class, which would match a single Unicode scalar regardless of matching mode. However, its representation was still included in the various character class types in the regex engine, leading to unreachable code and unclear requirements when changing or adding new code. This change removes that representation where possible.

The `DSLTree.Atom.CharacterClass` enum is left unchanged, since it is marked `@_spi(RegexBuilder) public`. Any use of that enum case is handled with a `fatalError("Unsupported")`, and it isn't produced on any code path.

Fixes #647.

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -702,9 +702,6 @@ fileprivate extension Compiler.ByteCodeGen {
       case .characterClass(let cc):
         // Custom character class that consumes a single grapheme
         let model = cc.asRuntimeModel(options)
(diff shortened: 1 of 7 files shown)


natecook1000:
@swift-ci Please test

milseman (approved):
LGTM, assuming we don't have any API way of getting `.anyScalar` out.

PR #644, opened 2023-04-03
General ascii fast paths for character classes

milseman:
When the portion of the string being matched is ASCII, use fast ASCII character class membership tests.

```
=== Regressions ======================================================================
- NotFoundAll                             7.12ms	7.03ms	89µs		1.3%
- EagarQuantWithTerminalWhole             2.64ms	2.62ms	18µs		0.7%
=== Improvements =====================================================================
- DiceRollsInTextAll                      50.3ms	64.9ms	-14.6ms		-22.5%
- EmailBuiltinCharacterClassAll           15.7ms	24.8ms	-9.06ms		-36.5%
- WordsAll                                14.8ms	22.5ms	-7.67ms		-34.1%
- BasicBuiltinCharacterClassAll           9.26ms	15.2ms	-5.96ms		-39.2%
- CompilerMessagesAll                     117ms	123ms	-5.91ms		-4.8%
- NumbersAll                              8.07ms	11.9ms	-3.87ms		-32.4%
- DiceNotation                            5.42ms	7.02ms	-1.6ms		-22.8%
- GraphemeBreakNoCapAll                   5.49ms	7ms	-1.51ms		-21.6%
- EmailRFCNoMatchesAll                    137ms	138ms	-1.16ms		-0.8%
- EmailRFCAll                             63ms	64ms	-999µs		-1.6%
- IntersectionCCC                         22.1ms	22.8ms	-671µs		-2.9%
- EmailLookaheadAll                       40.4ms	40.9ms	-487µs		-1.2%
- SubtractionCCC                          21.7ms	22.1ms	-403µs		-1.8%
- EmojiRegexAll                           73.4ms	73.8ms	-403µs		-0.5%
- InvertedCCC                             21.3ms	21.7ms	-351µs		-1.6%
- IPv4Address                             2.58ms	2.88ms	-304µs		-10.6%
- symDiffCCC                              49.4ms	49.7ms	-280µs		-0.6%
- AnchoredNotFoundWhole                   9.08ms	9.26ms	-182µs		-2.0%
- CssAll                                  3.84ms	4.02ms	-177µs		-4.4%
- CaseInsensitiveCCC                      11.9ms	12ms	-154µs		-1.3%
- BasicCCC                                10.7ms	10.8ms	-127µs		-1.2%
- HangulSyllableAll                       6.89ms	7.01ms	-121µs		-1.7%
- BasicRangeCCC                           11.1ms	11.3ms	-115µs		-1.0%
[...]

PR #642, opened 2023-04-02
Speed up general character class matching

milseman:
Speed up the character class matching general path by re-ordering expressions, short-circuiting prior to needing to invoke `Character.isASCII`.

The regressions are minor and inconsistent across runs, while the improvements are robust across runs.

```
=== Regressions ======================================================================
- CompilerMessagesAll                     119ms	118ms	1.27ms		1.1%
- EmailRFCNoMatchesAll                    134ms	133ms	1.12ms		0.8%
- symDiffCCC                              48.6ms	47.8ms	713µs		1.5%
- IntersectionCCC                         22.1ms	21.5ms	613µs		2.9%
- SubtractionCCC                          21.6ms	21ms	612µs		2.9%
- EmojiRegexAll                           71.5ms	71.1ms	362µs		0.5%
- InvertedCCC                             21ms	20.7ms	272µs		1.3%
- BasicCCC                                10.7ms	10.5ms	242µs		2.3%
- CaseInsensitiveCCC                      11.8ms	11.6ms	216µs		1.9%
- BasicRangeCCC                           11ms	10.8ms	202µs		1.9%
- HangulSyllableAll                       6.85ms	6.66ms	191µs		2.9%
- LiteralSearchAll                        6.59ms	6.42ms	164µs		2.6%
- LiteralSearchNotFoundAll                6.37ms	6.25ms	120µs		1.9%
- HangulSyllableFirst                     3.29ms	3.18ms	106µs		3.3%
- AnchoredNotFoundWhole                   8.93ms	8.83ms	103µs		1.2%
- ReluctantQuantWithTerminalWhole         9.07ms	8.99ms	78.9µs		0.9%
- CssAll                                  3.98ms	3.94ms	44.1µs		1.1%
- MACAddress                              3.02ms	2.98ms	35.8µs		1.2%
=== Improvements =====================================================================
- DiceRollsInTextAll                      62.6ms	66ms	-3.39ms		-5.1%
- EmailBuiltinCharacterClassAll           24.3ms	26ms	-1.76ms		-6.8%
- BasicBuiltinCharacterClassAll           14.6ms	15.9ms	-1.28ms		-8.0%
- NumbersAll                              11.6ms	12.6ms	-976µs		-7.8%
- WordsAll                                21.8ms	22.6ms	-823µs		-3.6%
- DiceNotation                            6.74ms	7.08ms	-345µs		-4.9%
- GraphemeBreakNoCapAll                   6.72ms	6.95ms	-234µs		-3.4%
[...]

PR #641, opened 2023-04-02
Update OS versions aliases

milseman:
Some post-release housekeeping: update OS version aliases, allowing a release stdlib to be used in conjunction with the package.

PR #652, opened 2023-04-05
Update 5.9 with state of swift/main

### LINKED ISSUE #609: Regex: Inconsistent atomic access

### LINKED ISSUE #625: Runtime crash when accessing DSL-composed Regex Match with named capture groups

milseman:
* Atomically load the lowered program (#610)

Since we're atomically initializing the compiled program in `Regex.Program`, we need to pair that with an atomic load.

Resolves #609.

* Add tests for line start/end word boundary diffs (#616)

The `default` and `simple` word boundaries have different behaviors at the start and end of strings/lines. These tests validate that we have the correct behavior implemented. Related to issue #613.

* Add tweaks for Android

* Fix documentation typo (#615)

* Fix abstract for Regex.dotMatchesNewlines(_:). (#614)

The old version looks like it was accidentally duplicated from anchorsMatchLineEndings(_:) just below it.

* Remove `RegexConsumer` and fix its dependencies (#617)

* Remove `RegexConsumer` and fix its dependencies

This eliminates the RegexConsumer type and rewrites its users to call through to other, existing functionality on Regex or in the Algorithms implementations. RegexConsumer doesn't take account of the dual subranges required for matching, so it can produce results that are inconsistent with matches(of:) and ranges(of:), which were rewritten earlier.

rdar://102841216

* Remove remaining from-end algorithm methods

This removes methods that are left over from when we were considering from-end algorithms. These aren't tested and may not have the correct semantics, so it's safer to remove them entirely.

* Improve StringProcessing and RegexBuilder documentation (#611)

This includes documentation improvements for core types/methods, RegexBuilder types along with their generated variadic initializers, and adds some curation. It also includes tests of the documentation code samples.

* Set availability for inverted character class test (#621)

This feature depends on running with a Swift 5.7 stdlib, and fails when that isn't available.

* Add type annotations in RegexBuilder tests

These changes work around a change to the way result builders are compiled that removes the ability for result builder closure outputs to affect the overload resolution elsewhere in an expression.

Workarounds for rdar://104881395 and rdar://104645543

* Workaround for fileprivate array issue

A recent compiler change results in fileprivate arrays sometimes not keeping their buffers around long enough. This change avoids that issue by removing the fileprivate annotations from the affected type.

* Fix an issue where named character classes weren't getting converted in the result builder. <rdar://104480703>

* Stop at end of search string in TwoWaySearcher (#631)

When searching for a substring that doesn't exist, it was possible for TwoWaySearcher to advance beyond the end of the search string, causing a crash. This change adds a `limitedBy:` parameter to that index movement, avoiding the invalid movement.

Fixes rdar://105154010

* Correct misspelling in DSL renderer (#627)

vertial -> vertical

rdar://104602317

* Fix output type mismatch with RegexBuilder (#626)

Some regex literals (and presumably other `Regex` instances) lose their output type information when used in a RegexBuilder closure due to the way the concatenating builder calls are overloaded. In particular, any output type with labeled tuples or where the sum of tuple components in the accumulated and new output types is greater than 10 will be ignored.

Regex internals don't make this distinction, however, so there ends up being a mismatch between what a `Regex.Match` instance tries to produce and the output type of the outermost regex. For example, this code results in a crash, because `regex` is a `Regex<Substring>` but the match tries to produce a `(Substring, number: Substring)`:

    let regex = Regex {
        ZeroOrMore(.whitespace)
        /:(?<number>\d+):/
        ZeroOrMore(.whitespace)
    }
    let match = try regex.wholeMatch(in: " :21: ")
    print(match!.output)

To fix this, we add a new `ignoreCapturesInTypedOutput` DSLTree node to mark situations where the output type is discarded. This status is propagated through the capture list into the match's storage, which lets us produce the correct output type. Note that we can't just drop the capture groups when building the compiled program because (1) different parts of the regex might reference the capture group and (2) all capture groups are available if a developer converts the output to `AnyRegexOutput`.

    let anyOutput = AnyRegexOutput(match)
    // anyOutput[1] == "21"
    // anyOutput["number"] == Optional("21")

Fixes #625. rdar://104823356

Note: Linux seems to crash on different tests when the two customTest overloads have `internal` visibility or are called. Switching one of the functions to be generic over a RegexComponent works around the issue.

* Revert "Merge pull request #628 from apple/result_builder_changes_workaround"

This reverts commit 7e059b7bd4e647bc884a71731fcfdc4fffda0700, reversing changes made to 3ca8b134e4c93e8c6d3f7b060b0aefcf06b92039.

* Use `some` syntax in variadics

This supports a type checker fix after the change in how result builder closure parameters are type-checked.

* Type checker workaround: adjust test

* Further refactor to work around type checker regression

* Align availability macro with OS versions (#641)

* Speed up general character class matching (#642)

Short-circuit Character.isASCII checks inside built in character class matching.

Also, make benchmark try a few more times before giving up.

* Test for \s matching CRLF when scalar matching (#648)

* General ascii fast paths for character classes (#644)

General ASCII fast-paths for builtin character classes

* Remove the unsupported `anyScalar` case (#650)

We decided not to support the `anyScalar` character class, which would match a single Unicode scalar regardless of matching mode. However, its representation was still included in the various character class types in the regex engine, leading to unreachable code and unclear requirements when changing or adding new code. This change removes that representation where possible.

The `DSLTree.Atom.CharacterClass` enum is left unchanged, since it is marked `@_spi(RegexBuilder) public`. Any use of that enum case is handled with a `fatalError("Unsupported")`, and it isn't produced on any code path.

---------

### DIFF
--- a/Sources/VariadicsGenerator/VariadicsGenerator.swift
+++ b/Sources/VariadicsGenerator/VariadicsGenerator.swift
@@ -14,7 +14,7 @@
 import ArgumentParser
 #if os(macOS)
 import Darwin
-#elseif os(Linux)
+#elseif canImport(Glibc)
 import Glibc
 #elseif os(Windows)
 import CRT
--- a/Sources/_StringProcessing/Unicode/ASCII.swift
+++ b/Sources/_StringProcessing/Unicode/ASCII.swift
@@ -1,0 +1,169 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+private var _lineFeed: UInt8 { 0x0A }
+private var _carriageReturn: UInt8 { 0x0D }
+private var _lineTab: UInt8 { 0x0B }
+private var _formFeed: UInt8 { 0x0C }
+private var _space: UInt8 { 0x20 }
+private var _tab: UInt8 { 0x09 }
+
+private var _0: UInt8 { 0x30 }
+private var _9: UInt8 { 0x39 }
+private func _isASCIINumber(_ x: UInt8) -> Bool {
+  return (_0..._9).contains(x)
+}
+
+private var _a: UInt8 { 0x61 }
+private var _z: UInt8 { 0x7A }
+private var _A: UInt8 { 0x41 }
+private var _Z: UInt8 { 0x5A }
+
+private var _underscore: UInt8 { 0x5F }
+
+extension UInt8 {
+  var _isASCII: Bool { self < 0x80 }
+
+  // TODO: Bitvectors for the below
+
+  /// Assuming we're ASCII, whether we match `\d`
+  var _asciiIsDigit: Bool {
+    assert(_isASCII)
+    return(_0..._9).contains(self)
+  }
+
+  /// Assuming we're ASCII, whether we match `\h`
+  var _asciiIsHorizontalWhitespace: Bool {
+    assert(_isASCII)
+    return self == _space || self == _tab
+  }
+
+  /// Assuming we're ASCII, whether we match `\v`
+  var _asciiIsVerticalWhitespace: Bool {
+    assert(_isASCII)
+    switch self {
+    case _lineFeed, _carriageReturn, _lineTab, _formFeed:
+      return true
+    default:
+      return false
+    }
+  }
+
+  /// Assuming we're ASCII, whether we match `\s`
+  var _asciiIsWhitespace: Bool {
+    assert(_isASCII)
+    switch self {
+    case _space, _tab, _lineFeed, _lineTab, _formFeed, _carriageReturn:
+      return true
+    default:
+      return false
+    }
+  }
+
+  /// Assuming we're ASCII, whether we match `[a-zA-Z]`
+  var _asciiIsLetter: Bool {
+    assert(_isASCII)
+    return (_a..._z).contains(self) || (_A..._Z).contains(self)
+  }
+
+  /// Assuming we're ASCII, whether we match `\w`
+  var _asciiIsWord: Bool {
+    assert(_isASCII)
+    return _asciiIsDigit || _asciiIsLetter || self == _underscore
+  }
+}
+
+extension String {
+  /// TODO: detailed description of nuanced semantics
+  func _quickASCIICharacter(
+    at idx: Index
+  ) -> (first: UInt8, next: Index, crLF: Bool)? {
+    // TODO: fastUTF8 version
+
+    if idx == endIndex {
+      return nil
+    }
+    let base = utf8[idx]
+    guard base._isASCII else {
+      assert(!self[idx].isASCII)
+      return nil
+    }
+
+    var next = utf8.index(after: idx)
+    if next == utf8.endIndex {
+      assert(self[idx].isASCII)
+      return (first: base, next: next, crLF: false)
+    }
+
+    let tail = utf8[next]
+    guard tail._isSub300StartingByte else { return nil }
+
+    // Handle CR-LF:
+    if base == _carriageReturn && tail == _lineFeed {
+      utf8.formIndex(after: &next)
+      guard next == endIndex || utf8[next]._isSub300StartingByte else {
+        return nil
+      }
+      assert(self[idx] == "\r\n")
+      return (first: base, next: next, crLF: true)
+    }
+
+    assert(self[idx].isASCII && self[idx] != "\r\n")
+    return (first: base, next: next, crLF: false)
+  }
+
+  func _quickMatch(
+    _ cc: _CharacterClassModel.Representation,
+    at idx: Index,
+    isScalarSemantics: Bool
+  ) -> (next: Index, matchResult: Bool)? {
+    /// ASCII fast-paths
+    guard let (asciiValue, next, isCRLF) = _quickASCIICharacter(
+      at: idx
+    ) else {
+      return nil
+    }
+
+    // TODO: bitvectors
+    switch cc {
+    case .any, .anyGrapheme:
+      return (next, true)
+
+    case .digit:
+      return (next, asciiValue._asciiIsDigit)
+
+    case .horizontalWhitespace:
+      return (next, asciiValue._asciiIsHorizontalWhitespace)
+
+    case .verticalWhitespace, .newlineSequence:
+      if asciiValue._asciiIsVerticalWhitespace {
+        if isScalarSemantics && isCRLF && cc == .verticalWhitespace {
+          return (utf8.index(before: next), true)
+        }
+        return (next, true)
+      }
+      return (next, false)
+
+    case .whitespace:
+      if asciiValue._asciiIsWhitespace {
+        if isScalarSemantics && isCRLF {
+          return (utf8.index(before: next), true)
+        }
+        return (next, true)
+      }
+      return (next, false)
+
+    case .word:
+      return (next, asciiValue._asciiIsWord)
+    }
+  }
+}
+
--- a/Documentation/ProgrammersManual.md
+++ b/Documentation/ProgrammersManual.md
@@ -1,0 +1,30 @@
+# Programmer's Manual
+
+## Programming patterns
+
+### Engine quick checks and fast paths
+
+In the engine nomenclature, a quick-check results in a yes/no/maybe while a thorough check always results in a definite answer.
+
+The nature of quick checks and fast paths is that they bifurcate testing coverage. One easy way to prevent this in simple cases is to assert that a definite quick result matches the thorough result.
+
+One example of this pattern is matching against a builtin character class. The engine has a `_matchBuiltinCC`
+
+```swift
+  func _matchBuiltinCC(...) -> Input.Index? {
+    // Calls _quickMatchBuiltinCC, if that gives a definite result
+    // asserts that it is the same as the result of
+    // _thoroughMatchBuiltinCC and returns it. Otherwise returns the
+    // result of _thoroughMatchBuiltinCC
+  }
+
+  @inline(__always)
+  func _quickMatchBuiltinCC(...) -> QuickResult<Input.Index?>
+
+  @inline(never)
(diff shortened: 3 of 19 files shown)


milseman:
@swift-ci please test

Sources/VariadicsGenerator/VariadicsGenerator.swift
```diff
@@ -14,7 +14,7 @@
 import ArgumentParser
 #if os(macOS)
```
stephentyrone, line 15: Seems like this should be `#if os(Darwin)`, even though we're not running it on any non-mac Darwin targets at present.

stephentyrone: (obviously, if we're gonna fix this, fix it on main first)

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -0,0 +1,169 @@
...
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2021-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+private var _lineFeed: UInt8 { 0x0A }
```
stephentyrone, line 12: These would maybe be more readable as `UInt8(ascii: "\n")` etc. Mostly a matter of taste, though.

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -0,0 +1,169 @@
...
+
+extension UInt8 {
+  var _isASCII: Bool { self < 0x80 }
+
+  // TODO: Bitvectors for the below
+
+  /// Assuming we're ASCII, whether we match `\d`
+  var _asciiIsDigit: Bool {
+    assert(_isASCII)
+    return(_0..._9).contains(self)
```
stephentyrone, line 40: Why do we define `_isASCIINumber` if not to use it here?

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -0,0 +1,169 @@
...
+    case _lineFeed, _carriageReturn, _lineTab, _formFeed:
+      return true
+    default:
+      return false
+    }
+  }
+
+  /// Assuming we're ASCII, whether we match `\s`
+  var _asciiIsWhitespace: Bool {
+    assert(_isASCII)
```
stephentyrone, line 62: I expect that we'd get better codegen out of something like:
```
if self > 0x20 { return false }       // all whitespace is 0x20 or lower
return 0x2e00 &>> self & 1 == 1 // table lookup for whitespace chars
```
Ideally LLVM would be clever enough to do this for us, but it isn't.

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -0,0 +1,169 @@
...
+      return true
+    default:
+      return false
+    }
+  }
+
+  /// Assuming we're ASCII, whether we match `[a-zA-Z]`
+  var _asciiIsLetter: Bool {
+    assert(_isASCII)
+    return (_a..._z).contains(self) || (_A..._Z).contains(self)
```
stephentyrone, line 74: Check codgen for this; it should be just an AND or OR or BIC, followed by a subtract and compare. If not, we might need to rewrite it that way explicitly.

Sources/VariadicsGenerator/VariadicsGenerator.swift
milseman: @natecook1000  thoughts?

Sources/_StringProcessing/Unicode/ASCII.swift
milseman: For ASCII character classes, would it be better to have 128-bit bit vectors that we lookup the corresponding ASCII value in?

stephentyrone: 128b bit vectors are quite a bit fussier; since there's no 128b bitwise shifts or comparisons on x86 or arm. Where we can get away with 32b or 64b bit vectors (like here), that's usually a win.

Sources/_StringProcessing/Unicode/ASCII.swift
milseman: Would that be better than a bit vector? 

These improvements are low priority as we also have such overwhelming inefficiencies elsewhere that time spent inside `_quickMatch` is close to negligible. However, I will definitely file something to track future improvements here.

stephentyrone: Yes, better than a bit vector.

milseman:
Testing passed (https://github.com/apple/swift/pull/64937). @stephentyrone good to merge to 5.9 and do these fixes on main?

stephentyrone (approved):

PR #656, opened 2023-04-11
Fix `firstRange(of:)` search

natecook1000:
Calls to `ranges(of:)` and `firstRange(of:)` with a string parameter actually use two different string searching algorithms. `ranges(of:)` uses the "z-searcher" algorithm, while `firstRange(of:)` uses a two-way search. Since it's better to align on a single path for these searches, the z-searcher has lower requirements, and the two-way search implementation has a correctness bug, this change removes the two-way search algorithm and uses z-search for `firstRange(of:)`.

The correctness bug in `firstRange(of:)` appears only when searching for the second (or later) occurrence of a substring, which you have to be fairly deliberate about. In the example below, the substring at offsets `7..<12` is missed:

```swift
let text = "ADACBADADACBADACB"
//          =====  -----=====
let pattern = "ADACB"
let firstRange = text.firstRange(of: pattern)!
// firstRange ~= 0..<5
let secondRange = text[firstRange.upperBound...].firstRange(of: pattern)!
// secondRange ~= 12..<17
```

This change also removes some unrelated, unused code in Split.swift, in addition to removing an (unused) usage of `TwoWaySearcher`.

rdar://92794248


natecook1000:
@swift-ci Please test

milseman:
How do the two compare in benchmarks? That is, should we consider fixing 2-way or abandon it?

milseman (approved):
We should (separate task) make sure we have a benchmark suite for the collection API.

natecook1000:
I opened #663 to track adding benchmarks — I think we probably need to optimize both algorithms before we can make any judgments about their relative performance. Going to merge for now and then we can bench test these and other algorithms later.

PR #657, opened 2023-04-11
Avoid long expression type checks

natecook1000:
These changes remove several seconds of type-checking time from the RegexBuilder test cases, bringing all expressions under 150ms (on the tested computer).


natecook1000:
@swift-ci Please test

milseman:
Do we want to add comments for these, or otherwise get them into the compiler type checker test suite?

PR #658, opened 2023-04-13
Bug fix and hot path for quantified `.`

milseman:
Apply our `.` fast-path to encoded quantify instructions (it had a separate implementation before).

```
=== Regressions ======================================================================
- IPv6Address                             3.95ms    3.78ms  169µs       4.5%
- BasicCCC                                10.2ms    10.2ms  49.5µs      0.5%
- MACAddress                              2.87ms    2.84ms  29.6µs      1.0%
=== Improvements =====================================================================
- CompilerMessagesAll                     90.6ms    111ms   -20.1ms     -18.2%
- EagarQuantWithTerminalWhole             1.12ms    2.5ms   -1.37ms     -55.0%
- LinesAll                                1.84ms    3.01ms  -1.17ms     -38.8%
- IntersectionCCC                         21.8ms    22.8ms  -1.07ms     -4.7%
- GraphemeBreakNoCapAll                   4.11ms    4.96ms  -841µs      -17.0%
- SubtractionCCC                          21.3ms    21.9ms  -646µs      -2.9%
- symDiffCCC                              48.3ms    48.9ms  -605µs      -1.2%
- InvertedCCC                             20.1ms    20.3ms  -226µs      -1.1%
- EmailBuiltinCharacterClassAll           13ms  13ms    -41.2µs     -0.3%
```

Built on top of https://github.com/apple/swift-experimental-string-processing/pull/655

PR #659, opened 2023-04-13
Run scalar-semantic benchmark variants

milseman:
Refactors benchmark runner a bit to also run scalar semantic versions of benchmarks.

Built on top of https://github.com/apple/swift-experimental-string-processing/pull/658

### DIFF
--- a/Sources/_StringProcessing/Engine/MEQuantify.swift
+++ b/Sources/_StringProcessing/Engine/MEQuantify.swift
@@ -1,31 +1,37 @@
 extension Processor {
   func _doQuantifyMatch(_ payload: QuantifyPayload) -> Input.Index? {
-    var next: Input.Index?
+    // FIXME: is the below updated for scalar semantics?
     switch payload.type {
     case .bitset:
-      next = input.matchBitset(
+      return input.matchBitset(
         registers[payload.bitset], at: currentPosition, limitedBy: end)
     case .asciiChar:
-      next = input.matchScalar(
+      return input.matchScalar(
         UnicodeScalar.init(_value: UInt32(payload.asciiChar)),
         at: currentPosition,
         limitedBy: end,
         boundaryCheck: true)
     case .builtin:
+      // FIXME: bounds check? endIndex or end?
+
       // We only emit .quantify if it consumes a single character
-      next = input._matchBuiltinCC(
+      return input._matchBuiltinCC(
         payload.builtin,
         at: currentPosition,
         isInverted: payload.builtinIsInverted,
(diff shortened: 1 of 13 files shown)


Sources/_StringProcessing/Engine/MEQuantify.swift
```diff
@@ -1,26 +1,37 @@
...
     case .any:
-      let matched = currentPosition != input.endIndex
-        && (!input[currentPosition].isNewline || payload.anyMatchesNewline)
-      next = matched ? input.index(after: currentPosition) : nil
+      // FIXME: endIndex or end?
+      guard currentPosition < input.endIndex else { return nil }
+
+      if payload.anyMatchesNewline {
+        return input.index(after: currentPosition)
+      }
```
natecook1000, line 30: This doesn't look safe — is this method only called in grapheme cluster mode?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -703,3 +687,86 @@ extension Processor {
...
+    // TODO: extremely quick-check-able
+    // TODO: can be sped up with string internals
+
+    assert(end <= endIndex)
+
+    guard pos < end, bitset.matches(char: self[pos]) else {
+      return nil
+    }
+
+    let idx = index(after: pos)
```
natecook1000, line 765: Scalar semantics seem to be implicitly supported here, since we're only going to match in the ASCII domain, but what about an inverted CCC? IIRC `[^abc]` has an ASCII bitset representation, but would match only the first scalar in a multi-scalar character in scalar mode.

Can you add a `FIXME` and I'll circle back to investigate?

Sources/_StringProcessing/Engine/MEQuantify.swift
```diff
@@ -1,26 +1,37 @@
 extension Processor {
   func _doQuantifyMatch(_ payload: QuantifyPayload) -> Input.Index? {
-    var next: Input.Index?
+    // FIXME: is the below updated for scalar semantics?
```
natecook1000, line 3: This method should be okay as long as it's deferring out to methods that handle the semantics correctly — let's make sure those are properly tested.

natecook1000 (commented):
Cool! Good catches on the scalar semantics danger flags — feel free to merge this when you're ready and I'll circle back to the FIXME's.

natecook1000:
Sorry, I missed the PR stack and was just looking at this one first. All three LGTM; I'll come back to the scalar semantics.

natecook1000 (approved):

milseman:
@swift-ci please test

PR #664, opened 2023-04-19
Refactor operations to be on String

milseman:


milseman:
@swift-ci please test

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -730,12 +700,23 @@ extension String {
...
+    if isCaseInsensitive {
+      guard curScalar.properties.lowercaseMapping == scalar.properties.lowercaseMapping
+      else {
+        return nil
+      }
+    } else {
+      guard curScalar == scalar else { return nil }
     }
 
     let idx = unicodeScalars.index(after: pos)
```
natecook1000, line 722: I can't add a comment to the next line, but I think it should be an assertion instead of a test… e.g. `assert(idx <= end)`. If `pos < end` (as checked above), then `unicodeScalars.index(after: pos)` can't be greater than `end` unless something is malformed.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -688,18 +638,29 @@ extension Processor {
...
   ) -> Index? {
     // TODO: This can be greatly sped up with string internals
     // TODO: This is also very much quick-check-able
     assert(end <= endIndex)
 
-    guard pos < end, self[pos] == char else { return nil }
+    guard pos < end else { return nil }
+
+    if isCaseInsensitive {
+      guard self[pos].lowercased() == char.lowercased() else { return nil }
```
natecook1000, line 660: Seems like we should:
(1) store `char` as lowercased so that we don't have to do it every time
(2) think about whether we can reduce the number of times we have to lowercase `self[pos]`

natecook1000 (approved):
Looks good!

Sources/_StringProcessing/Engine/Processor.swift
natecook1000: Oh that's right, `lowercased()` and `properties.lowercaseMapping` both return strings, so this is a bit more complex.

Sources/_StringProcessing/Engine/Processor.swift
milseman: It can if `end` is a sub-scalar index, such as pointing to the second byte of a multi-byte scalar

Sources/_StringProcessing/Engine/Processor.swift
milseman: Agreed, and then we would assert it is that way (and use the argument label to mention this assumption). For 2, do you mean quick checks, or what?

milseman: Also, if the known character to match is case-invariant, we can compile it to a case-sensitive comparison.

PR #667, opened 2023-05-03
[DRAFT] Printable regex

milseman:

PR #670, opened 2023-05-19
Provide literal pattern for convertible regexes

natecook1000:
This adds an (SPI) API for converting a `Regex` instance to a regex literal when possible, whether the regex starts as a literal, a run-time-constructed regex, or using the RegexBuilder syntax. The `_literalPattern` is round-trippable, so a regex created from a `_literalPattern` will have that same `_literalPattern`, but it could be different from the original pattern used to create a regex.

This is an alternative to #667, in that a library using `Regex` as a dependency could create a composed printable regex type with the required semantics.

### DIFF
--- a/Sources/_StringProcessing/LiteralPrinter.swift
+++ b/Sources/_StringProcessing/LiteralPrinter.swift
@@ -17,10 +17,20 @@ extension Regex {
   ///
   /// This is non-`nil` when used on a regex that can be represented as a
   /// string. The literal pattern may be different from the literal or string
-  /// that was used to create the regex. If this regex includes components
-  /// that cannot be represented in a regex literal, such as a capture transform
-  /// or a custom parser that conforms to the `CustomConsumingRegexComponent`
-  /// protocol, this property is `nil`.
+  /// that was used to create the regex, though parsing the `_literalPattern`
+  /// always generates the same internal representation as the original regex.
+  ///
+  ///     // The literal pattern for some regexes is identical to the original:
+  ///     let regex1 = /(\d+):(\d+)/
+  ///     // regex1._literalPattern == #"(\d+):(\d+)"#
+  ///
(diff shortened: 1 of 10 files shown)


Sources/_StringProcessing/LiteralPrinter.swift
```diff
@@ -0,0 +1,527 @@
...
+    switch self {
+    case let .converted(str):
+      return str
+    case .inconvertible:
+      return nil
+    }
+  }
+}
+
+fileprivate struct LiteralPrinter {
```
milseman, line 48: Should this replace `printAsCanonical`, or merge them?

Tests/RegexTests/MatchTests.swift
```diff
@@ -30,7 +44,8 @@ func _firstMatch(
 ) throws -> (String, [String?])? {
   var regex = try Regex(regexStr, syntax: syntax).matchingSemantics(semanticLevel)
   let result = try regex.firstMatch(in: input)
-
+  try _roundTripLiteral(regexStr, syntax: syntax)
```
milseman, line 47: Long term, we'll want to sink this deeper into the library. 

For this, I think we'd also want to test parity for parse only tests.

Sources/_StringProcessing/LiteralPrinter.swift
natecook1000: This doesn't necessarily need to _replace_ `printAsCanonical`, since they work on different things. If we have the AST -> DSL conversion, though, then this should be able to print anything, and we wouldn't need the existing AST printer any more. Not sure if we need the headache of keeping both in sync

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test Windows platform

natecook1000:
@swift-ci Please test

PR #675, opened 2023-05-22
Handle boundaries when matching in substrings

natecook1000:
Some of our existing matching routines use the start/endIndex of the input, which is basically never the right thing to do.

This change revises those checks to use the search bounds, by either moving the boundary check out of the matching method, or if the boundary is a part of what needs to be matched (e.g. word boundaries have different behavior at the start/end than in the middle of a string) the search bounds are passed into the matching method.

Testing is currently handled by piggy-backing on the existing match tests; we should add more tests to handle substring- specific edge cases.

### DIFF
--- a/Sources/_StringProcessing/Engine/MEBuiltins.swift
+++ b/Sources/_StringProcessing/Engine/MEBuiltins.swift
@@ -18,6 +18,7 @@ extension Processor {
     guard currentPosition < end, let next = input.matchBuiltinCC(
       cc,
       at: currentPosition,
+      limitedBy: end,
       isInverted: isInverted,
       isStrictASCII: isStrictASCII,
       isScalarSemantics: isScalarSemantics
@@ -119,35 +120,44 @@ extension Processor {
 
 // MARK: Matching `.`
 extension String {
-  // TODO: Should the below have a `limitedBy` parameter?
-
+  func characterAndEnd(at pos: String.Index, limitedBy end: String.Index) -> (Character, String.Index)? {
+    guard pos < end else { return nil }
+    let next = index(pos, offsetBy: 1, limitedBy: end) ?? end
+    return self[pos..<next].first.map { ($0, next) }
+  }
+  
   func matchAnyNonNewline(
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isScalarSemantics: Bool
   ) -> String.Index? {
-    assert(currentPosition < endIndex)
+    guard currentPosition < end else { return nil }
     if case .definite(let result) = _quickMatchAnyNonNewline(
       at: currentPosition,
+      limitedBy: end,
       isScalarSemantics: isScalarSemantics
     ) {
       assert(result == _thoroughMatchAnyNonNewline(
         at: currentPosition,
+        limitedBy: end,
         isScalarSemantics: isScalarSemantics))
       return result
     }
     return _thoroughMatchAnyNonNewline(
       at: currentPosition,
+      limitedBy: end,
       isScalarSemantics: isScalarSemantics)
   }
 
   @inline(__always)
   private func _quickMatchAnyNonNewline(
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isScalarSemantics: Bool
   ) -> QuickResult<String.Index?> {
-    assert(currentPosition < endIndex)
+    guard currentPosition < end else { return .definite(nil) }
     guard let (asciiValue, next, isCRLF) = _quickASCIICharacter(
-      at: currentPosition
+      at: currentPosition, limitedBy: end
     ) else {
       return .unknown
     }
@@ -163,37 +173,39 @@ extension String {
   @inline(never)
   private func _thoroughMatchAnyNonNewline(
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isScalarSemantics: Bool
   ) -> String.Index? {
-    assert(currentPosition < endIndex)
     if isScalarSemantics {
+      guard currentPosition < end else { return nil }
       let scalar = unicodeScalars[currentPosition]
       guard !scalar.isNewline else { return nil }
       return unicodeScalars.index(after: currentPosition)
     }
 
-    let char = self[currentPosition]
-    guard !char.isNewline else { return nil }
-    return index(after: currentPosition)
+    guard let (char, next) = characterAndEnd(at: currentPosition, limitedBy: end),
+          !char.isNewline
+    else { return nil }
+    return next
   }
 }
 
 // MARK: - Built-in character class matching
 extension String {
-  // TODO: Should the below have a `limitedBy` parameter?
-
   // Mentioned in ProgrammersManual.md, update docs if redesigned
   func matchBuiltinCC(
     _ cc: _CharacterClassModel.Representation,
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isInverted: Bool,
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> String.Index? {
-    assert(currentPosition < endIndex)
+    guard currentPosition < end else { return nil }
     if case .definite(let result) = _quickMatchBuiltinCC(
       cc,
       at: currentPosition,
+      limitedBy: end,
       isInverted: isInverted,
       isStrictASCII: isStrictASCII,
       isScalarSemantics: isScalarSemantics
@@ -201,6 +213,7 @@ extension String {
       assert(result == _thoroughMatchBuiltinCC(
         cc,
         at: currentPosition,
+        limitedBy: end,
         isInverted: isInverted,
         isStrictASCII: isStrictASCII,
         isScalarSemantics: isScalarSemantics))
@@ -209,6 +222,7 @@ extension String {
     return _thoroughMatchBuiltinCC(
       cc,
       at: currentPosition,
+      limitedBy: end,
       isInverted: isInverted,
       isStrictASCII: isStrictASCII,
       isScalarSemantics: isScalarSemantics)
@@ -219,13 +233,17 @@ extension String {
   private func _quickMatchBuiltinCC(
     _ cc: _CharacterClassModel.Representation,
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isInverted: Bool,
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> QuickResult<String.Index?> {
-    assert(currentPosition < endIndex)
+    guard currentPosition < end else { return .definite(nil) }
     guard let (next, result) = _quickMatch(
-      cc, at: currentPosition, isScalarSemantics: isScalarSemantics
+      cc,
+      at: currentPosition,
+      limitedBy: end,
+      isScalarSemantics: isScalarSemantics
     ) else {
       return .unknown
     }
@@ -237,12 +255,16 @@ extension String {
   private func _thoroughMatchBuiltinCC(
     _ cc: _CharacterClassModel.Representation,
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isInverted: Bool,
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> String.Index? {
-    assert(currentPosition < endIndex)
-    let char = self[currentPosition]
+    // TODO: Branch here on scalar semantics
+    // Don't want to pay character cost if unnecessary
+    guard var (char, next) =
+            characterAndEnd(at: currentPosition, limitedBy: end)
+    else { return nil }
     let scalar = unicodeScalars[currentPosition]
 
     let asciiCheck = !isStrictASCII
@@ -250,14 +272,8 @@ extension String {
     || char.isASCII
 
     var matched: Bool
-    var next: String.Index
-    switch (isScalarSemantics, cc) {
-    case (_, .anyGrapheme):
-      next = index(after: currentPosition)
-    case (true, _):
+    if isScalarSemantics && cc != .anyGrapheme {
       next = unicodeScalars.index(after: currentPosition)
-    case (false, _):
-      next = index(after: currentPosition)
     }
 
     switch cc {
@@ -285,7 +301,7 @@ extension String {
       if isScalarSemantics {
         matched = scalar.isNewline && asciiCheck
         if matched && scalar == "\r"
-            && next != endIndex && unicodeScalars[next] == "\n" {
+            && next != end && unicodeScalars[next] == "\n" {
           // Match a full CR-LF sequence even in scalar semantics
           unicodeScalars.formIndex(after: &next)
         }
--- a/Sources/_StringProcessing/Engine/MEBuiltins.swift
+++ b/Sources/_StringProcessing/Engine/MEBuiltins.swift
@@ -190,7 +190,7 @@ extension String {
     limitedBy end: String.Index,
     isScalarSemantics: Bool
   ) -> QuickResult<String.Index?> {
-    guard currentPosition < end else { return .definite(nil) }
+    assert(currentPosition < end)
     guard let (asciiValue, next, isCRLF) = _quickASCIICharacter(
       at: currentPosition, limitedBy: end
     ) else {
@@ -273,7 +273,7 @@ extension String {
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> QuickResult<String.Index?> {
-    guard currentPosition < end else { return .definite(nil) }
+    assert(currentPosition < end)
     guard let (next, result) = _quickMatch(
       cc,
       at: currentPosition,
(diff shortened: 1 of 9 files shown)


natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -15,7 +15,7 @@ extension Processor {
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> Bool {
-    guard let next = input.matchBuiltinCC(
+    guard currentPosition < end, let next = input.matchBuiltinCC(
```
milseman, line 18: Do you think it would make more sense for `matchBuiltinCC` to take a limited by parameter? What should happen if the substring splits a grapheme cluster like `/d` for `1\u{301}` where the substring doesn't have the combining accent?

Sources/_StringProcessing/Engine/MEQuantify.swift
```diff
@@ -27,8 +27,7 @@ extension Processor {
         isStrictASCII: payload.builtinIsStrict,
         isScalarSemantics: isScalarSemantics)
     case .any:
-      // FIXME: endIndex or end?
-      guard currentPosition < input.endIndex else { return nil }
+      guard currentPosition < end else { return nil }
```
milseman, line 30: Were you able to write a test case for this one?

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -194,9 +190,7 @@ extension String {
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> String.Index? {
-    guard currentPosition < endIndex else {
-      return nil
-    }
+    assert(currentPosition < endIndex)
```
milseman, line 193: Out of curiosity, why hoist these out? We'll probably want a naming convention for the assumption in this case, such as `atAssumingInBounds currentPosition`.

Sources/_StringProcessing/Engine/MEBuiltins.swift
natecook1000: You're right — I think having a `limitedBy` parameter is the right behavior, since it will let us address split-character indexes in a way that just having a precondition won't.

Sources/_StringProcessing/Engine/MEBuiltins.swift
natecook1000: As written, the check can't be done in this method, since we're missing the processor's `end`. With the `limitedBy` parameter we can add the guard back in.

natecook1000:
@swift-ci Please test

Sources/_StringProcessing/Unicode/WordBreaking.swift
```diff
@@ -51,11 +53,12 @@ extension Processor {
 extension String {
   func isOnWordBoundary(
     at i: String.Index,
+    in range: Range<String.Index>,
     using cache: inout Set<String.Index>?,
     _ maxIndex: inout String.Index?
   ) -> Bool {
     // TODO: needs benchmark coverage
-    guard i != startIndex, i != endIndex else {
+    guard i != range.lowerBound, i != range.upperBound else {
```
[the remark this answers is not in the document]
natecook1000: That's a programming error, added an assertion for that here.

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -99,8 +102,8 @@ extension String {
     }
 
     var next = utf8.index(after: idx)
-    if next == utf8.endIndex {
-      assert(self[idx].isASCII)
+    if next == end {
+      // assert(self[idx].isASCII)
```
milseman, line 106: What about these assertions?

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -110,10 +113,10 @@ extension String {
     // Handle CR-LF:
     if base == ._carriageReturn && tail == ._lineFeed {
       utf8.formIndex(after: &next)
-      guard next == endIndex || utf8[next]._isSub300StartingByte else {
+      guard next == end || utf8[next]._isSub300StartingByte else {
         return nil
       }
-      assert(self[idx] == "\r\n")
+      // assert(self[idx] == "\r\n")
```
milseman, line 119: What about these assertions?

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -85,11 +85,14 @@ extension String {
...
   func _quickASCIICharacter(
-    at idx: Index
+    at idx: Index,
+    limitedBy end: Index
   ) -> (first: UInt8, next: Index, crLF: Bool)? {
     // TODO: fastUTF8 version
-
-    if idx == endIndex {
+    assert(String.Index(idx, within: unicodeScalars) != nil)
+    assert(idx <= end)
```
milseman, line 93: Somebody in the call graph should be checking this. Where does that happen?

Sources/_StringProcessing/Unicode/ASCII.swift
natecook1000: It's no longer valid to assert here, since a substring can have a mid-character endpoint that makes the result of `self[idx].isASCII` different from `self[..<end][idx].isASCII`. For example, a substring like this:

```swift
let s = "e\u{301}"
let s1 = Substring(s.unicodeScalars.dropLast())
// s.first!.isASCII == false
// s1.first!.isASCII == true
```

Sources/_StringProcessing/Unicode/ASCII.swift
natecook1000: I was thinking this is the same case as above, but is it possible for a following scalar to combine with `"\n"`?

```swift
let s = "\r\n\u{301}"
let s1 = Substring(s.unicodeScalars.dropLast())
// s.count == 2
// s.first == "\r\n"
// s1.first == "\r\n"
```

Do we actually need the `_isSub300StartingByte` check here?

Sources/_StringProcessing/Engine/MEQuantify.swift
milseman: Should we hoist this bounds check to the top of the method, or else leave it as a concern for the String methods? Either way, we want a consistent place and the body of the switch would ideally just be doing dispatch.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -102,56 +103,86 @@ extension Processor {
...
+  ///   with the upper bound of that character. The upper bound is always
+  ///   scalar-aligned.
+  func characterAndEnd(at pos: String.Index, limitedBy end: String.Index) -> (Character, String.Index)? {
+    // FIXME: Sink into the stdlib to avoid multiple boundary calculations
+    guard pos < end else { return nil }
+    let next = index(pos, offsetBy: 1, limitedBy: end) ?? end
+    // Substring will round down non-scalar aligned indices
+    let substr = self[pos..<next]
+    return substr.first.map { ($0, substr.endIndex) }
+  }
```
milseman, line 152: Can you run the benchmark suite over this approach vs just `(self[idx], self.index(after: idx)`?

milseman (commented):
More feedback. Let's try to get this in, especially as we're doing perf work on the quantifications in which bounds checks start to have a measurable impact.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -102,56 +103,96 @@ extension Processor {
...
   }
 
   @inline(__always)
   private func _quickMatchAnyNonNewline(
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isScalarSemantics: Bool
   ) -> QuickResult<String.Index?> {
-    assert(currentPosition < endIndex)
+    guard currentPosition < end else { return .definite(nil) }
```
milseman, line 193: This `guard` isn't strictly necessary because the only caller has it. We can come up with an argument label or naming convention or even parameter comment to codify it. Definitely doesn't have to be part of this PR though.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -225,13 +268,17 @@ extension String {
   private func _quickMatchBuiltinCC(
     _ cc: _CharacterClassModel.Representation,
     at currentPosition: String.Index,
+    limitedBy end: String.Index,
     isInverted: Bool,
     isStrictASCII: Bool,
     isScalarSemantics: Bool
   ) -> QuickResult<String.Index?> {
-    assert(currentPosition < endIndex)
+    guard currentPosition < end else { return .definite(nil) }
```
milseman, line 276: (similarly)

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -291,7 +336,7 @@ extension String {
       if isScalarSemantics {
         matched = scalar.isNewline && asciiCheck
         if matched && scalar == "\r"
-            && next != endIndex && unicodeScalars[next] == "\n" {
+            && next != end && unicodeScalars[next] == "\n" {
```
milseman, line 339: BTW, is it `!=` or `<`?

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -536,9 +537,10 @@ extension Processor {
 
     case .consumeBy:
       let reg = payload.consumer
+      let consumer = registers[reg]
       guard currentPosition < searchBounds.upperBound,
-            let nextIndex = registers[reg](
-              input, currentPosition..<searchBounds.upperBound)
+            let nextIndex = consumer(input, currentPosition..<searchBounds.upperBound),
+            nextIndex <= end
```
milseman, line 543: I'm ok with being extra cautious, especially in the context of 3rd party consumer code. Technically this might mean that they are breaking the contract by not doing their own bounds checking, but that's such an onerous requirement that I think this is better.

Sources/_StringProcessing/Engine/MEBuiltins.swift
natecook1000: Should probably be `<` here

natecook1000:
Benchmarks show mostly mild churn with some increases, as we might expect from additional bounds checking. Something seems off with the `EmojiRegex_All` benchmark, however, which is showing a nearly 100% increase in execution time.

```
- EmojiRegex_All                          126ms	63.9ms	62.2ms		97.3%
- EmailRFCNoMatches_All                   166ms	140ms	26.6ms		19.0%
- symDiffCCC_All_Scalar                   58.3ms	47.5ms	10.8ms		22.7%
- symDiffCCC_All                          58.4ms	47.7ms	10.8ms		22.6%
- EmailRFC_All                            70.6ms	65.6ms	4.99ms		7.6%
- IntersectionCCC_All_Scalar              22.3ms	20.6ms	1.68ms		8.1%
- IntersectionCCC_All                     22.1ms	20.6ms	1.48ms		7.2%
- EmailLookaheadNoMatches_All             40.6ms	39.3ms	1.32ms		3.4%
- BasicRangeCCC_All                       10.7ms	9.66ms	1.08ms		11.2%
- BasicRangeCCC_All_Scalar                10.7ms	9.69ms	1.03ms		10.7%
- CaseInsensitiveCCC_All                  11.4ms	10.4ms	1.01ms		9.7%
- BasicCCC_All                            10.2ms	9.22ms	1e+03µs		10.8%
- CaseInsensitiveCCC_All_Scalar           11.5ms	10.5ms	997µs		9.5%
- BasicCCC_All_Scalar                     10.1ms	9.15ms	983µs		10.7%
- DiceRollsInText_All_Scalar              43.4ms	42.6ms	783µs		1.8%
- SubtractionCCC_All_Scalar               20.9ms	20.3ms	667µs		3.3%
- SubtractionCCC_All                      20.9ms	20.3ms	600µs		3.0%
- EmailLookahead_All                      41.5ms	41.1ms	322µs		0.8%
- EmailLookaheadNoMatches_All_Scalar      24.9ms	24.7ms	194µs		0.8%
- EmailBuiltinCharacterClass_All          10.9ms	10.8ms	150µs		1.4%
- NotFound_All                            5.97ms	5.82ms	148µs		2.5%
- BasicBuiltinCharacterClass_All_Scalar   6.29ms	6.19ms	102µs		1.6%
- IPv4Address                             2.62ms	2.52ms	93.7µs		3.7%
- IPv4Address_Scalar                      2.3ms	2.22ms	79.3µs		3.6%
- LiteralSearch_All_Scalar                4.5ms	4.43ms	70.9µs		1.6%
- DiceNotation                            5.21ms	5.15ms	58.4µs		1.1%
- EmailLookaheadList_Scalar               5.18ms	5.13ms	49.6µs		1.0%
- HangulSyllable_First                    2.72ms	2.69ms	30.7µs		1.1%
- HangulSyllable_First_Scalar             2.27ms	2.24ms	27.3µs		1.2%
- CompilerMessages_All                    93.6ms	99.6ms	-6.01ms		-6.0%
- CompilerMessages_All_Scalar             76.4ms	81.9ms	-5.47ms		-6.7%
- EmailRFCNoMatches_All_Scalar            125ms	127ms	-2.51ms		-2.0%
- EagarQuantWithTerminal_Whole_Scalar     835µs	1.4ms	-564µs		-40.3%
- EagarQuantWithTerminal_Whole            837µs	1.39ms	-552µs		-39.7%
- EmailRFC_All_Scalar                     47.1ms	47.5ms	-405µs		-0.9%
- GraphemeBreakNoCap_All_Scalar           3.51ms	3.83ms	-326µs		-8.5%
- ReluctantQuantWithTerminal_Whole_Scalar 5.33ms	5.6ms	-263µs		-4.7%
- GraphemeBreakNoCap_All                  4.22ms	4.39ms	-167µs		-3.8%
- LiteralSearchNotFound_All               5.15ms	5.29ms	-137µs		-2.6%
- AnchoredNotFound_First                  9.3ms	9.41ms	-118µs		-1.2%
- HangulSyllable_All_Scalar               5.15ms	5.26ms	-114µs		-2.2%
- Words_All                               14.2ms	14.3ms	-95.9µs		-0.7%
- MACAddress_Scalar                       2.63ms	2.67ms	-46.7µs		-1.7%
- Lines_All                               1.89ms	1.93ms	-38.8µs		-2.0%
```

Sources/_StringProcessing/Engine/MEBuiltins.swift
milseman: Post-merge performance tweak. We should hand-outline the rare and slow path, no need to pollute the i-cache and confuse the optimizer.

Sources/_StringProcessing/Engine/Processor.swift
```diff
@@ -720,7 +716,7 @@ extension String {
     }
 
     let idx = unicodeScalars.index(after: pos)
-    guard idx <= end else { return nil }
+    assert(idx <= end, "Input is a substring with a sub-scalar endIndex.")
```
milseman, line 719: Can we also put the assertion near the top of Processor proper? We'll want to rely on it throughout the engine

Sources/_StringProcessing/Unicode/ASCII.swift
```diff
@@ -85,11 +85,14 @@ extension String {
...
+    
+    if idx == end {
```
milseman, line 95: Post-merge cleanup: it seems like anything that's not permissive of an empty match should fail long before reaching this code

Sources/_StringProcessing/Unicode/WordBreaking.swift
```diff
@@ -37,7 +37,9 @@ extension Processor {
     if currentPosition == subjectBounds.lowerBound {
       return matchesWord(at: currentPosition)
     }
-    let priorIdx = input.index(before: currentPosition)
+    let priorIdx = semanticLevel == .graphemeCluster
+      ? input.index(before: currentPosition)
+      : input.unicodeScalars.index(before: currentPosition)
```
milseman, line 42: Do we have any new test cases that hit this?

Tests/RegexTests/MatchTests.swift
```diff
@@ -30,6 +30,68 @@ func _firstMatch(
 ) throws -> (String, [String?])? {
   var regex = try Regex(regexStr, syntax: syntax).matchingSemantics(semanticLevel)
   let result = try regex.firstMatch(in: input)
+  
+  func validateSubstring(_ substringInput: Substring) throws {
```
milseman, line 34: Post-PR testing: let's make a test input that contains many kinds of complex grapheme clusters and, for each test regex and each substring of that input, any match of that substring is equal to the match of a `String` copy of that substring.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -291,7 +336,7 @@ extension String {
       if isScalarSemantics {
         matched = scalar.isNewline && asciiCheck
         if matched && scalar == "\r"
-            && next != endIndex && unicodeScalars[next] == "\n" {
+            && next < end && unicodeScalars[next] == "\n" {
```
milseman, line 339: E.g. here is a place where we implicitly rely on `end` being scalar aligned (which is a good assumption to make and would be good to assert at the very top of a match operation).

milseman (approved):
LGTM, let's get this in. This is a very important correctness fix and we can get the perf back after merge.

milseman:
@swift-ci please test

PR #678, opened 2023-06-06
Provide literal pattern for convertible regexes

natecook1000:
This is to support building a toolchain with the change in #670.


natecook1000:
@swift-ci Please test

PR #682, opened 2023-07-19
Optimize search for start-anchored regexes

natecook1000:
When a regex is anchored to the start of a subject, there's no need to search throughout a string for the pattern when searching for the first match: a prefix match is sufficient.

This adds a regex compilation-time check about whether a match can only be found at the start of a subject, and then uses that to choose whether to defer to `prefixMatch` from within `firstMatch`.

rdar://111392228

### DIFF
--- a/Sources/_StringProcessing/Regex/DSLTree.swift
+++ b/Sources/_StringProcessing/Regex/DSLTree.swift
@@ -711,6 +711,105 @@ extension DSLTree.Node {
   }
 }
 
+extension DSLTree.Node {
+  /// Implementation for `canOnlyMatchAtStart`, which maintains the option
+  /// state.
+  ///
+  /// For a given specific node, this method can return one of three values:
+  ///
+  /// - `true`: This node is guaranteed to match only at the start of a subject.
+  /// - `false`: This node can match anywhere in the subject.
+  /// - `nil`: This node is inconclusive about where it can match.
+  ///
+  /// In particular, non-required groups and option-setting groups are
+  /// inconclusive about where they can match.
+  func _canOnlyMatchAtStartImpl(_ options: inout MatchingOptions) -> Bool? {
+    switch self {
+    // Defining cases
+    case .atom(.assertion(.startOfSubject)):
+      return true
+    case .atom(.assertion(.caretAnchor)):
+      return !options.anchorsMatchNewlines
+      
+    // Changing options doesn't determine `true`/`false`.
+    case .atom(.changeMatchingOptions(let sequence)):
+      options.apply(sequence.ast)
+      return nil
+      
+    // Any other atom or consuming node returns `false`.
+    case .atom, .customCharacterClass, .quotedLiteral:
+      return false
+      
+    // Trivia/empty have no effect.
+    case .trivia, .empty:
+      return nil
+      
+    // In an alternation, all of its children must match only at start.
+    case .orderedChoice(let children):
+      return children.allSatisfy { $0._canOnlyMatchAtStartImpl(&options) == true }
+      
+    // In a concatenation, the first definitive child provides the answer.
+    case .concatenation(let children):
+      for child in children {
+        if let result = child._canOnlyMatchAtStartImpl(&options) {
+          return result
+        }
+      }
+      return false
+
+    // Groups (and other parent nodes) defer to the child.
+    case .nonCapturingGroup(let kind, let child):
+      options.beginScope()
+      defer { options.endScope() }
+      if case .changeMatchingOptions(let sequence) = kind.ast {
+        options.apply(sequence)
+      }
+      return child._canOnlyMatchAtStartImpl(&options)
+    case .capture(_, _, let child, _):
+      options.beginScope()
+      defer { options.endScope() }
+      return child._canOnlyMatchAtStartImpl(&options)
+    case .ignoreCapturesInTypedOutput(let child),
+        .convertedRegexLiteral(let child, _):
+      return child._canOnlyMatchAtStartImpl(&options)
+
+    // A quantification that doesn't require its child to exist can still
+    // allow a start-only match. (e.g. `/(foo)?^bar/`)
+    case .quantification(let amount, _, let child):
+      return amount.requiresAtLeastOne
+        ? child._canOnlyMatchAtStartImpl(&options)
+        : nil
+
+    // For conditional nodes, both sides must require matching at start.
+    case .conditional(_, let child1, let child2):
+      return child1._canOnlyMatchAtStartImpl(&options) == true
+        && child2._canOnlyMatchAtStartImpl(&options) == true
+
+    // Extended behavior isn't known, so we return `false` for safety.
+    case .consumer, .matcher, .characterPredicate, .absentFunction:
+      return false
+    }
+  }
+  
+  /// Returns a Boolean value indicating whether the regex with this node as
+  /// the root can _only_ match at the start of a subject.
+  ///
+  /// For example, these regexes can only match at the start of a subject:
+  ///
+  /// - `/^foo/`
+  /// - `/(^foo|^bar)/` (both sides of the alternation start with `^`)
+  ///
+  /// These can match other places in a subject:
+  ///
+  /// - `/(^foo)?bar/` (`^` is in an optional group)
+  /// - `/(^foo|bar)/` (only one side of the alternation starts with `^`)
+  /// - `/(?m)^foo/` (`^` means "the start of a line" due to `(?m)`)
+  func canOnlyMatchAtStart() -> Bool {
+    var options = MatchingOptions()
+    return _canOnlyMatchAtStartImpl(&options) ?? false
+  }
+}
+
 // MARK: AST wrapper types
 //
 // These wrapper types are required because even @_spi-marked public APIs can't
@@ -817,6 +916,17 @@ extension DSLTree {
       }
       public static func range(_ lower: Int, _ upper: Int) -> Self {
         .init(ast: .range(.init(lower, at: .fake), .init(upper, at: .fake)))
+      }
+      
+      internal var requiresAtLeastOne: Bool {
+        switch ast {
+        case .zeroOrOne, .zeroOrMore, .upToN:
+          return false
+        case .oneOrMore:
+          return true
+        case .exactly(let num), .nOrMore(let num), .range(let num, _):
+          return num.value.map { $0 > 0 } ?? false
+        }
       }
     }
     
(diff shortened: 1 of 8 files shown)


Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -711,6 +711,105 @@ extension DSLTree.Node {
...
+    // Defining cases
+    case .atom(.assertion(.startOfSubject)):
+      return true
+    case .atom(.assertion(.caretAnchor)):
+      return !options.anchorsMatchNewlines
+      
+    // Changing options doesn't determine `true`/`false`.
+    case .atom(.changeMatchingOptions(let sequence)):
+      options.apply(sequence.ast)
+      return nil
```
milseman, line 737: Why are we mutating options in a query API?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -711,6 +711,105 @@ extension DSLTree.Node {
...
+      
+    // In an alternation, all of its children must match only at start.
+    case .orderedChoice(let children):
+      return children.allSatisfy { $0._canOnlyMatchAtStartImpl(&options) == true }
+      
+    // In a concatenation, the first definitive child provides the answer.
+    case .concatenation(let children):
+      for child in children {
+        if let result = child._canOnlyMatchAtStartImpl(&options) {
```
milseman, line 754: Instead of a loop, would this be something like `if let result = children.first?._canOnly...`

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -711,6 +711,105 @@ extension DSLTree.Node {
...
+    case .ignoreCapturesInTypedOutput(let child),
+        .convertedRegexLiteral(let child, _):
+      return child._canOnlyMatchAtStartImpl(&options)
+
+    // A quantification that doesn't require its child to exist can still
+    // allow a start-only match. (e.g. `/(foo)?^bar/`)
+    case .quantification(let amount, _, let child):
+      return amount.requiresAtLeastOne
+        ? child._canOnlyMatchAtStartImpl(&options)
+        : nil
```
milseman, line 781: What's the logic for this? If a child is quantified `2..<5` times, then how would it be a start-only match? What should the behavior be for the example in the comments?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -711,6 +711,105 @@ extension DSLTree.Node {
...
+
+    // For conditional nodes, both sides must require matching at start.
+    case .conditional(_, let child1, let child2):
+      return child1._canOnlyMatchAtStartImpl(&options) == true
+        && child2._canOnlyMatchAtStartImpl(&options) == true
+
+    // Extended behavior isn't known, so we return `false` for safety.
+    case .consumer, .matcher, .characterPredicate, .absentFunction:
+      return false
```
milseman, line 790: What's the difference between returning false and nil in practice?

Sources/_StringProcessing/Regex/DSLTree.swift
```diff
@@ -711,6 +711,105 @@ extension DSLTree.Node {
...
+  ///
+  /// For a given specific node, this method can return one of three values:
+  ///
+  /// - `true`: This node is guaranteed to match only at the start of a subject.
+  /// - `false`: This node can match anywhere in the subject.
+  /// - `nil`: This node is inconclusive about where it can match.
+  ///
+  /// In particular, non-required groups and option-setting groups are
+  /// inconclusive about where they can match.
+  func _canOnlyMatchAtStartImpl(_ options: inout MatchingOptions) -> Bool? {
```
milseman, line 726: Would this be `private`?

milseman (approved):
Overall LGTM. Review has some minor code comments. Do you have before/after benchmark results? Are you planning on more substring testing?

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: `internal`. I'll make it explicit.

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: If a child is quantified at least once, then whether it can match only at the start depends on the child. This isn't making a determination about whether it's possible for it to match at all, just that if it _can_ match, it only matches at the start. The key part is that a quantification that can be skipped doesn't actually impact whether or not the pattern can only match at the start, since we can skip an anchor inside the quanitification or jump to one after it:

- `a?^b`: can only match at the start — `"b"` matches, `"aaab"` doesn't
- `a?b`: can match anywhere
- `(^a)?b`: can match anywhere
- `a{2,4}b`: can match anywhere
- `a{2,4}^b`: can match anywhere (won't actually match anywhere, but that's a different optimization)
- `(^a){2,4}b`: can only match at the start (won't actually match, but that's a different optimization)

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: Returning `false` stops searching the tree and returns `false` all the way up the recursion stack. Returning `nil` allows searching to continue (which really only happens among concatenation siblings, see other comment).

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: This is how the options type is built — we can't really use the traversal stack because this kind of options change affects siblings, not just descendants.

Sources/_StringProcessing/Regex/DSLTree.swift
natecook1000: Unfortunately, no — a regex like `/(?m)^foo/` is a concatenation of `(?m)`, `^`, and `foo`, so we have to be able to look past the first child to find something that requires or rules out matching only at the start.

natecook1000:
@swift-ci Please test

natecook1000:
These are the benchmark results I see — I had to modify the `AnchorNotFound` benchmark to include a `firstMatch` search, and it shows a big improvement. AFAICT the rest are noise?

```
- CompilerMessages_All                    99.1ms	97.4ms	1.66ms		1.7%
- EmojiRegex_All                          64.1ms	63.1ms	932µs		1.5%
- EmojiRegex_All_Scalar                   41.4ms	40.5ms	923µs		2.3%
- EmailRFC_All_Scalar                     46.9ms	46.1ms	746µs		1.6%
- EmailLookaheadNoMatches_All             39ms	38.4ms	539µs		1.4%
- InvertedCCC_All_Scalar                  22.2ms	21.7ms	426µs		2.0%
- ReluctantQuant_Whole                    9.57ms	9.16ms	406µs		4.4%
- EmailLookahead_All                      40.5ms	40.1ms	370µs		0.9%
- SubtractionCCC_All                      20.2ms	19.8ms	347µs		1.7%
- IntersectionCCC_All                     20.5ms	20.2ms	318µs		1.6%
- LiteralSearch_All_Scalar                4.51ms	4.34ms	172µs		4.0%
- Css_All_Scalar                          3.01ms	2.87ms	138µs		4.8%
- AnchoredNotFound_All_Scalar             8.63ms	8.5ms	132µs		1.6%
- Numbers_All                             6.12ms	6ms	125µs		2.1%
- ReluctantQuantWithTerminal_Whole_Scalar 5.45ms	5.37ms	82.2µs		1.5%
- ReluctantQuantWithTerminal_Whole        5.44ms	5.38ms	55.3µs		1.0%
- IPv4Address_Scalar                      2.32ms	2.27ms	50.3µs		2.2%
- AnchoredNotFound_First                  9.13ms	13.3ms	-4.16ms		-31.3%
- AnchoredNotFound_First_Scalar           5.62ms	8.63ms	-3.01ms		-34.9%
- EmailRFCNoMatches_All_Scalar            126ms	129ms	-2.11ms		-1.6%
- DiceRollsInText_All_Scalar              41.8ms	42.4ms	-645µs		-1.5%
- GraphemeBreakNoCap_All_Scalar           3.54ms	3.89ms	-346µs		-8.9%
- BasicRangeCCC_All_Scalar                9.5ms	9.84ms	-339µs		-3.4%
- BasicCCC_All_Scalar                     9.03ms	9.37ms	-335µs		-3.6%
- InvertedCCC_All                         21.5ms	21.7ms	-263µs		-1.2%
- EmailLookaheadList_Scalar               5.02ms	5.24ms	-226µs		-4.3%
- LiteralSearchNotFound_All_Scalar        4.16ms	4.38ms	-221µs		-5.0%
- GraphemeBreakNoCap_All                  4.25ms	4.45ms	-208µs		-4.7%
- EmailBuiltinCharacterClass_All_Scalar   10.5ms	10.7ms	-199µs		-1.9%
- BasicRangeCCC_All                       9.48ms	9.67ms	-195µs		-2.0%
- CaseInsensitiveCCC_All                  10.2ms	10.4ms	-185µs		-1.8%
- IPv6Address_Scalar                      2.76ms	2.93ms	-170µs		-5.8%
- BasicCCC_All                            8.99ms	9.15ms	-160µs		-1.8%
- LiteralSearchNotFound_All               5.1ms	5.26ms	-156µs		-3.0%
- IPv4Address                             2.53ms	2.66ms	-133µs		-5.0%
- Numbers_All_Scalar                      5.28ms	5.41ms	-122µs		-2.3%
- CaseInsensitiveCCC_All_Scalar           10.2ms	10.3ms	-119µs		-1.1%
- HangulSyllable_All_Scalar               5.09ms	5.21ms	-119µs		-2.3%
- LiteralSearch_All                       5.27ms	5.39ms	-115µs		-2.1%
- Lines_All_Scalar                        1.88ms	1.99ms	-107µs		-5.4%
- HangulSyllable_All                      6.01ms	6.09ms	-79.4µs		-1.3%
- Lines_All                               1.9ms	1.92ms	-18.9µs		-1.0%
```

Sources/_StringProcessing/Regex/DSLTree.swift
milseman: Ah, so that's the recursion base-case for a member of a concatenation which will affect following siblings. Could you mention that in a comment?

Sources/_StringProcessing/Regex/DSLTree.swift
milseman: Awesome, thanks for clarifying! Could you update the comments for why `nil` is used in that case?

PR #685, opened 2023-07-20
Fix misuse of `XCTSkip()`

grynspan:
In a few tests, `XCTSkip()` is used as a function call, but it's actually a type conforming to the `Error` protocol and is meant to be thrown. (This is an understandable mistake, since in Objective-C it is a macro that ultimately throws an exception.)


grynspan:
Looks like I messed up the branches.

grynspan:
@swift-ci please test

PR #691, opened 2023-08-31
[nfc] Avoid pointer conversions

glessard:
Preparing for a compiler that's less permissive about pointer conversions.
1. Use the stdlib's pattern for lazy atomic references, by explicitly declaring the pointer as a computed property.
2. Use an explicit pointer when a C function isn't detected as such by the compiler.

### DIFF
--- a/Sources/_StringProcessing/Regex/Core.swift
+++ b/Sources/_StringProcessing/Regex/Core.swift
@@ -126,6 +126,13 @@ extension Regex {
   /// A program representation that caches any lowered representation for
   /// execution.
   internal final class Program {
+
+    // This stored property should be stored at offset zero.  We perform atomic
+    // operations on it.
+    //
+    /// Do not access this property directly - all accesses must go through `loweredProgram`.
+    fileprivate var _loweredProgramStorage: AnyObject? = nil
+
     /// The underlying IR.
     ///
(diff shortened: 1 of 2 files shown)


glessard:
@swift-ci please test

natecook1000:
LGTM - how did you discover these issues? New compiler, or certain flags?

glessard:
This was using a compiler branch that implements (mostly) my restricted pointer conversions pitch.

glessard:

milseman (approved):

Sources/_StringProcessing/Regex/Core.swift
```diff
@@ -126,6 +126,13 @@ extension Regex {
   /// A program representation that caches any lowered representation for
   /// execution.
   internal final class Program {
+
+    // This stored property should be stored at offset zero.  We perform atomic
+    // operations on it.
+    //
+    /// Do not access this property directly - all accesses must go through `loweredProgram`.
```
glessard, line 133:
```suggestion
    /// Do not access this property directly - all accesses must go through `_loweredProgramStoragePtr `.
```

natecook1000:
@swift-ci Please test

natecook1000 (approved):
🚢

PR #698, opened 2023-11-02
Adds SPI for a NSRE compatibility mode option

natecook1000:
NSRegularExpression matches at the Unicode scalar level, but also matches `\r\n` sequences with a single `.` when single-line mode is enabled. This adds a `_nsreCompatibility` property that enables both of those behaviors, and implements support for the special case handling of `.`.


natecook1000:
@swift-ci Please test

PR #699, opened 2023-11-15
[Windows] Workaround for making it work in Windows

rintaro:
* using generics in Error type seems to cause crashes
* CustomStirngConvertible is not visible


rintaro:
Found a fix. this is not needed.

PR #708, opened 2023-12-12
[DRAFT] Refactor and generalize quant nonmutating generalization

milseman:


milseman:
My attempt to make the refactoring a little faster and more general. Note that trying to simplify `runQuantify` in a way that removes static knowledge of zero-or-more results in further 50%-ish regressions, which is pretty nuts.

PR #709, opened 2023-12-13
[swift/main] Provide literal pattern for convertible regexes

natecook1000:
This adds an (SPI) API for converting a `Regex` instance to a regex literal when possible, whether the regex starts as a literal, a run-time-constructed regex, or using the RegexBuilder syntax. The `_literalPattern` is round-trippable, so a regex created from a `_literalPattern` will have that same `_literalPattern`, but it could be different from the original pattern used to create a regex.

This adds another confirmation test, as part of the parse testing. When a parsing test is expected to succeed, and contains supported syntax, this checks that a round-tripped literal pattern maintains its identity.

Cherry pick of #670.


natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test

natecook1000:
@swift-ci Please test Windows platform

PR #712, opened 2024-01-17
Add @retroactive to conformances that need it

tshortli:
SE-0364 requires that retroactive conformance be marked with `@retroactive` to suppress warnings about the dangers of conforming types you don't own to protocols you don't own. The `RegexBuilder` module declares conformances to `RegexComponent` (from the `Regex` module) on several types from the standard library, like `String`. Since `RegexBuilder` declares neither the protocol nor the conforming type, these conformances are technically retroactive even though the modules involved are all developed and distributed together and can be trusted to organize the conformances this way safely.

The compiler warnings about these conformances are quite spammy; they get emitted every time someone builds the Swift compiler, so I'd like to suppress them.

### DIFF
--- a/Sources/RegexBuilder/DSL.swift
+++ b/Sources/RegexBuilder/DSL.swift
@@ -32,21 +32,21 @@ internal protocol _BuiltinRegexComponent: RegexComponent {
 // MARK: - Primitive regex components
 
 @available(SwiftStdlib 5.7, *)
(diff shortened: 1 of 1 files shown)


tshortli:
@swift-ci please test

glessard (approved):
lgtm

natecook1000 (approved):

PR #719, opened 2024-02-13
Fix release mode testing issue

natecook1000:
The call to disable optimizations during testing was in an assert call, so it was elided in release builds, even though the code that expected optimizations to be disabled was still used. This created spurious failures during release mode testing, which appears to have recently been enabled in CI.


natecook1000:
@swift-ci Please test

PR #727, opened 2024-03-04
Suppress compiler warnings about Errors with non-Sendable members

tshortli:
A type conforming to `Error` must be `Sendable`, but `DelimiterLexError` and `CompilerLexError` carry source locations represented by `UnsafeRawPointer`, which is inherently non-`Sendable`. The compiler diagnoses the failure to meet `Sendable` requirements on these error types, which will prevent Swift 6 adoption in `_RegexParser` and also generates build log spam every time the compiler and standard library are built. There are a couple ways to resolve this problem:
    
1. Refactor so that the errors do not need to carry source locations.
2. Refactor to avoid using `Error` (the code using these errors would not require `Sendable` if it were not for `Error`'s inherent `Sendable` requirement).
3. Suppress the warnings with an `@unchecked Sendable` wrapper.
    
Option 1 probably requires significant refactoring and option 2 seems undesirable since it means that the implementation must forgo the convenience of Swift's error handling model. Therefore this PR implements option 3 as the most expedient way to stop the diagnostic spam, leaving improvement on the status quo as a future exercise.

Also includes a drive-by fix for a trivial, unrelated warning.

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/CompilerInterface.swift
+++ b/Sources/_RegexParser/Regex/Parse/CompilerInterface.swift
@@ -17,9 +17,19 @@ public let currentRegexLiteralFormatVersion = 1
 
 @_spi(CompilerInterface)
 public struct CompilerLexError: Error {
+  var underlyingLocation: UnsafeSourceLocation
+
   public var message: String
-  public var location: UnsafeRawPointer
+  public var location: UnsafeRawPointer { return underlyingLocation.ptr }
   public var completelyErroneous: Bool
+
+  init(
+    message: String, location: UnsafeRawPointer, completelyErroneous: Bool
+  ) {
+    self.message = message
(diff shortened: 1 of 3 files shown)


tshortli:
@swift-ci please test

tshortli:
@swift-ci please test

hamishknight (approved):
Thanks!

tshortli:
@swift-ci please test Windows

natecook1000:
Unless there are exceptional circumstances, please make changes on `main` before/at the same time as landing them on `swift/main`. Thanks!

Updating `main` in #729.

tshortli:
> Unless there are exceptional circumstances, please make changes on main before/at the same time as landing them on swift/main. Thanks!

Sorry, I think I'm just very confused about how this is supposed to work in this repo. It looked to me like `main` was very out of date and therefore was abandoned or something but I must have been looking at the wrong thing.

PR #731, opened 2024-03-14
Resolve `switch covers known cases, but 'Enum' may have additional unknown values` warnings

tshortli:
When building with CMake during the Swift compiler build, the libraries in this package are built with library evolution enabled. That causes the compiler to emit diagnostics that encourage use of `@unknown default` to handle unknown cases that might be added in the future. Silence these warnings by adding `@unknown default` cases where they are missing. Guard these with the `RESILIENT_LIBRARIES` compilation condition to avoid triggering a `Default will never be executed` diagnostic during the package build.

Note that this would not be required if `_RegexParser` could always build without library evolution, but that change had to be reverted (rdar://124541877).


tshortli:
https://github.com/apple/swift/pull/72333

@swift-ci please test

tshortli:
In order to get this in before the branch I'm going to ask for forgiveness instead of permission and land it now. If anyone has feedback let me know, I'm happy to follow up.

PR #660, opened 2023-04-14
Remove most consumer functions

natecook1000:
This is based _heavily_ off the work in #590, rebased onto main, with some changes to remove even more consumer uses. Consumer functions only have two remaining uses: non-ASCII ranges and Unicode lookups (for things like general category, binary properties, name, etc.).

This change primarily treats custom character classes as alternations around their contents, with set operations emitted as instructions instead of implemented via consumer function.

PR #733, opened 2024-03-27
[swift/main] Remove most consumer functions

natecook1000:
Cherry-picked from #660.

This is based heavily off the work in #590, rebased onto main, with some changes to remove even more consumer uses. Consumer functions only have two remaining uses: non-ASCII ranges and Unicode lookups (for things like general category, binary properties, name, etc.).

This change primarily treats custom character classes as alternations around their contents, with set operations emitted as instructions instead of implemented via consumer function.


natecook1000:
@swift-ci Please test

PR #734, opened 2024-03-29
[6.0] Remove most consumer functions

natecook1000:
**Explanation:** This change removes most uses of consumer functions in the regex processor, by implementing them as instructions instead.
**Scope:** All the changes are internal implementation within the `_StringProcessing` module.
**Issue:** n/a
**Original PR:** #660
**Risk:** Low. There's no API or ABI impact of this change.
**Testing:** This change adds compilation tests that verify that consumer functions are removed when appropriate, and uses the existing unit tests to verify no change in matching behavior.
**Reviewer:** @stephentyrone

---

_Original PR text:_
This is based heavily off the work in #590, rebased onto main, with some changes to remove even more consumer uses. Consumer functions only have two remaining uses: non-ASCII ranges and Unicode lookups (for things like general category, binary properties, name, etc.).

This change primarily treats custom character classes as alternations around their contents, with set operations emitted as instructions instead of implemented via consumer function.

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen.swift
+++ b/Sources/_StringProcessing/ByteCodeGen.swift
@@ -256,9 +256,11 @@ fileprivate extension Compiler.ByteCodeGen {
     }
   }
 
-  mutating func emitAlternation(
-    _ children: [DSLTree.Node]
-  ) throws {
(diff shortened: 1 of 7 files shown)


stephentyrone (approved):

natecook1000:
@swift-ci Please test

PR #737, opened 2024-04-23
Work around compiler type checker issue

natecook1000:
This refactors a test and disables the `RegexBenchmark` target to work around a type checker issue in the April 21 development toolchain, used by CI. The test is still running properly with these changes (moved some function-scoped symbols out to the global scope, marked `fileprivate`).

Related: rdar://126877024


natecook1000:
@swift-ci Please test

PR #738, opened 2024-05-16
Support change matching options in Regex refactoring

Azoy:
This adds support for tokens like `(?s)` and `(?-s)` in converting a regex literal to the pattern DSL. We need to make a new `Regex` scope to apply those options to.

Also, print out `CharacterClass.anyOf` in some situations to avoid ambiguity in the type checker.

Resolves: rdar://124965493 and rdar://119988624


Azoy:
@swift-ci please test

Azoy:
@swift-ci please test

PR #744, opened 2024-06-09
Fix word boundary matching in `matches(of:)`

natecook1000:
A prior change incorrectly treated the start of the `searchBounds` as an always-valid word boundary, which resulted in a word boundary at the start of a pattern unconditionally matching at the restart point when finding all matches in a string.

This change corrects the usage to recognize word boundaries only at the start of the `subjectBounds`.

rdar://129417643


natecook1000:
@swift-ci Please test

natecook1000:

milseman (approved):
LGTM

PR #753, opened 2024-07-17
Delete CODE_OF_CONDUCT.md

parispittman:
after this repo migrates to /swiftlang today, it will be opt-ing in to the org wide code of conduct that is present at the root .github config folder. Having this file present means that it is opt-ing out of the org wide and has its own process which will not be true.


stephentyrone (approved):

shahmishal:
@swift-ci test

PR #755, opened 2024-07-24
Migrate to internal imports

tshortli:
As of the Swift 6 compiler, `@_implementationOnly import` is deprecated in favor of `internal import` and as a result the use of `@_implementationOnly import` in this project is generating a lot of diagnostic noise when building the Swift standard library.

For Swift libraries with library evolution, `@_implementationOnly import` and `internal import` are roughly functionally equivalent, aside from improved diagnostics for `internal import`. For non-resilient libraries, the main difference is that `internal import` does not actually hide a module dependency from downstream clients because the layout of a type in a non-resilient library may depend on types coming from an `internal import` (with `@_implementationOnly import` the same situation would result in a silent mis-compile, which is the reason that `@_implementationOnly import` is deprecated). The `_RegexParser` module dependency does not need to be hidden from clients since it is installed in standard locations in the SDK/toolchain. Therefore this migration should be safe, regardless of library resilience mode.

### DIFF
--- a/Sources/RegexBenchmark/Benchmark.swift
+++ b/Sources/RegexBenchmark/Benchmark.swift
@@ -1,5 +1,5 @@
 @_spi(RegexBenchmark) import _StringProcessing
-@_implementationOnly import _RegexParser
+internal import _RegexParser
 import Foundation
 
 protocol RegexBenchmark: Debug {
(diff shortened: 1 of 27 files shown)


tshortli:
@swift-ci please smoke test

ktoso (approved):

tshortli:
@swift-ci please test

PR #756, opened 2024-07-25
[swift/main] Migrate to internal imports

tshortli:
Cherry pick of https://github.com/swiftlang/swift-experimental-string-processing/pull/755.

As of the Swift 6 compiler, `@_implementationOnly import` is deprecated in favor of `internal import` and as a result the use of `@_implementationOnly import` in this project is generating a lot of diagnostic noise when building the Swift standard library.

For Swift libraries with library evolution, `@_implementationOnly import` and `internal import` are roughly functionally equivalent, aside from improved diagnostics for `internal import`. For non-resilient libraries, the main difference is that `internal import` does not actually hide a module dependency from downstream clients because the layout of a type in a non-resilient library may depend on types coming from an `internal import` (with `@_implementationOnly import` the same situation would result in a silent mis-compile, which is the reason that `@_implementationOnly import` is deprecated). The `_RegexParser` module dependency does not need to be hidden from clients since it is installed in standard locations in the SDK/toolchain. Therefore this migration should be safe, regardless of library resilience mode.

### DIFF
--- a/Sources/RegexBenchmark/Benchmark.swift
+++ b/Sources/RegexBenchmark/Benchmark.swift
@@ -1,5 +1,5 @@
 @_spi(RegexBenchmark) import _StringProcessing
-@_implementationOnly import _RegexParser
+internal import _RegexParser
 import Foundation
 
 protocol RegexBenchmark: Debug {
--- a/Sources/RegexBuilder/Anchor.swift
+++ b/Sources/RegexBuilder/Anchor.swift
@@ -9,7 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_implementationOnly import _RegexParser
+internal import _RegexParser
 @_spi(RegexBuilder) import _StringProcessing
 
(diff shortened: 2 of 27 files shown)


tshortli:
@swift-ci please test

tshortli:
@swift-ci please test Windows

shahmishal:
@swift-ci please test Windows

tshortli:
@swift-ci please test Windows

milseman (approved):
LGTM.

I'm not familiar with `internal import`, is there a semantic difference or is that the public spelling for it?

tshortli:
> I'm not familiar with internal import, is there a semantic difference or is that the public spelling for it?

`internal import` is the product of [SE-409](https://forums.swift.org/t/se-0409-access-level-modifiers-on-import-declarations/67290). There are some type checking and compile time differences between the two, but `internal import` should be thought of as the official replacement for `@_implementationOnly import`.

PR #760, opened 2024-08-18
[WIP] Regex lookbehind support

JacobHearst:
A mostly complete implementation of lookbehind support for Regex via reverse matching.

## Forewarning
I'm a casual user of regex, not an expert. I wanted lookbehinds in Swift and was willing to dive into a project I knew nothing about. As a result, I may have made assumptions or choices that don't seem to make sense to more knowledgeable readers.

## Some notes
- There's a lot of duplication of code between `emitReverseXYZ` functions and their `emitXYZ` counterparts.
- The code for forwards matching makes heavy use of the fact that a `String`'s `endIndex` is still a valid index even though there's no character at that index. This means that the reverse variants of these matching functions need to take a different approach. I've tried to address this in some places (reverse quantifications), but others I haven't gotten to yet (`^` assertions)


Sources/RegexBuilder/Anchor.swift
```diff
@@ -226,3 +226,64 @@ public struct NegativeLookahead<Output>: _BuiltinRegexComponent {
...
 }
+
+/// A regex component that allows a match to continue only if its contents
+/// match at the given location.
+///
+/// A lookbehind is a zero-length assertion that its included regex matches at
+/// a particular position. Lookbehinds do not advance the overall matching
+/// position in the input string — once a lookbehind succeeds, matching continues
+/// in the regex from the same position.
+@available(SwiftStdlib 5.7, *) // TODO: How should this be gated?
```
milseman, line 237: We probably want to adjust the wording to denote the "behind" part of the assertion. CC @natecook1000

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -16,6 +16,7 @@ import Swift
 
 extension Compiler {
   struct ByteCodeGen {
+    var reverse = false
```
milseman, line 19: @natecook1000 what do you think about modeling reverse as an option? Do you think that would clean up the implementation?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -326,6 +382,9 @@ fileprivate extension Compiler.ByteCodeGen {
 
     builder.buildSave(success)
     builder.buildSave(intercept)
+    if reverse {
+       builder.buildReverse(1)
```
milseman, line 386: What's the logic behind emitting this reverse?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -365,19 +427,18 @@ fileprivate extension Compiler.ByteCodeGen {
...
-    }
+    reverse = !kind.forwards
     if kind.positive {
-      try emitPositiveLookahead(child)
+      try emitPositiveLookaround(child)
     } else {
-      try emitNegativeLookahead(child)
+      try emitNegativeLookaround(child)
     }
+    reverse = false
```
milseman, line 441: Are nested lookarounds possible? If so, do we need to restore the prior reverse value instead of clearing it?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -743,22 +917,22 @@ fileprivate extension Compiler.ByteCodeGen {
         guard let val = c._singleScalarAsciiValue else {
           return false
         }
-        builder.buildQuantify(asciiChar: val, kind, minTrips, maxExtraTrips, isScalarSemantics: isScalarSemantics)
+        builder.buildReverseQuantify(asciiChar: val, kind, minTrips, maxExtraTrips, isScalarSemantics: isScalarSemantics)
```
milseman, line 920: Should these be guarded by an `if reverse`? (With the diff I can't quite see the full logic here).

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -1324,6 +1505,51 @@ extension DSLTree.Node {
...
     }
   }
+
+  /// A Boolean value indicating whether this node reverses the match position
+  /// on a successful match.
+  ///
+  /// For example, an alternation like `(a|b|c)` always advances the position
+  /// by a character, but `(a|b|)` has an empty branch, which matches without
+  /// advancing.
+  var guaranteesBackwardProgress: Bool {
```
milseman, line 1515: Is the logic in here different than guaranteeing forward progress?

Sources/_StringProcessing/Engine/MEBuilder.swift
```diff
@@ -153,33 +153,61 @@ extension MEProgram.Builder {
...
       .init(.advance, .init(distance: n, isScalarDistance: true)))
   }
-  
+
   mutating func buildConsumeNonNewline() {
     instructions.append(.init(.matchAnyNonNewline, .init(isScalar: false)))
   }
-                        
+
+  mutating func buildReverseConsumeNonNewline() {
```
milseman, line 174: As a thought, would it makes sense to have a required `reverse: Bool` parameter instead of a separate function? This would mean that every call site has to think about whether it is reversing or not, which could help avoid bugs from forgetting to check for reverse. If every call site needs to think about (or pass along) a reverse value, that would also clean up the branches there.

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -30,6 +30,27 @@ extension Processor {
     return true
   }
 
+  mutating func reverseMatchBuiltinCC(
```
milseman, line 33: We have zero-width alternation matches (`a|b|`), are there zero width CC matches?

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -160,7 +181,46 @@ extension String {
...
+  ///     can result in a `nil` return or a character that includes only part of
+  ///     the `self[pos]` character.
+  ///
+  /// - Parameters:
+  ///   - pos: The position to load a character from.
+  ///   - start: The limit for the character at `pos`.
+  /// - Returns: The character at `pos`, bounded by `start`, if it exists, along
+  ///   with the upper bound of that character. The upper bound is always
+  ///   scalar-aligned.
+  func characterAndStart(at pos: String.Index, limitedBy start: String.Index) -> (Character, String.Index)? {
```
milseman, line 207: It's actually the lower bound of the character. Let's add a label to the return tuple value that's something like:
`(Character, characterStart: String.Index)`

Sources/_StringProcessing/Engine/MEBuiltins.swift
```diff
@@ -367,4 +546,87 @@ extension String {
     }
     return next
   }
+
+  @inline(never)
```
milseman, line 550: I'll think over how much code duplication we can avoid here later (and otherwise skip these implementation functions in the review). If we're implementing this much duplication, we'll want to make sure we have very, very solid testing in place.

Tests/RegexTests/MatchTests.swift
```diff
@@ -1575,27 +1577,57 @@ extension RegexTests {
...
     firstMatchTest(
       #"(*positive_lookbehind:USD)\d+"#,
-      input: "Price: USD100", match: "100", xfail: true)
-    // engines generally enforce that lookbehinds are fixed width
+      input: "Price: USD100", match: "100")
+
+    // TODO: Why is a match not found when unoptimized?
     firstMatchTest(
-      #"\d{3}(?<=USD\d{3})"#, input: "Price: USD100", match: "100", xfail: true)
+      #"\d{3}(?<=USD\d{3})"#, input: "Price: USD100", match: "100", validateOptimizations: false)
```
milseman, line 1591: In the unoptimized case we emit `reverse(1)`, I wonder if that's related?

Tests/RegexTests/MatchTests.swift
```diff
@@ -1575,27 +1577,57 @@ extension RegexTests {
...
+
+    firstMatchTests(
+      #"(?<=^\d{1,3})abc"#,
+      ("123abc", "abc"),
+      ("12abc", "abc"),
+      ("1abc", "abc"),
+      ("1234abc", nil), // FIXME: Shouldn't match but does because `^` assertions are broken
+      ("z123abc", nil), // FIXME: Same as above
+      validateOptimizations: false
+    )
```
milseman, line 1630: Let's see if we can have nested lookarounds (like a positive inside a negative inside a positive, etc). You can cross reference other engine's behavior such as Javascript and .NET (e.g., via https://regex101.com).

Tests/RegexTests/MatchTests.swift
```diff
@@ -1575,27 +1577,57 @@ extension RegexTests {
...
     firstMatchTest(
-      #"(*nlb:USD)\d+"#, input: "Price: JYP100", match: "100", xfail: true)
+      #"(*nlb:USD)\d+"#, input: "Price: JYP100", match: "100")
     firstMatchTest(
       #"(*negative_lookbehind:USD)\d+"#,
-      input: "Price: JYP100", match: "100", xfail: true)
-    // engines generally enforce that lookbehinds are fixed width
+      input: "Price: JYP100", match: "100")
+
+    firstMatchTest(
```
milseman, line 1601: I expect us to add lots more tests, so maybe extract into a separate function.

milseman (commented):
This is looking really awesome! I'm going to be trying to think about ways to clean the code up a bit (and avoid as much duplication).

@natecook1000 can you help me review this and integrate it with me?

I also want to get much more sanity checking in place. For example, we run every test both optimized and unoptimized, which helps catch a lot of bugs. We are _so_ close to generalized reverse matching that I wonder if we can similarly do something for reverse matching. For example, every test that matches forwards from the start of a string will also match when compiled in reverse from the end of a string. That will also give us an extra layer of nesting.

natecook1000:
Thanks so much for this contribution, @JacobHearst! 

Two additional notes outside of the implementation:

1. We'll need to have a discussion about whether and how to limit the kinds of patterns that can be used inside a lookbehind. Some other regex engines only allow fixed-length patterns (Perl, Python, e.g. `/(?<=aa)b/`) or finite-length patterns (Java, e.g. `/(?<=aa?)b/`), while others are more open-ended (.NET, Javascript). If we do want to limit the kinds of patterns, we might need an additional protocol to constrain the kinds of patterns that can appear in a `Lookbehind` or `NegativeLookbehind` regex builder block. ([more info](https://www.regular-expressions.info/lookaround.html#limitbehind))

2. We'll also need to look at how to manage availability with literals, since older Swift runtimes won't have the necessary support for lookbehinds. (This is an issue that we would need to manage with support for any additional regex syntax — this is just the first one to raise the question.)

milseman:
Here's a couple examples of nesting (https://regex101.com/r/ZnXq00/1). We'll want to expand this testing to cover even deeper nesting, test nesting alternations, custom character classes, etc., to make sure we get full coverage of our complex constructs:

```swift
input = "abcdefg"

// firstMatch results:

/abcd(?<=c(?=d)d)/     // abcd
/abcd(?<=cd(?=d).)/    // no match (. consumed the d)
/abcd(?<=c(?=e)d)/     // no match (local position expecting a lookahead of d)
/abcd(?<=bc(?=de)d)/   // abcd
/abcd(?<=bc(?=de).)/   // abcd
```

I'd suggest switching to the extended syntax (https://github.com/swiftlang/swift-evolution/blob/main/proposals/0354-regex-literals.md#extended-delimiters-- ) for the more complex nesting at some point

Sources/_StringProcessing/ByteCodeGen.swift
JacobHearst: All the `match` instructions match the current position, so my logic was in order to start matching in reverse, we needed to step back once and _then_ start the match

Sources/_StringProcessing/ByteCodeGen.swift
JacobHearst: I haven't tried yet but I suspect with this implementation they are not. Seems like a good thing to support though

Sources/_StringProcessing/ByteCodeGen.swift
JacobHearst: Nope, this is part of `tryEmitFastReverseQuant`. The diff is showing `tryEmitFastQuant` as an addition and the reverse variant as a change to the original

Sources/_StringProcessing/Engine/MEBuilder.swift
JacobHearst: Yeah I think that would drastically cut down on duplicated code as well.

Sources/_StringProcessing/ByteCodeGen.swift
JacobHearst: Doesn't look like it, I'll remove it

Sources/_StringProcessing/ByteCodeGen.swift
milseman: Are we going to be emitting the reverse variant of those matching instructions?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -125,9 +125,9 @@ fileprivate extension Compiler.ByteCodeGen {
         let boundaryCheck = idx == lastIdx
         let scalar = s.unicodeScalars[idx]
         if options.isCaseInsensitive && scalar.properties.isCased {
-          builder.buildMatchScalarCaseInsensitive(scalar, boundaryCheck: boundaryCheck)
+          builder.buildMatchScalarCaseInsensitive(scalar, boundaryCheck: boundaryCheck, reverse: false)
```
milseman, line 128: Let's add an assertion at the top of this function that we're not in reverse mode.

```swift
assert(!reverse)
```

That will help us catch times we're calling this in error

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -153,9 +153,9 @@ fileprivate extension Compiler.ByteCodeGen {
         let boundaryCheck = idx == lastIdx
         let scalar = s.unicodeScalars[idx]
         if options.isCaseInsensitive && scalar.properties.isCased {
-          builder.buildReverseMatchScalarCaseInsensitive(scalar, boundaryCheck: boundaryCheck)
+          builder.buildMatchScalarCaseInsensitive(scalar, boundaryCheck: boundaryCheck, reverse: true)
```
milseman, line 156: Similarly, an assertion that we are in reverse mode

Sources/_StringProcessing/ByteCodeGen.swift
JacobHearst: I believe so. Assuming the save point machinery doesn't have to change, we call `emitNode` for the pattern inside the lookbehind so it should be reversed so long as `reverse` is true.

JacobHearst: So long as the `reverse` flag is set, yes. Thinking about nested lookarounds though, I think my current implementation of the `reverse` flag will have a bug here because of what you mentioned in another comment

> Are nested lookarounds possible? If so, do we need to restore the prior reverse value instead of clearing it?

Sources/_StringProcessing/Engine/MEBuiltins.swift
JacobHearst: Is it possible for a character class to be zero-width?

milseman:
Here's what we need to ultimately ship:

1. Expand testing. The large amount of semi-duplicated code in the engine for handling reverse match variants is acceptable if we have really good testing.
2. Reject custom regex consumers in the builder syntax from being embedded in a lookbehind, as these are not reversible.
3. Limit availability. We need to error out at parse time for literals that use lookbehind when targeting an OS that may not support them.

milseman:
@JacobHearst could you rebase this work on top of the current swift/main?

@natecook1000 when I try to build these tests, sometimes it works and sometimes I get the error:

```
Undefined symbol: static RegexBuilder.AlternationBuilder.buildExpression<A where A: _StringProcessing.RegexComponent>(A) -> A
... 31 more errors
```

have you seen this before?

Sources/_StringProcessing/ByteCodeGen.swift
```diff
@@ -110,7 +111,8 @@ fileprivate extension Compiler.ByteCodeGen {
   mutating func emitQuotedLiteral(_ s: String) {
```
JacobHearst, line 111: @milseman Curious to get your thoughts on combining the `emitReverseQuotedLiteral` and `emitQuotedLiteral` like I've done here. I'm a bit torn on whether it's better to duplicate code or add complexity here.

JacobHearst: Obviously this doesn't preclude the need for unit tests but I think it'll make maintenance easier in the long term if there aren't two nearly identical copies of functions all over the place

Tests/RegexTests/MatchTests.swift
JacobHearst: I ended up fixing this test case by checking if we're at the `start` during `reverseMatch` _and_ `reverseScalarMatch`

Sources/_StringProcessing/ByteCodeGen.swift
milseman: Avoiding duplication in the compilation makes sense. For the processor itself, I don't want a dynamic branch at the lowest-most part of matching to choose between forwards and reverse, so there's more nuance there. We might be able to extract functionality into common shared code, or we might do some kind of static meta programming.

JacobHearst:
After many many months I am finally back to working on this feature. I was finding it quite difficult to pick up where I left off so I've decided to start (mostly) fresh on a new branch. @milseman how would you like to proceed with this PR? Close it and open a new one from the new branch?

JacobHearst:
Just for transparency, the main reasons for the long gap in my work on this are:
- Most of my brainpower is soaked up by my job these days leaving me with little to no energy for outside projects
- The longer I go without working on this, the more intimidating it is to start up again

Knowing this, my current plan to bring this work to completion revolves around identifying functions I wrote reverse variants of and writing unit tests for the original and my variant. My goal right now is to unit test 1 pair of functions a week so we'll see how that goes. I would really like to get this into Swift 6.2 (I suspect I've already missed or will miss the boat for 6.1) and I think this plan will let me keep some level of momentum to achieve this.

PR #763, opened 2024-09-24
Add missing imports of `_RegexParser`

tshortli:
While experimenting with adopting the `MemberImportVisibility` experimental feature (SE-0444) in the standard library build, I found that the `_StringProcessing` module was relying on transitive imports of `_RegexParser` in many source files.

### DIFF
--- a/Sources/_StringProcessing/Unicode/NFC.swift
+++ b/Sources/_StringProcessing/Unicode/NFC.swift
@@ -9,6 +9,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+internal import _RegexParser
+
 @_spi(_Unicode)
 import Swift
 
--- a/Sources/_StringProcessing/Engine/MEQuantify.swift
+++ b/Sources/_StringProcessing/Engine/MEQuantify.swift
@@ -1,3 +1,5 @@
+internal import _RegexParser
+
 private typealias ASCIIBitset = DSLTree.CustomCharacterClass.AsciiBitset
 
 extension Processor {
--- a/Sources/_StringProcessing/Engine/Processor.swift
+++ b/Sources/_StringProcessing/Engine/Processor.swift
@@ -9,6 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+internal import _RegexParser
 
 enum MatchMode {
   case wholeString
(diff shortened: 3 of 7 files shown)


tshortli:
@swift-ci please smoke test

tshortli:
@swift-ci please test

Sources/_StringProcessing/Unicode/NFC.swift
```diff
@@ -9,8 +9,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(_Unicode)
-import Swift
+@_spi(_Unicode) import Swift
```
Azoy, line 12: I'd prefer the attribute to appear on a newline above the import.

Azoy (approved):

Sources/_StringProcessing/Unicode/NFC.swift
tshortli: Ok. How do you see this style mixing with multiple imports in the same file? To me this looked rather unusual and made it hard to scan the block of imports. I could also add a newline before `internal import _RegexParser` but that too looks unusual.

Azoy: My personal preference for attributes is always above the declaration it's attributing to. In the standard library for functions we have:
```swift
@available(SwiftStdlib 6.0, *)
func something() {
}
```
right, and for property wrappers my preference is also above the property declaration:
```swift
struct A {
  @propertyWrapper
  let b: Int

  let c: String
}
```
I think the newline before the 2nd property here makes sense to me and would similarly apply it to imports:
```swift
@attribute
import Swift

import A
import B
```
But again, this is just personal preference. I don't think we have a defined style yet for this because we just simply don't write imports with attributes often at all.

tshortli: I generally agree that attributes belong above declarations, especially when the list of attributes is likely to grow and line lengths can become an issue. For whatever reason, though, vibes-wise I don't think of `imports` as following the same rules as other declarations because it's so common for there to be an uninterrupted block of them at the top of a file. `@_spi` on a separate line really sticks out like a sore thumb to me, never seen it outside of this codebase. I'll follow your recommendation here, though, as I agree it seems to be the most consistent rule.

tshortli:
@swift-ci please test

PR #767, opened 2024-10-01
Customize `NSError.localizedDescription` for Regex

benrimmington:
Regex initializers currently throw `any Error` if the given pattern is invalid.

```swift
import Foundation

do {
  let regex = try Regex(#"\"#)
} catch {
  print(error)
  print(error.localizedDescription)
}
```

* The diagnostic message is useful:

  > expected escape sequence

  but can it be relied upon? This pull request adds a test.

* The localized description contains a 138-character error domain, and a meaningless error code:

  > The operation couldn’t be completed. (_RegexParser.Source.LocatedError<_RegexParser.Diagnostics.(unknown context at $26f700fac).(unknown context at $26f700fb4).ErrorDiagnostic> error 1.)

  This pull request customizes the error domain, and appends the diagnostic message:

  > The operation couldn’t be completed. (_RegexParser.Diagnostics error 0 - expected escape sequence)

  The error code is still meaningless, but so are those derived from ParseError — because the compiler seems to reorganize enum cases by payload.

---

Alternatives:

1.  Typed throws?

2.  Use the same error domain and code as [NSRegularExpression](https://github.com/swiftlang/swift-corelibs-foundation/blob/swift-6.0.1-RELEASE/Sources/CoreFoundation/CFRegularExpression.c#L145) but without the NSInvalidValue key?

### DIFF
--- a/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
+++ b/Sources/_RegexParser/Regex/Parse/Diagnostics.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+// Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


benrimmington:
@swift-ci Please test

PR #791, opened 2024-11-28
[build] Add missing source file to CMake

egorzhdan:
`SubstringSearcher.swift` wasn't in the list of source files for CMake target _StringProcessing, which caused a compilation error for me locally while building the Swift compiler.

### DIFF
--- a/Sources/_StringProcessing/CMakeLists.txt
+++ b/Sources/_StringProcessing/CMakeLists.txt
@@ -6,6 +6,7 @@ add_library(_StringProcessing
   Algorithms/Algorithms/Replace.swift
   Algorithms/Algorithms/Split.swift
   Algorithms/Algorithms/StartsWith.swift
+  Algorithms/Algorithms/SubstringSearcher.swift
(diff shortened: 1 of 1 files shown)


egorzhdan:
@swift-ci please test

milseman:
@natecook1000 any idea how this was building without this change?

egorzhdan:
Perhaps this isn't normally built with CMake, and that is why we haven't seen the failure before?

@milseman @natecook1000 do you think we could land this change?

natecook1000:
Did some looking into this, and the answer has two parts: (1) we don't normally build this project with CMake and don't have tests for doing that, and (2) when this project is included in the Swift project build, it doesn't use the CMake machinery here at all. Instead, it has its own CMakeLists where it just grabs all the Swift files: https://github.com/swiftlang/swift/blob/main/stdlib/public/StringProcessing/CMakeLists.txt#L30-L39

As such, I'm going to recommend we actually just remove the CMake machinery here instead of fixing it.

PR #803, opened 2025-01-23
Conversion of Regex syntax to BNF syntax

milseman:
Adds support in `_RegexParser` for an initial approach to converting from regex syntax to BNF syntax.


milseman:
@swift-ci please test

PR #825, opened 2025-09-17
Include optionality in AnyRegexOutput.Element.type

natecook1000:
When reconstructing the type of a captured element for access through the `AnyRegexOutput.Element.type` API, any optionality is being omitted (e.g. `Int?` is returned as `Int`). This is both observable through that API and results in a bug when accessing part of a match's output through the dynamic member subscript (e.g. `match.1`).

In that case, the lack of optionality causes an incorrect calculation of the output tuple member's position, resulting in the wrong member being loaded and returned.

This change adds the missing optionality to the type.

Fixes https://github.com/swiftlang/swift/issues/83022. (rdar://155710126)

### DIFF
--- a/Sources/_StringProcessing/Regex/AnyRegexOutput.swift
+++ b/Sources/_StringProcessing/Regex/AnyRegexOutput.swift
@@ -381,7 +381,12 @@ extension AnyRegexOutput.ElementRepresentation {
   }
 
   var type: Any.Type {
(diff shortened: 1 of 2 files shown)


natecook1000:

milseman:
LGTM. Did you try a test with nested optionals?

natecook1000:

PR #826, opened 2025-09-18
Make CCC compilation deterministic

natecook1000:
This fixes an issue where using a set to unique the elements of a CCC led to a nondeterministic ordering, which could then lead to unexpected coalescing of separate scalars into a single character. (e.g. `[\nA\r]` -> `["\r", "\n", "A"]` -> `["\r\n", "A"]`)

This change keeps the order of the original CCC, so coalescing into characters should not occur unless the scalars are already in that order.

Fixes https://github.com/swiftlang/swift/issues/81427. (rdar://151046715)


natecook1000:
@swift-ci Please smoke test Linux platform

natecook1000:

PR #828, opened 2025-09-29
Start of line in negative lookahead optimization fix

natecook1000:
Don't use the content of negative lookaheads when determining whether a start-of-line assertion can require that match has to start at the beginning of the subject.

Fixes https://github.com/swiftlang/swift#81789.
rdar://152119639


natecook1000:

stephentyrone (approved):

PR #829, opened 2025-09-29
Remove CMakeList.txts

natecook1000:
This library doesn't support CMake builds directly, only via inclusion in the Swift project, so these files are out of date and aren't necessary to maintain.


natecook1000:

PR #856, opened 2026-03-11
Fix crash in Regex builder with many concatenated components

rjmansfield:
Guard against empty children in indexOfCoalescableAtom to prevent range underflow in 0..<(children.count - 1).


rjmansfield:
@swift-ci please test

PR #858, opened 2026-04-10
Reduce amount of \Q...\E quoting in _literalPattern

natecook1000:
This change switches to manually escape regex metacharacters,
which lessens the circumstances that use the more cumbersome
regex quoting syntax.


natecook1000:

PR #859, opened 2026-04-10
Fix a concatenation coalescing issue

natecook1000:
This addresses an out of bounds access when coalescing certain concatenation patterns.

rdar://174353630


natecook1000:

PR #866, opened 2026-06-20
Fix forward progress detection for lookaround groups

### LINKED ISSUE #865: RegexBuilder infinite loop when nullable capture starts with NegativeLookahead
### Description 

A RegexBuilder pattern can hang when an unbounded quantifier repeats a body that can match the empty string, where that body begins with `NegativeLookahead`.

In the reducer below, matching `"A"` repeatedly invokes the capture transform with an empty substring without advancing through the input.

### Reproduction

```swift
import RegexBuilder

let regex = Regex {
    ZeroOrMore {
        Capture {
            NegativeLookahead { "a" }
            ZeroOrMore(.digit)
        } transform: { String($0) }   // invoked repeatedly with ""
    }
}

_ = "A".matches(of: regex)   // never returns
```

Reduced string form:

```swift
_ = try! Regex(#"(?:(?!a)\d*)*"#).firstMatch(in: "A")   // never returns
```

### Expected Behavior

Matching completes, similar to other nullable quantified forms such as `(?:\w|)+` (#558) and `(a*)*` (#542). Since the repeated body can match the empty string, the engine should stop repeating when the current position does not advance.

### Environment

`swiftlang/swift:nightly-6.4.x-noble` (Swift 6.4-dev) or `swiftlang/swift:nightly-main` (Swift 6.5-dev). Works as expected on `swiftlang/swift:nightly-6.3-noble` (Swift 6.3.3-dev) and `swift:6.3` (Swift 6.3.2 release).
  
### Notes

This appears to be a regression from the DSL-list flattening work in #845, which was reverted in #847, then re-introduced by #849. The reducer passes at `d474af96^` and fails at `d474af96`. The issue is in the same forward-progress class as #851, which taught the walker to skip a nullable quantification's child subtree. Lookaround groups appear to need the same treatment.

ekassos:
When a quantified concatenation begins with a lookaround group, the forward progress checker fails to skip over the assertion's child subtree, leaving the analysis cursor misaligned and leading to false positives.

This change skips the lookaround's child subtree, matching the treatment added for nullable quantifications in #851.

Fixes #865.

### DIFF
--- a/Sources/_StringProcessing/ByteCodeGen+DSLList.swift
+++ b/Sources/_StringProcessing/ByteCodeGen+DSLList.swift
@@ -369,6 +369,8 @@ fileprivate extension Compiler.ByteCodeGen {
     case .nonCapturingGroup(let kind):
       switch kind.ast {
       case .lookahead, .negativeLookahead, .lookbehind, .negativeLookbehind:
+        list.skipNode(&position)
+        position += 1
         return false
       default:
         return _guaranteesForwardProgressImpl(list, position: &position)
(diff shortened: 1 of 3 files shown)


natecook1000:

natecook1000:
Thanks for this fix, @ekassos!

PR #870, opened 2026-07-29
Replace per-save-point array snapshots with an undo trail

natecook1000:
Creating a save point previously copied the four mutable register arrays (ints, positions, values, and storedCaptures) to enable restoration during backtracking. This led to a large amount of reference counting traffic and copy-on-write of the original register arrays whenever they were updated.

This change moves the restoration data from using full copies of each array in every save point, to tracking a log of the changes to each array in the processor itself and only storing the current state of the logs in each save point. This allows `SavePoint` to be a trivial type (almost) and removes all the copy-on-write traffic. Backtracking now uses the tracking logs to unwind the registers to the specific previous state.

This change also includes a minor refactor, moving the `storedCaptures` array to the `Registers` struct, alongside the other mutable registers. Future work includes breaking out the mutable registers into their own type, creating a flat structure for these instead of maintaining four different dynamically allocated arrays, and moving further toward making `SavePoint` trivial.


natecook1000:

natecook1000:

PR #872, opened 2026-08-27
Include progress checks with bounded repetition

natecook1000:
This enables forward progress checks for all quantification kinds, which were previously skipped for bounded quantification (i.e. `/^{,5}/`) as an optimization. Even bounded quantifications can cause runaway matching execution time when not progressing through the input, particularly if nested.

rdar://185731142


Sources/_StringProcessing/ByteCodeGen+DSLList.swift
clintonpi:
```suggestion
        <if can't guarantee forward progress>:
```

clintonpi:
```suggestion
        <if can't guarantee forward progress>:
```

Sources/_StringProcessing/ByteCodeGen+DSLList.swift
```diff
@@ -585,9 +585,7 @@ fileprivate extension Compiler.ByteCodeGen {
     // we have an unbounded quantification
```
clintonpi:
```suggestion
// if we aren't sure if the child node will have forward progress
```

